PR #19, opened 2020-07-26
Language and country selection modal

ankitecd:
Changes in this pull request:
- Added language and country selection modal opened from footer


bash.exe.stackdump
```diff
@@ -0,0 +1,9 @@
+Stack trace:
```
norbertschuler, line 1: This surely should not be committed, should it?

src/utils/supportedLanguages.json
```diff
@@ -0,0 +1,30 @@
...
+    "languageName": "Brazilian Portuguese"
+  },
+  {
+    "langCode": "de",
+    "languageName": "German"
+  },
+  {
+    "langCode": "sv",
+    "languageName": "Swedish"
+  }
```
norbertschuler, line 29: If you don't want to translate the language name in every language later on, I recommend using the native language name here as I did to the Treecounter app here (without Swedish): https://github.com/Plant-for-the-Planet-org/treecounter-app/blob/175f7421d5aeb51a6adeabf0b263c720bbb361f3/app/components/Menu/index.js#L91-L98

src/utils/countriesData.json
```diff
@@ -0,0 +1,1752 @@
+[
+  {
+    "countryName": "New Zealand",
+    "countryCode": "NZ",
+    "currencyName": "New Zealand Dollars",
+    "currencyCode": "NZD",
+    "currencyCountryFlag": "NZ"
+  },
```
norbertschuler, line 8: If you ever want to use translations for the country names, you need this file in every language we want to support. This is still an unsolved problem in the Treecounter app yet (but there are free sources on the internet providing this information).

bash.exe.stackdump
ankitecd: Yes it should not be. Deleted the file thanks :)

src/utils/countriesData.json
harshvitra: We will add this later

PR #38, opened 2020-08-07
fixed marker hover issue

sarveshpro:
Fixes #

Changes in this pull request:
- added .3s delay in detection of hover
- popup goes away .3s after unhover
- fixed padding inside popup

@


sagararyal:
introduces some ui issues: as you removed the class for project

sarveshpro:
ah okay, I'll fix them

PR #82, opened 2020-08-20
Feature/tenant config

harshvitra:
Tenant Configurations


.env
```diff
@@ -2,4 +2,5 @@
 ## Overridden by .env.local or .env.development or .env.production
 ## depending upon the environment
 
-TENANT=
+TENANT=salesforce
```
sagararyal, line 5: this should not be added in .env

PR #134, opened 2020-09-02
Beta Release

sagararyal:


norbertschuler (changes requested):

.env.local.example
```diff
@@ -0,0 +1,14 @@
+MAPBOXGL_ACCESS_TOKEN=
+API_ENDPOINT=app-development.plant-for-the-planet.org
+CDN_URL=cdn.pp.eco
```
Plant-for-the-Planet, line 3: cdn.plant-for-the-planet.org

PR #135, opened 2020-09-02
added Salesforce font

ankitecd:
Changes in this pull request:
- Added Salesforce font

@harshvitra please check


src/theme/_fonts.scss
```diff
@@ -1,2 +1,2 @@
-$primaryFontFamily: 'Raleway', sans-serif;
+$primaryFontFamily: 'Salesforce Sans', sans-serif;
```
sagararyal, line 1: @ankitecd 
This should be overridden on tenant config, and not on app level configuration.

Default font on other tenant is still Raleway.

Also fonts needs to cover Bold and other texts so additional variants need to be loaded as well.

src/theme/global.scss
```diff
@@ -9,3 +9,9 @@
 .mapboxgl-ctrl.mapboxgl-ctrl-attrib {
   background: transparent;
 }
+
+@font-face {
+  font-family: 'Salesforce Sans';
+  src: url('https://cdn.plant-for-the-planet.org/media/fonts/salesforce/SalesforceSans-Regular.woff')
```
sagararyal, line 15: again, not  on global css; but instead tenant based config.

sagararyal (approved):

sagararyal (changes requested):
Approval was by mistake, it was request for changes

PR #173, opened 2020-09-09
Fix visa logo

sarveshpro:
Changes in this pull request:
- Fix visa logo


sagararyal (approved):
works on safari and chrome.

PR #185, opened 2020-09-13
Feature/responsiveness

sarveshpro:
Fixes #

Changes in this pull request:
-


sagararyal:
Screen Shot 2020-09-13 at 9 48 05 PM


the transform events takes too much memory. We don't need transform for individual divs

Also the wobbling is still there for scroll

sarveshpro:
Will change the scroll implementation. this won't work properly.

pages/index.tsx
```diff
@@ -52,8 +52,6 @@ export default function Donate() {
     return () => window.removeEventListener('wheel', handleScroll);
   });
   return (
-    <Layout>
-      {projects ? <ProjectsList {...DonateProps} /> : <h2>Loading...</h2>}
-    </Layout>
+    <Layout>{projects ? <ProjectsList {...DonateProps} /> : <h2></h2>}</Layout>
```
Plant-for-the-Planet, line 55: can we make the screen color same as map blue when this loads?
or instead load the project loader here?

sarveshpro: I am sure this is done in fix scroll PR. the color is not accurate though I have messaged Killian to get the actual color.

sarveshpro: also, scroll PR should have all the changes in this PR so instead merge that and close this PR.

PR #206, opened 2020-09-16
Config loader is added

harshvitra:
A loader has been added until config is added in local storage


Plant-for-the-Planet:
Do not merge, looking for alternatives.

PR #220, opened 2020-09-17
Share donations Bugs Fix

### LINKED ISSUE #197: Remaining Items from Donation Share PR
- [ ] Need to increase the size/resolution of the downloaded image.

- [ ] When clicking the twitter icon or the email icon on my android it opens the share card instead of twitter even though I have the twitter app installed (same for insta and fb but I don't have those apps)

- [ ] In the download the logo and the trillion trees is super low res while the adjusted text looks great. But I think you know that
 
- [ ] The plant trees at www.trillion does not need "www." And it should be right at the bottom of the card. Ideally, in the card preview it would be a bit further up (as it is now) to make space for the share buttons, but once you actually download it, the text would be further down. Is that possible?

nirali25parekh:
Fixes #197 

Changes in this pull request:
- image download quality improvement

@sagararyal


harshvitra:
This is better than before

nirali25parekh:
We can still improve quality as much as we want, but then the image size keeps on increasing. This settings gives about 2.2Mb image

sagararyal (commented):
If you can fix. Not all browsers support webp; fallback image would be great.

Another is image downloaded as datafile. 
On iOS devices top part of image is cut off.

sagararyal:
> We can still improve quality as much as we want, but then the downloaded file size keeps on increasing. This setting gives about 2.2Mb image

Image size shouldn't be a problem.

There is however an issue when share icons are shared it is not working properly. - it uses browser share api. 

Only one share button can be clicked after a donation

Should instead redirect properly to Facebook/etc.

nirali25parekh:
changes:
- image quality increased 
- share on phone browser fixed (facebook doesn't support sharing intent to app, so browser share opens on clicking fb logo)
- fallback jpeg used instead of webp

PR #236, opened 2020-09-19
added i18n

norbertschuler:
This is a just a simple test of https://react.i18next.com/ and https://www.i18next.com/ integration (mostly copied from https://lazertechnologies.com/blog/how-to-add-localization-to-your-nextjs-app).

Read the manual to learn more about:
- https://react.i18next.com/guides/multiple-translation-files
- https://react.i18next.com/guides/extracting-translations
- https://www.i18next.com/translation-function/interpolation
- https://www.i18next.com/translation-function/formatting
- https://www.i18next.com/plurals.html

Open issues:
- [ ] https://github.com/Plant-for-the-Planet/planet-webapp/pull/236#issuecomment-697180410
- [ ] https://github.com/Plant-for-the-Planet/planet-webapp/pull/236#issuecomment-698202800
- [ ] https://github.com/Plant-for-the-Planet/planet-webapp/pull/236#issuecomment-698203271
- [x] https://github.com/Plant-for-the-Planet/planet-webapp/pull/236#issuecomment-698150755
- [x] https://github.com/Plant-for-the-Planet/planet-webapp/pull/236#issuecomment-698726776

PR #245, opened 2020-09-20
added translation to pages

ankitecd:
@harshvitra , @norbertschuler , @sagararyal please check

[do not merge until approved]


public/locales/de/donate.ts
```diff
@@ -0,0 +1,54 @@
+import { ILocaleObject, ILocaleParams } from './../localeInterfaces';
+
+const donate: ILocaleObject = {
+  topProjects: 'Top Projekte',
+  allCountProjects: ({ projectCount }: ILocaleParams) =>
```
norbertschuler, line 5: I do not know if LingoHub can work with TypeScript files. (I doubt it.) Especially when writing functions within the resource files to fill placeholders. Let's hope I am wrong. Otherwise how should we give this to translators?

ankitecd: Actually you are right. Just had a talk with Sagar also checked the format LingoHub supports this won't work.

norbertschuler: I demonstrated how to use multiple JSON files and placeholders in my pull request https://github.com/Plant-for-the-Planet/planet-webapp/pull/236.

norbertschuler: @ankitecd would you like to move your translation to #236 ?

PR #255, opened 2020-09-22
initialize sentry

sarveshpro:


norbertschuler (approved):
This works for me now. 

Only Vercel (having the env variable VERCEL_GITHUB_COMMIT_SHA set) will upload source maps to Sentry for now, so they are only available for the production systems if the related branches are also build with Vercel. 

At least the env variable NEXT_PUBLIC_SENTRY_DSN will have to be added on the production systems to make them report issues to Sentry.

sagararyal:
Is it possible to enable sentry without uploading source maps? For heroku @sarveshpro

sarveshpro:
Yes @sagararyal just add `NEXT_PUBLIC_SENTRY_DSN` in env and it will work

PR #264, opened 2020-09-23
Round off Amount

harshvitra:
Rounding off amount value

IMP please test, and consider we might lose on lot of information


harshvitra:
Just realised stripe is getting the full value and we are sending backend rounded value this will lead to inconsistency

PR #265, opened 2020-09-23
I18n automatic detect of browser language

norbertschuler:
- Make automatic browser language detection work by using same localStorage 'language' for i18n and our own app.
- Also removed 'Loading...' and instead show empty page.
- Do not force DACH users to use German language, do we?

### DIFF
--- a/i18n/client.js
+++ b/i18n/client.js
@@ -6,14 +6,16 @@ export default new NextI18Next({
   use: [LanguageDetector, initReactI18next],
   defaultLanguage: 'en',
   fallbackLng: 'en',
+  debug: false,
   otherLanguages: ['de'], // list all languages here
   detection: {
-    // check if language is cached in cookies, if not check local storage
-    order: ['cookie', 'localStorage'],
+    // check if language is cached in cookies, if not check local storage, 
+    // last retrieve from browser language
+    order: ['cookie', 'localStorage', 'navigator'],
 
     // next-i18next by default searches for the 'next-i18next' cookie on server requests
-    lookupCookie: 'next-i18next',
-    lookupLocalStorage: 'i18nextLng',
+    lookupCookie: 'language',
+    lookupLocalStorage: 'language',
 
     // cache the language in cookies and local storage
     caches: ['cookie', 'localStorage'],
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -29,8 +29,8 @@ if (process.env.NEXT_PUBLIC_SENTRY_DSN) {
 }
 
 export default function PlanetWeb({ Component, pageProps, err }: any) {
-  const { Trans, useTranslation } = i18next;
-  const { i18n } = useTranslation();
+  // const { useTranslation } = i18next;
+  // const { i18n } = useTranslation();
 
   const tagManagerArgs = {
     gtmId: process.env.NEXT_PUBLIC_GA_TRACKING_ID,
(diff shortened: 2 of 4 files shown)


i18n/client.js
```diff
@@ -6,14 +6,16 @@ export default new NextI18Next({
...
   defaultLanguage: 'en',
   fallbackLng: 'en',
+  debug: false,
   otherLanguages: ['de'], // list all languages here
   detection: {
-    // check if language is cached in cookies, if not check local storage
-    order: ['cookie', 'localStorage'],
+    // check if language is cached in cookies, if not check local storage, 
+    // last retrieve from browser language
+    order: ['cookie', 'localStorage', 'navigator'],
```
norbertschuler, line 14: Added 'navigator' for automatic browser language detection here if cookie or storage are empty.

i18n/client.js
```diff
@@ -6,14 +6,16 @@ export default new NextI18Next({
...
 
     // next-i18next by default searches for the 'next-i18next' cookie on server requests
-    lookupCookie: 'next-i18next',
-    lookupLocalStorage: 'i18nextLng',
+    lookupCookie: 'language',
```
norbertschuler, line 17: Using the same name for language storage as our own app to share this data between i18n and our own application.

pages/_app.tsx
```diff
@@ -72,16 +72,17 @@ export default function PlanetWeb({ Component, pageProps, err }: any) {
...
   }, []);
 
-  React.useEffect(() => {
-    if (localStorage.getItem('language') !== null) {
-      i18n.changeLanguage(localStorage.getItem('language'));
-    } else {
-      i18n.changeLanguage('en');
-    }
-  }, []);
+  // Norbert: language gets detected by i18next-browser-languagedetector
```
norbertschuler, line 75: We do not need this. If browser detection does not work, i18n also stores 'en' as fallback.

pages/_app.tsx
```diff
@@ -72,16 +72,17 @@ export default function PlanetWeb({ Component, pageProps, err }: any) {
...
+  // React.useEffect(() => {
+  //   if (localStorage.getItem('language') !== null) {
+  //     i18n.changeLanguage(localStorage.getItem('language'));
+  //   } else {
+  //     i18n.changeLanguage('en');
+  //   }
+  // }, []);
 
   if (!initialized) {
-    return <p>Loading...</p>;
```
norbertschuler, line 84: Empty space looks better then 'Loading...'.

src/features/common/Footer/SelectLanguageAndCountry.tsx
```diff
@@ -52,6 +52,7 @@ export default function TransitionsModal(props) {
   function handleOKClick() {
     window.localStorage.setItem('language', modalLanguage);
     setLanguage(modalLanguage);
+    i18n.changeLanguage(modalLanguage);
```
norbertschuler, line 55: This is a better place for the i18n.changeLanguage() call then below.

src/features/common/Footer/index.tsx
```diff
@@ -55,14 +55,15 @@ export default function Footer() {
...
 
-    if (
-      selectedCountry === 'DE' ||
-      selectedCountry === 'AT' ||
-      selectedCountry === 'CH'
-    ) {
-      setLanguage('de');
-      localStorage.setItem('language', 'de');
-    }
+    // Norbert: we do not want users in DACH force to use the website in German, do we?
```
norbertschuler, line 58: I don't think we really need this. What exactly is the use case for this?

PR #269, opened 2020-09-25
Added session IDs for projects and single project endpoint

harshvitra:
Adds session ID for projects and single project endpoint


Plant-for-the-Planet:
@harshvitra I think we should use cookies/session-storage for storing the sessionID.
@norbertschuler @jmiridis what do you suggest?


https://scotch.io/@PratyushB/local-storage-vs-session-storage-vs-cookie

Issue #312, opened 2020-10-12
Implement Postal Code Validation in Address 


sagararyal:
Here's a json with Country Code and regex for zip codes around the world.
https://gist.github.com/ShreyKumar/43ebb040383c90ed976439245423aa38

For countries that doesn't exist in this list no validation is necessary.

PR #319, opened 2020-10-13
zip code validation

### LINKED ISSUE #315: Disallow donation of 0 trees
The Continue button can be disabled when the donation amount is 0 or less. However this might be tricky with payment request button.

As an alternative if 0 trees are selected override the value with minimum tree amount.
    
Screen Shot 2020-10-12 at 11 30 46 PM

Screen Shot 2020-10-12 at 11 31 32 PM

Screen Shot 2020-10-12 at 11 32 21 PM

captainamiedi:
Fixes #315 #312 #327 

Changes in this pull request:
- zip code validation
- set tree count to 1 if user input 0 tree
- Email Validation
- [x] on submission, need to be click once not twice

@sagararyal 
tree
zip code


src/utils/countryZipCode/mergedCountry.js
```diff
@@ -0,0 +1,16 @@
...
+const mergedCountries = () => {
+  const mergedConutryArray = [];
+  Country.forEach((element) => COUNTRY_ADDRESS_POSTALS.forEach((ele) => {
+    if (element.countryName === ele.name) {
+      const newArr = { ...element, ...ele };
+      mergedConutryArray.push(newArr);
+    }
+  }));
+
+  return mergedConutryArray;
```
sagararyal, line 13: please correct the typo.

src/utils/countryZipCode/mergedCountry.js
```diff
@@ -0,0 +1,16 @@
+import COUNTRY_ADDRESS_POSTALS from './index';
+import Country from '../countryCurrency/countriesData.json';
+
+const mergedCountries = () => {
+  const mergedConutryArray = [];
```
norbertschuler, line 5: Fix type `mergedConutryArray` to `mergedCountryArray`.

norbertschuler (changes requested):

src/utils/countryZipCode/mergedCountry.js
captainamiedi: My bad. I have fixed it

norbertschuler:
I made a small test and found two issues:

- [ ] if you fill in the form automatically with the Chrome browser, the error message for the zip code get displayed, even if the automatically filled in value is correct and it does not go away that easily any more (automatically filled in form in Chrome are marked in light blue color): 
Bildschirmfoto 2020-10-20 um 16 34 30

- [ ] if you change the zip code to invalid data (e.g. numbers and letters for countries only allowing numbers), the error message is misleading by telling you that only numbers and letters are allowed - so maybe change it to "zip code is invalid" or we need to have several different error messages:
Bildschirmfoto 2020-10-20 um 16 35 33

sagararyal:
- [ ] Disable Zip/Postal code field if the country doesn't support it.

harshvitra (approved):

PR #321, opened 2020-10-14
Feature/direct gift

### LINKED ISSUE #289: Implement Direct Gift
url: /s/`username`

Direct gift increases the tree counter of the recipient through B Value (community value). 
Unlike other cases, In a direct gift, the donor knows the profile of the recipient. 

For the UI, we will need to convert - the username, to full name;

Once the validity of the username is confirmed, the username can be stored in local storage until removed by a user action.

Opening multiple links like
/support/username1 and then
/support/username2 would lead to store username2 in the list.

Screen Shot 2020-10-05 at 9 23 21 AM

--
The UI Below inherits the UI from multiple Gifts.

Screen Shot 2020-10-05 at 10 02 06 AM

Clicking `Remove` will remove the recipient from the local storage; and instead load the normal gift form.

Disabling the Gift, will simply disable gift without affecting local storage. 

When support for multiple gifts are added, the user can add additional recipients.

sarveshpro:
Fixes #289 

Changes in this pull request:
- direct gift


harshvitra:
When user clicks on cancel, remove the key from local storage

public/static/locales/en/donate.json
```diff
@@ -25,6 +25,7 @@
   "isACompanyDonation": "This is a company donation",
   "companyName": "Company Name",
   "giftRecipient": "Gift Recipient",
+  "directGiftRecipient": "This donation supports",
```
sagararyal, line 28: maybe add the string {{recipientName}}  as a part of the the translation text.

src/features/public/Donations/screens/DonationsPopup.tsx
```diff
@@ -54,7 +61,10 @@ function DonationsPopup({ onClose, project }: Props): ReactElement {
         const res = await fetch(
           `${process.env.API_ENDPOINT}/app/projects/${project.id}/paymentOptions?country=${country}`,
```
sagararyal, line 62: aren't the APIs being called from central location?

sagararyal: @harshvitra

pages/s/[id].tsx
```diff
@@ -0,0 +1,33 @@
...
+interface Props {}
+
+export default function DirectGift({}: Props): ReactElement {
+  const router = useRouter();
+  //   const [profile, setProfile] = React.useState(null);
+  React.useEffect(() => {
+    if (router.query.id !== undefined) {
+      async function loadPublicUserData() {
+        const newProfile = await getRequest(
+          `/public/v1.0/en/treecounter/${router.query.id}`
```
Comment (sagararyal) [inline pages/s/[id].tsx:14]: Should use a new profile api; also fetch the api using the aggregate function;

And add tenant key and session id.

sagararyal:
Besides the minor change suggestions, everything else looks good!

pages/s/[id].tsx
Comment (harshvitra) [reply pages/s/[id].tsx:14]: This is already using tenant key and session id

We cannot use new profile API it is not yet ready

src/features/public/Donations/screens/DonationsPopup.tsx
harshvitra: This is resolved

harshvitra (approved):

PR #325, opened 2020-10-15
Feature/pwa

harshvitra:
Creating PWA + offline support

- [ ] 1 error to resolve - Unsupported File type


next.config.js
```diff
@@ -106,4 +107,31 @@ module.exports = withSourceMaps({
...
+  },
+  async rewrites() {
+    return [
+      {
+        source: '/service-worker.js',
+        destination: '/service-worker.js',
+      },
+    ]
+  }
 });
```
norbertschuler, line 135: Where is the `service-worker.js` file? I cannot find it in this PR, so does my browser:
```
SW registration failed:  DOMException: 
Failed to register a ServiceWorker for scope ('https://planet-webapp-mwjjcg90l.vercel.app/') with script 
('https://planet-webapp-mwjjcg90l.vercel.app/service-worker.js'): 
The script has an unsupported MIME type ('text/html').
```

norbertschuler (changes requested):

PR #330, opened 2020-10-17
Feature/country translation

### LINKED ISSUE #270: Country Name Translations
Country Name should be localized [low priority]

Screen Shot 2020-09-23 at 9 15 27 AM

norbertschuler:
Fixes #270

Changes in this pull request:
- added country translation files (a few keys are not available from my source https://stefangabos.github.io/world_countries/, where I just keep the English names from https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/develop/src/utils/countryCurrency/countriesData.json if no translation is available)
- changed default language to English (in tenant config files)


public/tenants/planet/config.ts
```diff
@@ -11,7 +11,7 @@ const config = {
     secondaryFontFamily: '"Open Sans",Helvetica,Arial,sans-serif',
     secondaryFontURL: `${process.env.CDN_URL}/media/fonts/opensans/open-sans.css?v1.0`,
   },
-  languages: ['de', 'en'],
+  languages: ['en', 'de'],
```
norbertschuler, line 14: First language in this array becomes the default language!

src/features/common/Layout/Footer/SelectLanguageAndCountry.tsx
```diff
@@ -136,7 +129,10 @@ export default function TransitionsModal(props) {
 
 // Maps the radio buttons for countries
 function MapCountry(props) {
-  const { sortedCountriesData, value, handleChange } = props;
+  const { t } = useTranslation(['country']);
+  
+  const { value, handleChange } = props;
+  const sortedCountriesData = sortCountriesByTranslation(t);
```
norbertschuler, line 135: This add some unnecessary sorting to the app at runtime, but I cannot do the sorting on the translated country names before i18n is initialized.

src/features/common/Layout/Footer/index.tsx
```diff
@@ -75,6 +65,10 @@ export default function Footer() {
         countryCode = localStorage.getItem('countryCode');
         if (countryCode) setSelectedCountry(countryCode);
       }
+      if (localStorage.getItem('language')) {
+        langCode = localStorage.getItem('language');
+        if (langCode) setLanguage(langCode);
```
norbertschuler, line 70: Loading from local storage as other wise the language shown in the footer is the default `English` if i18n was not yet initialized.

public/tenants/stern/config.ts
```diff
@@ -1,7 +1,7 @@
 const config = {
   tenantName: 'stern',
   tenantURL: 'baeume.stern.de',
-  languages: ['de', 'en'],
+  languages: ['en', 'de'],
```
harshvitra, line 4: We might want to keep de as default for Stern

harshvitra (approved):

harshvitra (approved):

public/tenants/stern/config.ts
norbertschuler: In that case we would need to add some countries in `de/country.json` as they only exists in `en/country.json` for now and therefore Stern would show the placeholder value if `de` in the default country an a user has chosen the German language. But I assume the country resources will now be transferred to LingoHub and I can add German translations for the remaining country names there.

PR #331, opened 2020-10-17
Allow donate button when card details fails

### LINKED ISSUE #322: Allow clicking Donate button when card details are changed after a failed donation
Screen Shot 2020-10-14 at 5 38 13 PM

This bug is present because the Donate button is disabled after clicking once in order to prevent accidental clicks that could create multiple donations.
However this has created a bug that would create an issue when a donation fails.

Currently the donate button is only enabled if the user clicks back to go to contact details page; and then continues to step 3: Payments.

captainamiedi:
Fixes #322 

Changes in this pull request:
- Add an onChange to the card element
- Use the onchange to determine when to disable or enable donate button
-

@sagararyal


harshvitra (approved):
There is an issue, but that is part of develop so we will do it separately

PR #351, opened 2020-10-24
map search

### LINKED ISSUE #198: Map Search – Bring back pins when Letters are deleted
When typing in the search bar, the map pins disappear beautifully as the search gets more precise. But when deleting letters, the map does not update until all letters are deleted. Could this be fixed so that the map is updated with every deleted letter?

Low priority

captainamiedi:
Fixes #198 

Changes in this pull request:
- search projects should reflex on map
-
-

@sagararyal


harshvitra:
Very well done @captainamiedi 

I was wondering if we could shift the projects to index instead of app, @sarveshpro what was the reason of bringing this to app ?

harshvitra (approved):

PR #361, opened 2020-10-29
Show the new API values of TPO address, fixes crash due to API change

norbertschuler:
Fixes #

Changes in this pull request:
- show the new results to TPO address + fix crash introduced with the API change

### DIFF
--- a/src/features/public/Donations/components/SingleProjectDetails.tsx
+++ b/src/features/public/Donations/components/SingleProjectDetails.tsx
@@ -52,6 +52,13 @@ function SingleProjectDetails({ project }: Props): ReactElement {
     ? getImageUrl('project', 'large', project.image)
     : '';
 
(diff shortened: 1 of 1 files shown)


harshvitra (commented):
The address is not visible

Screenshot 2020-10-29 at 8 22 14 PM

norbertschuler:
> The address is not visible
> 
> Screenshot 2020-10-29 at 8 22 14 PM

@harshvitra that's because Vercel is using `app-staging.plant-for-the-planet.org` and not `api-develop.plant-for-the-planet.org` and the patch probably is not yet pushed to the staging backend (check the API response for `tpo.address`). Should I rewrite the PR to be able to work with the new and old API (I wouldn't want to write unnecessary compatibility code)?

harshvitra (approved):

Issue #356, opened 2020-10-28
Enhance i18 support?


norbertschuler:
**Describe the feature you'd like to see included in Plant-for-the-Planet Web App.**
The current delayed initialization of i18n shows the translation resource labels first before the real translation resources are loaded and shown. I know we wanted to enhance the site loading performance by not blocking until the translation resources are shown, but maybe there is another solution?

**What are the use cases for this feature?**
Do not show ugly translation labels names.

1. For about a second showing this on https://beta.plant-for-the-planet.org/ or https://dev.pp.eco/
<img width="952" alt="Bildschirmfoto 2020-10-28 um 17 28 46" src="https://user-images.githubusercontent.com/1532418/97467594-a6181a00-1944-11eb-8929-df8c33949608.png">

2. Finally showing the real page (e.g. in German):
<img width="950" alt="Bildschirmfoto 2020-10-28 um 17 28 38" src="https://user-images.githubusercontent.com/1532418/97467616-aca69180-1944-11eb-8bb6-9b4ee1bef0e9.png">

sarveshpro:
Heather is working on this issue.

CallmeHongmaybe:
Should we close the issue ? Bc I don't see this problem happening anymore

norbertschuler:
> Should we close the issue ? Bc I don't see this problem happening anymore

The issue will be closed automatically after the linked pull request got merged.

PR #364, opened 2020-10-30
fix/i18n-initialization

hsong124:
Changes in this pull request:
- make sure text in home page waits for i18next translation to load before showing text

Taken from this issue: 
https://github.com/Plant-for-the-Planet-org/planet-webapp/issues/356

### DIFF
--- a/src/features/common/Layout/CookiePolicy/index.tsx
+++ b/src/features/common/Layout/CookiePolicy/index.tsx
@@ -7,6 +7,13 @@ export default function CookiePolicy() {
   const [showCookieNotice, setShowCookieNotice] = useState(true);
   const { useTranslation } = i18next;
   const { t } = useTranslation(['leaderboard']);
+
+  const [initialized, setInitialized] = React.useState(false);
+
+  React.useEffect(() => {
+    i18next.initPromise.then(() => setInitialized(true));
+  }, []);
+
   React.useEffect(() => {
     let prev = localStorage.getItem('cookieNotice');
     if (prev === 'false') {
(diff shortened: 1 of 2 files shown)


src/features/common/Layout/CookiePolicy/index.tsx
```diff
@@ -7,6 +7,13 @@ export default function CookiePolicy() {
   const [showCookieNotice, setShowCookieNotice] = useState(true);
   const { useTranslation } = i18next;
   const { t } = useTranslation(['leaderboard']);
+
+  const [initialized, setInitialized] = React.useState(false);
+
+  React.useEffect(() => {
+    i18next.initPromise.then(() => setInitialized(true));
+  }, []);
```
norbertschuler, line 15: I assume we cannot reuse the initialized state created here? 
https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/c21ddeee8d189a26f8ed58261759a1e15deff4a3/pages/_app.tsx#L45-L49
Are several initializations of i18n no problem?

norbertschuler:
This patch results into showing the page like this before it is finally completely loaded - it's a possible solution if there are not better one:

Bildschirmfoto 2020-10-30 um 16 56 21

src/features/common/Layout/CookiePolicy/index.tsx
norbertschuler: - I tested i18n in debug mode and it does not seem to be a problem call i18next.initPromise several times. 
- I also think it not worth the effort to reach down the initialized values by properties through the Layout into the CookiePolicy components and similar for the NavBar - maybe once we could think of using i18n as HOC?

PR #382, opened 2020-11-04
og image for user profile

### LINKED ISSUE #320: Implement OG Image generator for Profile Pages

captainamiedi:
Fixes #320 

Changes in this pull request:
- create og image for user
-
-

@


.env.development
```diff
@@ -7,5 +7,7 @@
 ## `.env.development` stays in git with default or empty values
 
 MAPBOXGL_ACCESS_TOKEN=
-API_ENDPOINT=https://app-development.plant-for-the-planet.org
-CDN_URL=
+API_ENDPOINT=app-development.plant-for-the-planet.org
+AUTH0_CUSTOM_DOMAIN=pftp.eu.auth0.com
+AUTH0_CLIENT_ID=yxABVrSIbmSO6QeTfZjkJefDnW2vrE7y
```
norbertschuler, line 12: Is this an information we want to make public?

norbertschuler (changes requested):
This patch probably is based on `Feature/login`? But this PR is configured to show the difference to `develop` right now.

.env.development
captainamiedi: I have remove it. these feature depends on the login that's why i had to branch off feature/login

norbertschuler: Ok, but as that branch is now already merged into develop, I would suggest you should rebase your branch to develop as well. (`git rebase develop`).

sagararyal:
Text and info in the og flag has to be rendered server side.

The problem with current implementation is that web scrapers cannot see the og:description or og:image as react needs to load before it can be created.

norbertschuler:
> Text and info in the og flag has to be rendered server side.

I thought we deploy our production website as a static site. Can you do server side rendering with a static Nextjs deployment?

sagararyal:
> I thought we deploy our production website as a static site. Can you do server side rendering with a static Nextjs deployment?

@norbertschuler Correct we do serve as a static site. And there's no problem with this approach. However if og images loaded with static approach, most crawlers will never get to see them.

This is why I suggested we separate the creation part to either a worker, or a server side rendering function and simply pass a url. 

@sarveshpro can probably add more as we already use some meta and tags in the app that need dynamic content on the Donate Page.

sarveshpro:
For meta tags to work they should be present in the initial render document sent from the server. to check this, view 
`view-source:https://dev.pp.eco/t/sagar-aryal` this is what a crawler bot sees. the meta tags should be present in this code for SEO to work.

next.js tends to render react/javascript code serverside unless it depends on local storage or an API call happening on client-side. right now all our API calls happen client-side. The only reason why SEO works for the homepage is that the data it returns is already present in the tenant config files. To fix SEO on other pages with dynamic data we will have to make API calls using serverless functions or maybe swr(I don't know much about swr) this way calls are made serverside and entire page is rendered serverside.

Step 1:
Make the `/app/profiles/${slug}` API call serverside to get user data, get static map image with planted markers using Mapbox static map API(also serverside), use the URL which Mapbox returns as the og:image.

Step 2:
generate a new image with the static map as background, user's profile image on top-right and treecounter bar at the bottom, somehow store it temporarily and use the URL.

sagararyal:
> Make the `/app/profiles/${slug}` API call serverside to get user data, get static map image with planted markers using Mapbox static map API(also serverside), use the URL which Mapbox returns as the og:image.

I would avoid this for every tenant, as with increasing number of tenants we would be missing resources. I would instead suggest an alternative where og:image is generated either on a main domain, separate project or Cloudflare worker. This way all installations can refer to one single image. When cached, the calls to both map box and backend is reduced significantly.

PR #392, opened 2020-11-05
Clean URLs

sagararyal:
@harshvitra I think the page can further be cleaned to remove unnecessary conditions.

Please check.


harshvitra:
> @harshvitra I think the page can further be cleaned to remove unnecessary conditions.

Can the other images be fetched with same CDN URL ?

harshvitra (approved):

PR #424, opened 2020-11-11
Tenant/energizer

sagararyal:
@harshvitra 
In this PR I've also tried to simplify the tenant setup by adding some config params into tenant config. We can still simplify more. Further maybe add a possibility of including localized content.

- [x] Logo

- [ ] Favicons

- [x] Final Texts


norbertschuler (approved):
Works for me.

PR #437, opened 2020-11-12
Feature/redeem urls

### LINKED ISSUE #405: Redeem URLs
Create pages for Redeem URLs

User's will use link pp.eco/CODE -
Fetch data from API using this code and
redirect them to 
URL - tenantUrl/redeem/CODE

Different designs for 
- Planted trees
- Gifted trees
- Organisation gifts

Different for logged in users

harshvitra:
Fixes #405 

Added Redeem page with dummy data. 

### **To do -** 

- [ ] Integrate API  (backend needed)
- [x] Enable Sign in
- [x] Enable Signup
- [ ] Enable Add to my profile (backend needed)
- [x] Add map
- [x] Add markers
- [ ] Profile image or company image (backend needed)


harshvitra:
When we talk about sign in and Sign up it is the same with respect to Auth0 they both are the same

sagararyal:
@harshvitra  yes, I think you can pass the signin parameter with the callback  or return to Url so user comes back to the redeem page after either of the action.

mariahosfeld:
Will be implemented in different way (#1568 & #1554)

PR #441, opened 2020-11-12
Hotfix: TPO name on donation form

sarveshpro:
Changes in this pull request:
- removed lower case filter from TPO name


norbertschuler (commented):
This fix missed to change two more places in our source code:

Bildschirmfoto 2020-11-12 um 14 19 55

PR #449, opened 2020-11-14
fix responsiveness in Tree donate for small screens

### LINKED ISSUE #443: Handle responsiveness of Tree count options
On smaller screens the design breaks

Screen Shot 2020-11-11 at 8 26 51 PM

captainamiedi:
Fixes #443 

Changes in this pull request:
-
-
-

@sagararyal @harshvitra


harshvitra (changes requested):
@captainamiedi please re-iterate the media queries. 

Use the mixins created in theme. 

Also if the css is same for various media queries - only write the changing properties 
In this case that is the width of the card container and padding top of the header.

src/features/public/Donations/components/TreeDonation.tsx
```diff
@@ -123,7 +130,7 @@ function TreeDonation({
     <>
       <div
         className={styles.cardContainer}
-        style={{ alignSelf: isGift ? 'start' : 'center' }}
+        style={{ alignSelf: isGift ? 'start' : 'center', width: `${screenWidth}px` }}
```
harshvitra, line 133: Because of the width specified here the design breaks on the screens with size < 412. 

Screenshot 2020-11-19 at 11 02 19 AM

harshvitra (changes requested):

src/features/public/Donations/components/TreeDonation.tsx
harshvitra: If the screen width is below 412, set it to 100% width

captainamiedi: done.

harshvitra (approved):

PR #471, opened 2020-11-19
added headers to protect against iframes

norbertschuler:
Fixes #

Changes in this pull request:
- added headers to protect against iframes


norbertschuler:
Does not seem to be testable on Vercel.

PR #481, opened 2020-11-23
added sentry env variables

norbertschuler:
Changes in this pull request:
- added sentry env variables

These are the env variables which always are the same.


norbertschuler:
From the build log at https://vercel.com/planetapp/planet-webapp/70muylrw4 I would say it still works with Vercel. For veryfying the source maps got uploaded to Sentry we would have to wait a little bit and then check https://sentry.io/settings/plant-for-the-planet/projects/app-20-next/source-maps/

PR #486, opened 2020-11-24
fix country not found on single project

### LINKED ISSUE #484: TypeError: w is null
Sentry Issue: [APP-20-NEXT-8Z](https://sentry.io/organizations/plant-for-the-planet/issues/2050814548/?referrer=github_integration)

```
TypeError: w is null
  at c (webpack://_N_E/./src/features/public/Donations/components/TreeDonation.tsx:265:15)
  at Ch (webpack://_N_E/./node_modules/react-dom/cjs/react-dom.production.min.js:157:137)
  at ck (webpack://_N_E/./node_modules/react-dom/cjs/react-dom.production.min.js:267:460)
  at bk (webpack://_N_E/./node_modules/react-dom/cjs/react-dom.production.min.js:250:347)
  at ak (webpack://_N_E/./node_modules/react-dom/cjs/react-dom.production.min.js:250:278)
...
```

captainamiedi:
Fixes #484 

Changes in this pull request:
-
-
-

@sagararyal

### DIFF
--- a/src/features/projects/screens/SingleProjectDetails.tsx
+++ b/src/features/projects/screens/SingleProjectDetails.tsx
@@ -11,6 +11,7 @@ import CancelIcon from '../../../../public/assets/images/icons/CancelIcon';
 import ExpandIcon from '../../../../public/assets/images/icons/ExpandIcon';
 import ProjectInfo from '../components/projectDetails/ProjectInfo';
 import ProjectSnippet from '../components/ProjectSnippet';
+import storeConfig from '../../../utils/storeConfig';
 
 const { useTranslation } = i18next;
 interface Props {
@@ -39,6 +40,7 @@ function SingleProjectDetails({ project }: Props): ReactElement {
 
   React.useEffect(() => {
     window.scrollTo(0, 0);
+    storeConfig();
   }, []);
 
   const [openModal, setModalOpen] = React.useState(false);


src/features/projects/screens/SingleProjectDetails.tsx
```diff
@@ -39,6 +40,7 @@ function SingleProjectDetails({ project }: Props): ReactElement {
 
   React.useEffect(() => {
     window.scrollTo(0, 0);
+    storeConfig();
```
norbertschuler, line 43: I do not yet know React that well. Is this code executed before other components are loaded which might try to load local storage items? Or do we maybe get a race condition where getting the item from the local storage is executed before this code has the chance to store the item in the local storage?

norbertschuler: If we cannot guaranty that local storage is existing, all component trying to make use a such values just have to be build more robust, i.d. must not crash due to null values and render again once the item is available. Or we have to find another architecture to prevent component rendering until the config is loaded? (I think we already have the same problem with i18n where a lot of components get rendered before i18n initialization is finished.)

captainamiedi: The API is called before the DOM is completely render. so these will enable the country code to be stored on the local storage

norbertschuler: Ok, loading and storing the config is also triggered here https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/e46776ad0d7d825aca5288822a5176a3ce0a0e8a/pages/index.tsx#L34-L36 - probably this is not loaded when someone is opening the website directly with an URL like https://dev.pp.eco/agroforestry-to-assist-forest-restoration-in-the-dry-bolivian-andes - is your code replacing the about loading of the config or is there maybe a better general place to load the config which is always triggered despite which page is loaded?

norbertschuler: What about doing it here at https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/develop/pages/_app.tsx instead of the other two places?

captainamiedi: The code replaces the about loading. since the user didn't visit the about page, then these code will do the same function as about page

norbertschuler: To better explain what I meant, I create a new PR https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/487

norbertschuler: One more point: @harshvitra I am not sure, what exactly you planned with _components_ and _screens_, but if screens should be pure UI functions as in classic [MVC](https://de.wikipedia.org/wiki/Model_View_Controller) architecture, I would think that API calls should not be triggered within screens, right?

PR #489, opened 2020-11-25
Edit Profile fixes

captainamiedi:
Fixes #

Changes in this pull request:
-
-
-

@sagararyal


norbertschuler:
@captainamiedi can you link the issue which is fixed by this or add a small description was this is fixing?

src/features/user/UserProfile/components/AddTargetModal.tsx
```diff
@@ -67,9 +69,14 @@ export default function AddTargetModal({
               defaultValue={userprofile.score.target ? userprofile.score.target : null}
               onChange={(e) => setTarget(e.target.value)}
               variant="outlined"
+              inputRef={register({
+                min: 0
```
norbertschuler, line 73: The backend declines zero as a possible target - so either the backend or the client should be changed:

Bildschirmfoto 2020-11-26 um 13 36 48

src/features/user/UserProfile/components/Settings.tsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react'
```
norbertschuler, line 1: You should use an editor which can automatically nicely format our source code as in the file before.

src/features/user/UserProfile/components/AddTargetModal.tsx
norbertschuler: Another issue: When I enter -100 as a value, I get this view with the string "invalid" after I clicked on the save button before the dialog closed automatically after the API call failed and my target does not get changed: 
Bildschirmfoto 2020-11-26 um 13 35 06

norbertschuler: I would suggest to show the user an error message if the API call fails and the target can not be changed.

norbertschuler (changes requested):

norbertschuler:
Tested again, but this still don't work for me (latest MacOS with latest Chrome). I neither see any loader when clicking on the save button for the new target value nor do I get an error message if I enter 0 or -100 as a value. Also the arrows allow me to decrease the value down to 0 instead only down to 1.

norbertschuler (approved):
LGFM

src/features/user/UserProfile/screens/IndividualProfile.tsx
```diff
@@ -113,11 +115,11 @@ export default function IndividualProfile({ userprofile, changeForceReload,
...
-        variant="filled"
-        onClose={handleTextCopiedSnackbarClose} 
-        severity="success"
+        <MuiAlert
+          elevation={6}
+          variant="filled"
+          onClose={handleTextCopiedSnackbarClose}
+          severity="success"
         >
           Text Copied to Clipboard!
```
sagararyal, line 124: @captainamiedi should this be translated?
@harshvitra do we still plan to use this alerts?

sagararyal (commented):
.

src/features/user/UserProfile/screens/IndividualProfile.tsx
captainamiedi: These has been replaced with the share functionality.

sagararyal: @captainamiedi Can you clean up the code?

captainamiedi: ok

PR #524, opened 2020-12-01
Fix issues with currency selector

### LINKED ISSUE #520: Issues with country/currency selector
This is present in develop
Screen Shot 2020-11-30 at 11 18 16 PM

sarveshpro:
Fixes #520

Changes in this pull request:
- used React Context API to store and update user currency

context created in `src/utils/Context/currencyContext.js`
`currencyContext.Provider` loads currency stored in LocalSorage at first render (fallback to USD by default)
this value can be used/updated from anywhere inside the Provider's children using `useContext`
if currency value is updated, new tree cost is fetched from backend

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -5,6 +5,7 @@ import GetAllProjectsMeta from '../src/utils/getMetaTags/GetAllProjectsMeta';
 import getStoredCurrency from '../src/utils/countryCurrency/getStoredCurrency';
 import { getRequest } from '../src/utils/apiRequests/api';
 import DirectGift from '../src/features/donations/components/treeDonation/DirectGift';
+import currencyContext from '../src/utils/Context/CurrencyContext';
 
 interface Props {
   initialized: Boolean;
@@ -14,7 +15,7 @@ interface Props {
   setShowSingleProject: Function;
   showProjects: Boolean;
   setShowProjects: Function;
-  setsearchedProjects: any
+  setsearchedProjects: any;
 }
 
 export default function Donate({
@@ -28,6 +29,7 @@ export default function Donate({
   setsearchedProjects,
 }: Props) {
   const router = useRouter();
+  const { currency } = React.useContext(currencyContext);
   const [directGift, setDirectGift] = React.useState(null);
   const [showdirectGift, setShowDirectGift] = React.useState(true);
 
@@ -60,14 +62,14 @@ export default function Donate({
     async function loadProjects() {
       const currencyCode = getStoredCurrency();
       const projects = await getRequest(
-        `/app/projects?_scope=map&currency=${currencyCode}`,
+        `/app/projects?_scope=map&currency=${currency}`
       );
       setProjects(projects);
       setProject(null);
       setShowSingleProject(false);
     }
     loadProjects();
-  }, []);
+  }, [currency]);
   const ProjectsProps = {
     projects,
     showProjects,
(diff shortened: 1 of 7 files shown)


pages/index.tsx
```diff
@@ -60,14 +62,14 @@ export default function Donate({
     async function loadProjects() {
       const currencyCode = getStoredCurrency();
```
norbertschuler, line 63: I assume this line (and the import of getStoredCurrency ) is not necessary anymore, is it?

norbertschuler (changes requested):
Should we also rewrite https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/d2c3e52359f41662b9c571c03a0db1763df7a950/pages/%5Bp%5D.tsx#L27-L30 as you did it with `index.tsx`?

norbertschuler:
Does the code changing the currency in our localStorage also have to change the context value now at https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/d2c3e52359f41662b9c571c03a0db1763df7a950/src/features/common/Layout/Footer/SelectLanguageAndCountry.tsx#L56-L70?

pages/index.tsx
```diff
@@ -5,6 +5,7 @@ import GetAllProjectsMeta from '../src/utils/getMetaTags/GetAllProjectsMeta';
 import getStoredCurrency from '../src/utils/countryCurrency/getStoredCurrency';
 import { getRequest } from '../src/utils/apiRequests/api';
 import DirectGift from '../src/features/donations/components/treeDonation/DirectGift';
+import currencyContext from '../src/utils/Context/CurrencyContext';
```
norbertschuler, line 8: Isn't this creating a different context as the one created in _app.tsx?

pages/index.tsx
norbertschuler: But before you change this, please consider my comment https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/524#discussion_r534021348 below.

norbertschuler:
- [ ] This PR does not yet work for me. I am still getting different currencies (and languages) displayed looking at the small footer on the homepage compared with the bigger footer on other pages, e.g. profile page. This probably needs to be thoroughly tested and all changed to use the same overall context and keep it in sync with the localStorage and the currency defined in the user profile.

src/utils/Context/CurrencyContext.js
```diff
@@ -0,0 +1,8 @@
+import { createContext } from 'react';
+
+const currencyContext = createContext({
+  currency: 'USD',
```
norbertschuler, line 4: We usually use 'EUR' as the default currency if nothing was defined before.

pages/index.tsx
sarveshpro: Yes removed it. and rewrote with context in `planet-webapp/pages/[p].tsx`

pages/index.tsx
sarveshpro: No, context is created in the context file `src/utils/Context/CurrencyContext.js`, a provider is created in `_app.tsx` so that the context is accessible anywhere in the app.

sarveshpro: then you can use `useContext` to get the value and the setter function.

norbertschuler: Ok, it's a singleton pattern. Sorry, sometimes I am still getting confused about this.

norbertschuler:
@sarveshpro this PR does not solve the following issue, does it? https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/531#issuecomment-737167509

norbertschuler:
@sarveshpro please fix merge errors or close if we do not want to introduce context yet. Maybe this PR could solve #609 once the question is clarified?

PR #498, opened 2020-11-27
Feature/react auth0

harshvitra:
Fixes #507

Replaces Auth0 Setup

TODO:
- [x] On Vercel still login and logout are a problem (maybe just because of the dynamic Vercel URLs) https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/498#issuecomment-734824586 and https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/498#issuecomment-735236715

PR #528, opened 2020-12-01
added session to payDonation function calls

norbertschuler:
Fixes failing Paypal payments for logged in users.

Changes in this pull request:
- added session to payDonation function calls

TODO:
- [x] check if missing `session` in https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/develop/src/features/legacyDonations/index.tsx is a problem (e.g. if the user is accidentally logged in in the web browser opened by the native app) -> **We won't add the Authorization header with the session to the API call of the legacy donations as that would require the users of the native apps to login into their smartphone browsers if they are logged in on the native apps.**
- [x] check and solve these errors also in https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/498

@harshvitra

### DIFF
--- a/src/features/donations/screens/PaymentDetails.tsx
+++ b/src/features/donations/screens/PaymentDetails.tsx
@@ -219,7 +219,7 @@ function PaymentDetails({
       paymentMethod,
       donorDetails,
       taxDeductionCountry: isTaxDeductible ? country : null,
-      token:token?token : null
+      token: token
     };
     payWithCard({ ...payWithCardProps });
   };
--- a/src/features/donations/screens/PaymentDetails.tsx
+++ b/src/features/donations/screens/PaymentDetails.tsx
@@ -307,7 +307,7 @@ function PaymentDetails({
       }
     }
 
-    createDonation(createDonationData, session)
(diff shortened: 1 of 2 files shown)


norbertschuler:
@harshvitra please feel free to take over this PR!

norbertschuler:
@harshvitra the backend team removed the authentication mismatch protection - so this patch theoretically is not necessary any more. But I would feel better to keep it because it's just feels the right way to do :-)

norbertschuler:
@harshvitra I refactored this patch and still would like to merge it as I think it also fixes a bug.

src/features/donations/screens/PaymentDetails.tsx
```diff
@@ -307,7 +307,7 @@ function PaymentDetails({
       }
     }
 
-    createDonation(createDonationData, session)
+    createDonation(createDonationData, token)
```
norbertschuler, line 310: There is no more `session`, but it worked as interpreted as `null` I assume.

harshvitra (approved):

PR #531, opened 2020-12-02
Fix country/currency selector on homepage

### LINKED ISSUE #520: Issues with country/currency selector

norbertschuler:
Fixes #520

Changes in this pull request:
- make sure right values are loaded from localStorage before showing dialog to change language and country currency

@

### DIFF
--- a/src/features/common/Layout/Footer/index.tsx
+++ b/src/features/common/Layout/Footer/index.tsx
@@ -50,23 +50,19 @@ export default function Footer() {
     },
   ];
 
-  // changes the language and selected currency id found in local storage
+  // changes the language and selected country as found in local storage
   useEffect(() => {
-    let langCode;
-    let currencyCode;
-    let countryCode;
-
     if (typeof Storage !== 'undefined') {
       if (localStorage.getItem('currencyCode')) {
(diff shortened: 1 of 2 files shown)


norbertschuler:
@sarveshpro quick solution without introducing context as a quick fix as refactoring to using context might be more complex.

captainamiedi:
> @sarveshpro quick solution without introducing context as a quick fix as refactoring towards using context as in #524 might be more complex.

I think using context is a lot easier and less code to make it work. if these change is needed else where in the application, you don't need to write these much again. Context is not too much of a complex solution

norbertschuler:
> I think using context is a lot easier and less code to make it work. if these change is needed else where in the application, you don't need to write these much again. Context is not too much of a complex solution

O.k., I understood, but #524 does not yet fix the issue #520 which I understood is the preselection of the country 'US' (which is our default country) if your open the Language/Currency chooser from the map while in the localStorage a different countryCode is already defined:

Bildschirmfoto 2020-12-02 um 12 14 05

(This patch solves it.)

harshvitra (approved):

PR #586, opened 2020-12-05
added robots.txt

norbertschuler:
Fixes #

Changes in this pull request:
- added robots.txt

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Disallow: /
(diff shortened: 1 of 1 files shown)


sagararyal:
we don't want to do this.
This will also prevent googlebot and other search engine from crawling the pages.

harshvitra:
> we don't want to do this.
> This will also prevent googlebot and other search engine from crawling the pages.

We don't have a sitemap right now. 
Which will lead to huge negative score in rankings. 

So just some more time and we'll remove the robot file

PR #617, opened 2020-12-09
use ESRI basemap

sarveshpro:
Changes in this pull request:
- change basemap and satellite map to ESRI


src/features/projects/components/MapboxMap.tsx
```diff
@@ -362,7 +416,7 @@ export default function MapboxMap({
         mapboxApiAccessToken={mapboxToken}
```
norbertschuler, line 416: I see all network requests changed from mapbox to arcgis, so I wonder if the requests still count on our mapbox account as the token is still configured here?

sagararyal:
@norbertschuler Mapbox is still used in register tree and profile maps.

Lets approve this and we make other changes later.

PR #620, opened 2020-12-09
add planet logo with about us

sarveshpro:
Changes in this pull request:
- Planet logo with about us text


norbertschuler:
The logo and the fine lines in the new logo do not match the other navigation items. It does not look good to me :-)

Bildschirmfoto 2020-12-09 um 19 01 27

norbertschuler:
How do we localize the text "About us" in the logo? Do we have a logo in different languages?

sarveshpro:
@norbertschuler is it better now?

norbertschuler:
Localization should be possible now 👍 

@sarveshpro I just compare the top navi: The color of your new logo is different now and there is no padding at the top - what do you think?

1. https://www.trilliontreecampaign.org/
Bildschirmfoto 2020-12-10 um 10 18 15

2. This PR:
Bildschirmfoto 2020-12-10 um 10 18 28

3. https://a.plant-for-the-planet.org/
Bildschirmfoto 2020-12-10 um 10 18 36

sarveshpro:
the color is greyscale because the page is not active. the about us logo should be larger. I can add more padding but Felix had suggested to make the logo as big as possible.

image

norbertschuler:
> the color is greyscale because the page is not active

Ok, do we want this item to be active/inactive - currently it never can get active as the we open another website like for every tenant ;-)

sarveshpro:
It will always be inactive as far as I know. It's active only on a.plant-for-the-planet.org

norbertschuler (approved):

PR #621, opened 2020-12-09
fixed sharing if no user name is available

### LINKED ISSUE #607: TypeError: Cannot read property 'firstName' of undefined
Sentry Issue: [APP-20-NEXT-C9](https://sentry.io/organizations/plant-for-the-planet/issues/2075271692/?referrer=github_integration)

```
TypeError: Cannot read property 'firstName' of undefined
  at c (./src/features/donations/components/ShareOptions.jsx:23:41)
  at None (/_next/static/chunks/pages/t/%5Bid%5D-a8ebdfda6df7b8077f9d.js:1:47729)
...
```

norbertschuler:
Fixes #607

Changes in this pull request:
- fixed sharing if no user name is available

I know this file is .jsx and not .tsx yet, but I had no time to rewrite it. So I add least added prop-types.

### DIFF
--- a/src/features/donations/components/ShareOptions.jsx
+++ b/src/features/donations/components/ShareOptions.jsx
@@ -1,4 +1,7 @@
 import React, { useRef } from 'react';
+import ReactDOM from 'react-dom';
+import domtoimage from 'dom-to-image';
+import PropTypes from 'prop-types';
 import styles from './../styles/ThankYou.module.scss';
 import EmailIcon from '../../../../public/assets/images/icons/share/Email';
 import EmailSolid from '../../../../public/assets/images/icons/share/EmailSolid';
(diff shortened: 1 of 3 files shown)


harshvitra:
@norbertschuler I just need a code for testing claim trees and then I have another solution for this problem.

src/features/donations/components/ShareOptions.jsx
```diff
@@ -145,4 +151,11 @@ const ShareOptions = (props) => {
   ) : null;
 };
 
+ShareOptions.propTypes = {
+  treeCount: PropTypes.string.isRequired,
+  sendRef: PropTypes.node.isRequired,
+  handleTextCopiedSnackbarOpen: PropTypes.func,
+  contactDetails: PropTypes.node
```
captainamiedi, line 158: Prop types of node means we can pass anything that can be rendered, such as numbers, string, DOM elements, arrays, or fragments. why not use array since we are expecting an array.

src/features/donations/components/ShareOptions.jsx
```diff
@@ -8,20 +11,23 @@ import DownloadIcon from '../../../../public/assets/images/icons/share/Download'
...
 
   const titleToShare = ready ? t('donate:titleToShare') : '';
   const urlToShare = config.tenantURL;
   const linkToShare = config.tenantURL;
-  const userName = props.contactDetails.firstName + ' ' + props.contactDetails.lastName;
-  const textToShare = ready ? t('donate:textToShareLinkedin', { name: userName}) : '';
+  let textToShare = '';
+  // props.contactDetails may be undefined or empty for legacy donations or redeem
+  if (props.contactDetails && (props.contactDetails.firstName || props.contactDetails.lastName)) {
+    textToShare = ready ? t('donate:textToShareLinkedin', { name: `${props.contactDetails.firstName} ${props.contactDetails.lastName}` }) : '';
```
captainamiedi, line 27: if props.contactDetails.firstname is null or props.contactDetails.lastName is now, i think these may also crash as well.

src/features/donations/components/ShareOptions.jsx
norbertschuler: Ok, I wasn't sure what kind of object contactDetails might be.

src/features/donations/components/ShareOptions.jsx
norbertschuler: I assumed an undefined value would be parsed into an empty string, so there is only a space too much showing.

norbertschuler:
will add this into #622

PR #637, opened 2020-12-11
Gift type invitation

harshvitra:
Fixes gift invitation issue where the gift details were not being passed to the API


src/features/donations/components/treeDonation/GiftForm.tsx
```diff
@@ -72,5 +80,5 @@ export default function GiftForm({
         </div>
       </div>
     </div>
-  ) : null;
+  ) : <></>;
```
norbertschuler, line 83: I put this `null` value into very many components. Do we have to replace it with  `<></>` everywhere (>70 times in our codebase)?

PR #640, opened 2020-12-11
Feature/stripe method

harshvitra:
Dynamically loading stripe method from the payment setup API. 

- [x] Change gateway to stripe


harshvitra:
It fails for develop server

Screenshot 2020-12-11 at 4 21 47 PM

PR #659, opened 2020-12-15
Tenant/andalusia

harshvitra:


public/tenants/andalusia/config.ts
```diff
@@ -0,0 +1,64 @@
+const config = {
+  tenantName: 'andalusia',
+  tenantURL: 'wald.energizer.de',
```
norbertschuler, line 3: URL and text in this file aren't yet swapped from Energizer :-)

harshvitra: currently this branch is broken, need to rewrite the Navbar which can break other branches too

harshvitra:
Closing the PR and keeping the branch

PR #663, opened 2020-12-15
Fix/basemap not loading

sarveshpro:
Fixed map loading issue


norbertschuler (approved):
On my 📱 it works well with Safari, Chrome or Firefox.

PR #680, opened 2020-12-17
Netlify Integration

sagararyal:


package.json
```diff
@@ -8,6 +8,7 @@
   "scripts": {
     "dev": "next dev",
     "build": "next build",
+    "export": "next export",
```
norbertschuler, line 11: Latest additions of redirects might make problems here if you look at the message when running this:
```
warn  - rewrites, redirects, and headers are not applied when exporting your application, detected (redirects). See more info here: https://err.sh/next.js/export-no-custom-routes
```

norbertschuler: I tested this by generating the static files and run a [webserver](https://css-tricks.com/snippets/html/start-a-web-server-with-one-terminal-command-on-os-x/) with `php -S localhost:3000` in the created `out` folder. I was not able to call the redirects e.g. http://localhost:3000/project/1 or http://localhost:3000/competition/137 in this deployment. Indead I was not able to call any route e.g. http://localhost:3000/all - @sagararyal do you already have testing this on Netlify?

PR #698, opened 2020-12-23
adapt currency

### LINKED ISSUE #667: Adapt tree prices after currency change
**Adapt tree prices after currency change**
After you switched the currency the prices for the trees are still the same.

Production 

**To Reproduce**
1. Go to https://www.trilliontreecampaign.org/
2. Switch the currency
4. See wrong price labeling

**Expected behavior**
Show prices in the selected currency

**What actually happens**
prices do not change

**Screenshots**
*If applicable, add screenshots to help explain your problem.*

Browser Chrome & Firefox

**Additional context**
Maybe you can just reload the page after the user switched the currency?

captainamiedi:
Fixes #667 

Changes in this pull request:
- when currency changes, the api is refetched for both single project and all projects

@sagararyal

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3779,6 +3779,15 @@
       "resolved": "https://registry.npmjs.org/binary-extensions/-/binary-extensions-2.1.0.tgz",
       "integrity": "sha512-1Yj8h9Q+QDF5FzhMs/c9+6UntbD5MkRfRwac8DoEm9ZfUBZ7tZ55YcGVAzEe4bXsdQHEk+s9S5wsOKVdZrw0tQ=="
     },
+    "bindings": {
+      "version": "1.5.0",
+      "resolved": "https://registry.npmjs.org/bindings/-/bindings-1.5.0.tgz",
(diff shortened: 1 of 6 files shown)


norbertschuler (commented):
- [ ] There is still a problem on first load of the main page showing the default currency some times as two API calls get triggered (one with the chosen currency and one without):

Bildschirmfoto 2020-12-30 um 12 27 55
Bildschirmfoto 2020-12-30 um 12 36 46

norbertschuler:
- [ ] Triggering more than one API calls to load the projects on the pages http://localhost:3000 and http://localhost:3000/yucatan-reforestation is also a load issue for the backend. We should write the code the way only one API call is necessary!

norbertschuler (changes requested):
- [ ] Triggering more than one API call to load the projects on the pages http://localhost:3000 and http://localhost:3000/yucatan-reforestation is a load issue for the backend. We should write the code the way only one API call is necessary!

norbertschuler (approved):
I tried to fix the duplicate API calls and resulting errors by maintaining an internal currency state also which you can compare to the _global_ state.

PR #522, opened 2020-12-01
testing setup

captainamiedi:
Fixes #

Changes in this pull request:
- start testing on Tree donate component. Test is not comprehensive enough

- [x] Improving testing
- [x] End all test connection on browser stack

To-Do's:

- [x] https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/522#discussion_r545785592
- [x] https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/522#discussion_r546755539
- [x] https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/522#discussion_r547885939

@sagararyal

PR #718, opened 2021-01-04
added basic test setup with jest and selenium web driver on BrowserStack

norbertschuler:
Fixes #522 (all the issue I have mentioned)

Changes in this pull request:
- doing some donation tests
- making changes in the database, so only run on test systems
- depends on English language (not always using test-ids for elements)
- only testing with latest Chrome on Windows
- requires .env file to exists with BrowserStack credentials


sagararyal (approved):
Looks good. Need to solve a package conflict.

PR #732, opened 2021-01-11
load user info from server on login

### LINKED ISSUE #720: Wrong userInfo after login with Auth0
**Describe the bug**
Sometimes a user got logged out, but in the browser's localStorage the `userInfo` value is still stored. After the user logged in again with Auth0 with another account, this value of `userInfo` in the localStorage does not get updated, so e.g. the Navbar component is still showing the wrong user image.

production.

**To Reproduce**
1. Find a situation where you are logged out without doing it yourself (as then the userInfo value is deleted)
2. Login again
3. See error

**Expected behavior**
Update the `userInfo` value with an API call everytime a new login with Auth0 is done.

**What actually happens**
`userInfo` is not updated.

**Screenshots**
Screen Shot 2020-12-23 at 1 38 52 PM

- OS: latest MacOS
- Browser: latest Chrome

norbertschuler:
Fixes #720

Changes in this pull request:
- remove userInfo in callback from Auth0 to force new load from API <- this is the fix
- only more use one place to handle 'userInfo' in localStorage
- changed duplicate getUserInfo function name
- adding missing ui_locales to loginWithRedirect

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -17,6 +17,7 @@ import Layout from '../src/features/common/Layout';
 import MapLayout from '../src/features/projects/components/MapboxMap';
 import { useRouter } from 'next/router';
 import { storeConfig } from '../src/utils/storeConfig';
+import { removeLocalUserInfo } from '../src/utils/auth0/localStorageUtils';
 
 if (process.env.NEXT_PUBLIC_SENTRY_DSN) {
   const config = getConfig();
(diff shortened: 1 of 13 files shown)


pages/_app.tsx
```diff
@@ -36,6 +37,7 @@ if (process.env.NEXT_PUBLIC_SENTRY_DSN) {
 }
 
 const onRedirectCallback = (appState) => {
+  removeLocalUserInfo();
```
norbertschuler, line 40: This is the fix, the rest of the PR is refactoring work.

harshvitra (approved):

PR #733, opened 2021-01-11
fixed plural form

norbertschuler:
Fixes Jira Issue from customer support

Changes in this pull request:
- using a plural form in i18n requires a number value for `count` - a formatted string could spoil the automatic choice of the right form, please test! This is just the number after redeeming a voucher code - for all the other strings we currently always show the plural form also for 1 tree!


norbertschuler:
If we want, we could implement plural forms for these resources as well (probably need to add a numeric `count` parameter for i18n to work):

Bildschirmfoto 2021-01-11 um 20 56 22

harshvitra (approved):

PR #735, opened 2021-01-12
format and parse the tree cost of a project in user language

### LINKED ISSUE #679: Localize number parsing
**Describe the feature you'd like to see included in Plant-for-the-Planet Web App.**
Currently we try to parse numbers for (currency) amounts with one regex for German, English and other number formats which is causing confusion and maybe errors for the users. We should parse the number according to the language the user has chosen only. 
If using the German language, the numbers are parsed in German number format, e.g. `2.505,50`, and for English users the numbers are parsed in the English number format, e.g. `2,505.50`.

**What are the use cases for this feature?**
Avoid irritations and errors for the input of numbers.

**Avoid bugs**
- API rejects `amount: "2000,50"` which I was able to enter in the frontend (`2000,50` is the German writing for `2000.50`):

Bildschirmfoto 2020-12-09 um 10 48 49

We can either choose some existing modules for a numberParser with locale support, e.g. https://github.com/BrightspaceUI/intl or https://github.com/globalizejs/globalize#number-module or (if we do not want to add too much big npm modules into our codebase) try to use our own small helper method as e.g. found here: https://stackoverflow.com/questions/55364947/is-there-any-javascript-standard-api-to-parse-to-number-according-to-locale as following:
```javascript
class NumberParser {
  constructor(locale) {
    const format = new Intl.NumberFormat(locale);
    const parts = format.formatToParts(12345.6);
    const numerals = Array.from({ length: 10 }).map((_, i) => format.format(i));
    const index = new Map(numerals.map((d, i) => [d, i]));
    this._group = new RegExp(`[${parts.find(d => d.type === "group").value}]`, "g");
    this._decimal = new RegExp(`[${parts.find(d => d.type === "decimal").value}]`);
    this._numeral = new RegExp(`[${numerals.join("")}]`, "g");
    this._index = d => index.get(d);
  }
  parse(string) {
    return (string = string.trim()
      .replace(this._group, "")
      .replace(this._decimal, ".")
      .replace(this._numeral, this._index)) ? +string : NaN;
  }
}

const formatter = new Intl.NumberFormat('ar-EG');
const parser = new NumberParser('ar-EG');
console.log(formatter.format(1234.5));               // ١٬٢٣٤٫٥
console.log(parser.parse(formatter.format(1234.5))); // 1234.5
```

norbertschuler:
Fixes #679

Changes in this pull request:
- format and parse the tree cost of a project in user language
- changed project spending to integer as backend rounds anyway

### DIFF
--- a/src/features/user/ManageProjects/components/BasicDetails.tsx
+++ b/src/features/user/ManageProjects/components/BasicDetails.tsx
@@ -13,6 +13,7 @@ import {
   putAuthenticatedRequest,
 } from '../../../../utils/apiRequests/api';
 import addServerErrors from '../../../../utils/apiRequests/addServerErrors';
+import { getFormattedNumber, parseNumber } from '../../../../utils/getFormattedNumber';
 
 const { useTranslation } = i18next;
 
@@ -130,7 +131,7 @@ export default function BasicDetails({
         website: projectDetails.website,
         description: projectDetails.description,
         acceptDonations: projectDetails.acceptDonations,
-        treeCost: projectDetails.treeCost,
+        treeCost: getFormattedNumber(i18n.language, projectDetails.treeCost || 0),
         publish: projectDetails.publish,
         visitorAssistance: projectDetails.visitorAssistance,
         enablePlantLocations: projectDetails.enablePlantLocations,
@@ -156,8 +157,6 @@ export default function BasicDetails({
   }, [projectDetails]);
 
   const onSubmit = (data: any) => {
-    // console.log('data.treeCost', data.treeCost.replace(/,/g, '.'));
-
     setIsUploadingData(true);
     let submitData = {
       name: data.name,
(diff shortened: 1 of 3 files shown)


harshvitra:
Screenshot 2021-01-13 at 7 19 17 PM


@norbertschuler now the user can type text too, that is why we added the code to replace anything except those characters. Is there are reason for allowing the characters ?

norbertschuler:
> @norbertschuler now the user can type text too, that is why we added the code to replace anything except those characters. Is there are reason for allowing the characters ?

Well, it allows to type alphanumeric characters, but as you also see from the red text beneath, the value is not accepted as a valid input. We use formattedNumber and parseNumber to format the number into the localized representation for the language of the user and then parse it again before storing it into the backend. No invalid should be able to be stored.

ankitecd:
I think since it is a field which accepts only number so we can avoid entering characters in the first place and also avoiding error than showing them would be better in this case

norbertschuler:
> I think since it is a field which accepts only number so we can avoid entering characters in the first place and also avoiding error than showing them would be better in this case

As this field allows entering a floating number value, it also allows other characters used to present floating number as e.g. a `.` or a `,`. If we ever want to support language not using just numeric characters, we should keep this solution, e.g. 
Bildschirmfoto 2021-01-15 um 08 43 35

harshvitra (approved):

PR #736, opened 2021-01-13
Remove paypal extra condition

harshvitra:
Removed a condition which will never occur in Paypal


harshvitra:
@norbertschuler this can be merged

norbertschuler (approved):
Just confirmed, Paypal donation is still working with this patch.

PR #757, opened 2021-01-18
always display selected and default country first in chooser dialog

norbertschuler:
Fixes #

Changes in this pull request:
- always display selected and default country first in chooser dialog in country & language chooser
- always display DE and US and selected and default country first in chooser dialog in currency chooser
- removed duplicates from `src/utils/countryCurrency/countriesData.json`
- avoid crash if currency has no min value defined (also crashes in the `develop` branch)

### DIFF
--- a/src/features/common/Layout/Footer/SelectLanguageAndCountry.tsx
+++ b/src/features/common/Layout/Footer/SelectLanguageAndCountry.tsx
@@ -10,6 +10,7 @@ import {
   getCountryDataBy,
   sortCountriesByTranslation,
 } from '../../../../utils/countryCurrency/countryUtils';
+import { getStoredConfig } from '../../../../utils/storeConfig';
 import supportedLanguages from '../../../../utils/language/supportedLanguages.json';
 import { ThemeContext } from '../../../../theme/themeContext';
 import GreenRadio from '../../InputTypes/GreenRadio';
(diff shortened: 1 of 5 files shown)


harshvitra:
> * API call for payment options, e.g. app/projects/proj_0CgTQJMm5zX624sdpVDet93O/paymentOptions?country=AS can reset the chosen currency to `EUR`, so the selected value gets changed and the dialog does not display any preselected country+currency anymore. If chosen such a currency twice, it keeps being displayed with the wrong amounts of `EUR`, until the payment finally fails. (This is also in the `develop` branch.)

Do we want to create a separate issue for this or we solve it in this PR ?

norbertschuler:
> > * API call for payment options, e.g. app/projects/proj_0CgTQJMm5zX624sdpVDet93O/paymentOptions?country=AS can reset the chosen currency to `EUR`, so the selected value gets changed and the dialog does not display any preselected country+currency anymore. If chosen such a currency twice, it keeps being displayed with the wrong amounts of `EUR`, until the payment finally fails. (This is also in the `develop` branch.)
> 
> Do we want to create a separate issue for this or we solve it in this PR ?

I created a new bug issue for it: https://github.com/Plant-for-the-Planet-org/planet-webapp/issues/773

PR #778, opened 2021-01-25
Replaced payment Request Button in legacyDonations

### LINKED ISSUE #775: Optimise Payment Request Button to be used for legacy donations
### Optimise Payment Request Button to be used for legacy donations

Currently we have two files - 
Used for main donations - src/features/donations/components/PaymentRequestForm.tsx
Used for legacy donations - src/features/legacyDonations/components/PaymentRequestForm.tsx

The only difference in both is that for the main donations we needed 2 buttons - **Native pay + Continue button**

For Legacy we need only one - **Native pay**

So either we separate out the buttons or we pass a prop to decide whether to show the button or not.

rishsingh07:
Fixes #775 

- [x] Clean up code
- [x] Separate payment methods
- [x] Use stripe key from payment setup
- [x] For legacy donations same changes as normal donations
- [ ] Browserstack tests for donations 
- [ ] Testing the mobile donations


sagararyal:
@harshvitra 
IntegrationError: Can only create one Element of type cardNumber.
instrument.js:109 IntegrationError: Can only create one Element of type cardExpiry.


For Payment Request Button, API is somehow using tokens endpoint.

Bug:

If amount is changed, reinitialization does not happen properly and user is stuck at /v1/tokens


Screen Shot 2021-01-27 at 2 19 01 AM

Issue #791, opened 2021-02-03
New Tenant : Culcha Candela


sarveshpro:
Please contact partner at plant-for-the-planet dot org before creating this request.

Type of Tenant [Please select all that apply]

- [ ] Custom Domain: http://wald.culchacandela.de/

- [ ] Custom Leaderboard Page (if necessary, please add link to designs: )

- [ ] Multiple Language (please list languages )

- [ ] Tenant Description on the Leaderboard Page (If default template) and the tree target:

- [ ] If hosting at subdirectory, current server type (NGINX/APACHE)?

- [ ] Enable Login (Auth0 SPA App) get Client ID

Callback URLs : [tenantURL]/verify-email, [tenantURL]/login

Logout URLs: [tenantURL]/verify-email, [tenantURL]/

**Checklist for Developer**

- [ ] Tenant is present in the Backend
- [ ] Completed Meta.json file and any translations
- [ ] Completed tenant.config file
- [ ] Svg logo url:
- [ ] Custom fonts (woff, woff2) url:
- [ ] Generate new maxbox token
- [ ] Use ASSET_PREFIX with maxcdn if high traffic site
- [ ] Generate Favicons at: https://iconifier.net

PR #795, opened 2021-02-04
Tenant/Culchacandela

sarveshpro:
#791


norbertschuler (approved):
LGFM, Browserstack tests probably do not work for hot fix branches:

Bildschirmfoto 2021-02-04 um 09 50 31

norbertschuler (approved):
As this is already on production, I approve it.

PR #799, opened 2021-02-08
Sat1

harshvitra:
Tenant Sat1

use tenant name = sateins in env


harshvitra:
Not to be merged in develop

src/features/projects/styles/MapboxMap.module.scss
```diff
@@ -434,7 +434,7 @@
 
 @include mdTabletView {
   .mapContainer {
-    padding-top: 60px;
+    // padding-top: 60px; // Removed for iframe
```
norbertschuler, line 437: We still have this iFrame protection for Heroku deployment in our codebase at https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/d4928069530a5ee401b03c3958af29b6e53d78f6/server.js#L48-L52 and for Vercel at https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/develop/vercel.json - we probably have no iFrame protection for the static export we use at Nextjs.

norbertschuler: I did a small test embedding one of our Netlify websites into an iFrame as following:
```
<html>
<body>
  <iframe src="https://www.trilliontreecampaign.org/" style="height:600px;width:600px;" title="test">
</body>
</html>
```
Serving this file leads to the following error (probably from within `auth0-spa-js`): `DOMException: Failed to read the 'sessionStorage' property from 'Window': Access is denied for this document.` - so even if we allow iFrames the page probably won't work.

norbertschuler: I tested with the `sateins` tenant environment without navigation included into an iframe and there are no problems as long as iframe protection is not active.

harshvitra:
Closing PR keeping the branch

PR #811, opened 2021-02-12
give credit card frame more time

norbertschuler:
Changes in this pull request:
- added more time


norbertschuler:
does not work

PR #816, opened 2021-02-14
explicitly added font-family for some text elements

norbertschuler:
Fixes Slack reported issue

Changes in this pull request:
- added font-family for setting items in modal dialog
- added font-family for p in menu items

Screen Shot 2021-02-14 at 3 03 15 AM

### DIFF
--- a/src/features/common/Layout/Navbar/index.tsx
+++ b/src/features/common/Layout/Navbar/index.tsx
@@ -124,7 +124,7 @@ export default function NavbarComponent(props: any) {
                   <div className={'link_icon'}>
                     <UserProfileIcon />
                   </div>
-                  <p className={router.pathname === '/t/[id]' ? 'active_icon' : ''}>
+                  <p className={router.pathname === SingleLink.onclick ? 'active_icon' : ''}>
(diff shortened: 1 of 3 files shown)


src/features/common/Layout/Navbar/index.tsx
```diff
@@ -124,7 +124,7 @@ export default function NavbarComponent(props: any) {
                   <div className={'link_icon'}>
                     <UserProfileIcon />
                   </div>
-                  <p className={router.pathname === SingleLink.onclick ? 'active_icon' : ''}>
+                  <p className={router.pathname === '/t/[id]' ? 'active_icon' : ''}>
```
harshvitra, line 127: This highlights other profiles as well. 
Example if you visit my profile, Me will be highlighted

norbertschuler: Thank you for pointing this out. In that case I revert this change as otherwise it would be wrong. As the router.pathname is just the general path `t/[id]` I do not yet see any possibility to check if a user is currently viewing the "Me" area.

harshvitra (approved):

PR #822, opened 2021-02-17
Remove leaderboard and added bg image

harshvitra:


norbertschuler (approved):
Leaderboard API call is still done, but otherwise looks good to me ;-)

PR #830, opened 2021-02-18
Feature/projectSiteEnhancement

rishsingh07:
Aligned "Save and Add Project" button

### DIFF
--- a/src/features/user/ManageProjects/styles/StepForm.module.scss
+++ b/src/features/user/ManageProjects/styles/StepForm.module.scss
@@ -169,18 +169,27 @@
 
 .inlineLinkButton {
   font-family: $primaryFontFamily;
-  width: 100%;
-  font-size: 14px;
+ width: 100%;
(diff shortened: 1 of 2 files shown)


src/features/user/ManageProjects/styles/StepForm.module.scss
```diff
@@ -169,18 +169,27 @@
...
   letter-spacing: normal;
-  text-align: center;
   color: #68b030;
+  display: flex;
+  justify-content: center;
+  margin-top: 10px;
 }
 .inlineLinkButton:hover {
   cursor: pointer;
+  font-weight: bold;
```
harshvitra, line 186: Why are we changing weight on hover ?

rishsingh07: Because the hover effect was not visible.

harshvitra (approved):

PR #838, opened 2021-02-22
fixed local storage bug

### LINKED ISSUE #833: TypeError: Cannot read property 'getItem' of null
If the browser cannot get items from local storage, can we set default values for language/currency?

Login would def not work, but that should be fine.

Sentry Issue: [APP-20-NEXT-15T](https://sentry.io/organizations/plant-for-the-planet/issues/2184837800/?referrer=github_integration)

```
TypeError: Cannot read property 'getItem' of null
  at call (./src/utils/storeConfig.ts:4:33)
  at new Promise (<anonymous>)
  at storeConfig (./src/utils/storeConfig.ts:3:1)
  at h (./pages/_app.tsx:68:5)
  at doRender (../../client/index.tsx:521:7)
...
(22 additional frame(s) were not displayed)
```

rishsingh07:
fixed #833

### DIFF
--- a/src/utils/storeConfig.ts
+++ b/src/utils/storeConfig.ts
@@ -1,7 +1,13 @@
 import getsessionId from './apiRequests/getSessionId';
 
 export async function storeConfig() {
-    const userLang = localStorage.getItem('language') || 'en';
+  let userLang;
+  if(localStorage){
+    userLang = localStorage.getItem('language') || 'en';
+  }
+  else{
+    userLang = "en";
+  }
     await fetch(`${process.env.API_ENDPOINT}/public/v1.2/${userLang}/config`, {
         headers: {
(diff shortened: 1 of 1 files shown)


norbertschuler (commented):
This is an improvement for this special case, but we should write an abstraction for use localStorage in general everywhere in our codebase to avoid such errors. Compare #712

harshvitra:
> This is an improvement for this special case, but we should write an abstraction for use localStorage in general everywhere in our codebase to avoid such errors. Compare #712

I'm working on this, meanwhile this can be merged

norbertschuler (approved):
Code formatting could be nicer (use some tool?), but looks good for me.

harshvitra:
> Code formatting could be nicer (use some tool?), but looks good for me.

@rishsingh07 for next time

PR #856, opened 2021-02-25
Added loader at signup

### LINKED ISSUE #738: Add a loader on Complete Signup action
Currently the UI is just squashed and it doesn't have a good UX. 

Adding a loader once the user clicks on the action will improve the UX

rishsingh07:
Fixed #738

### DIFF
--- a/src/features/user/CompleteSignup/index.tsx
+++ b/src/features/user/CompleteSignup/index.tsx
@@ -36,6 +36,7 @@ export default function CompleteSignup() {
 
   const [token, setToken] = React.useState('')
 
+  const [submit, setSubmit] = React.useState(false)
   React.useEffect(() => {
     async function loadFunction() {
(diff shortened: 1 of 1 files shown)


harshvitra:
@rishsingh07 the design is still glitched. The white card changes shape after we click on it. 
Check the donations popup and how it works

src/features/user/CompleteSignup/index.tsx
```diff
@@ -145,6 +146,7 @@ export default function CompleteSignup() {
   }, [type])
 
   const createButtonClicked = async (data: any) => {
+    setSubmit(true);
```
harshvitra, line 149: What happens if Signup fails ? 

when are you setting this loader to false?

harshvitra (changes requested):

harshvitra (approved):

PR #476, opened 2020-11-20
Zoom 3

sarveshpro:
Changes in this pull request:
- Refactored map component
- Time Travel Feature (Former Zoom 3)
- Vegetation Change on some projects
- multiple satellites sources, select range (2014-2021)
- custom node API integration
- removed Mapbox token dependency

Requires `SITE_IMAGERY_API_URL` in env

Platforms in use for this feature - 
- Google Earth Engine
- Planet Labs

dragging and panning on time travel is still buggy, but it works. please test the entire projects map flow. Home -> Single Project -> Time Travel and back

PR #862, opened 2021-02-28
Feature/new zoom 3

sarveshpro:
Almost Ready

known issues - 
- vegetation change not available on some projects - requires standardization of project site geometry format
- ~planetLabs source is available only for some specific projects~ added auto fallback to sentinel
- ~zooming and panning on time travel is glitchy~ fixed but still has some issues
- sentinel 2017 is not visible for some projects (at least for me)


norbertschuler:
Does this replace #476? Maybe you want to copy the hint about the necessary env variable and close the old pull request.

PR #863, opened 2021-02-28
Upgrades

sagararyal:
Include SEPA sofort warning texts


norbertschuler:
- [ ] On Browserstack this test fails as the element showing the supported recipient of the donation is not shown in the donation dialog, but when I test it with https://planet-webapp-etesoou1x-planetapp.vercel.app/s/sagar-aryal it seems to work:

FAIL __test__/RemoveDirectGift.test.js (48.07 s)
  ● remove › should remove gift direct
https://user-images.githubusercontent.com/1532418/109436604-a0c8fa00-7a20-11eb-8721-aee613f33b3b.mp4

Plant-for-the-Planet (approved):

PR #870, opened 2021-03-02
Added samesite none and secure true

harshvitra:


i18n/index.js
```diff
@@ -24,6 +24,9 @@ export default new NextI18Next({
 
     // cache the language in cookies and local storage
     caches: ['cookie', 'localStorage'],
+    cookieSameSite: 'none',
+    cookieSecure: true,
+    cookieExpirationDate: new Date(firstDay.getTime() + 7 * 24 * 60 * 60 * 1000),
```
norbertschuler, line 29: This does not work for me, but the following does:

```
    cookieMinutes: 7 * 24 * 60,
    cookieOptions: { path: '/', sameSite: 'none', secure: true },    
```

Also a secure cookie is only set if serving the website with SSL, so it won't working testing on localhost without SSL.

Bildschirmfoto 2021-03-02 um 18 41 54

norbertschuler: Unfortunately a secure cookies does not seem to solve the problem of showing errors accessing the locaStorage if embedding the website with an iframe from a different domain.

harshvitra: We do have localstorage as the fallback if cookie doesn't work, but I guess in case of iframes we are not even getting that.

PR #888, opened 2021-03-08
Add Login Tests

sagararyal:


sagararyal:
It should be possible to disable 2FA for a specific user using Auth0 rule:


Screen Shot 2021-03-08 at 6 47 38 PM

__test__/manage-project.test.js
```diff
@@ -0,0 +1,81 @@
...
+  });
+  it('should create a project and update data for a project.', async () => {
+    const url = await driver.getCurrentUrl();
+    await driver.navigate().to(`${url}/login`);
+    (await driver).sleep(200);
+    //Login Process
+    await driver.findElement(By.id('username')).sendKeys('test-tpo@plant-for-the-planet.org');
+    await driver.findElement(By.name('action')).click();
+    await driver.findElement(By.id('password')).sendKeys('CcCFg2enJ@C7XrV3ukqHbYYbaN-2hBW7hh6_Ye8kBorZAwczZfdM*TJnMLdgpbi');
+    await driver.findElement(By.name('action')).click();
```
sagararyal, line 17: After password, find name /id code

Screen Shot 2021-03-18 at 1 31 57 PM

Add OTP generated by the app and then find name('action) and click.

sagararyal:
It is possible to generate OTP with following code:

```
    <script src="https://unpkg.com/@otplib/preset-browser@^12.0.0/buffer.js"></script>
    <script src="https://unpkg.com/@otplib/preset-browser@^12.0.0/index.js"></script>

    <script type="text/javascript">
      // window.otplib.authenticator
      // window.otplib.hotp
        const token = window.otplib.totp.generate('HGUFYPKRNP4A2RAZ');
        document.write(token);
    </script>
    ```
    
    Secret above is test secret.

Secret (HGUFYPKRNP4A2RAZ) can be added in .env

**Comment by norbertschuler** [comment]
This is what I see now running a test on e.g. https://dev.pp.eco (so maybe also applies to other staging Vercel deployments):

Bildschirmfoto 2021-03-29 um 14 08 31

@sagararyal I don't this this is OTP any more, is it?

**Comment by sagararyal** [comment]
@norbertschuler you should select Google Authenticator, that is OTP.

File: __test__/manage-project.test.js [hunk 2]
```diff
@@ -0,0 +1,123 @@
+const { By, until, Key } = require('selenium-webdriver-3');
+const { getOtp } = require('./getotp');
```
norbertschuler, line 2: I think Linux systems are case sensitive regarding file names, so this should be `getOtp` for being able to work with our GitHub actions.

next.config.js
```diff
@@ -121,6 +121,7 @@ module.exports = withPlugins([[withBundleAnalyzer], [withSourceMaps]], {
     VERCEL_URL: process.env.VERCEL_URL,
     SITE_IMAGERY_API_URL: SITE_IMAGERY_API_URL,
     WIDGET_URL: process.env.WIDGET_URL,
+    TEST_MFA_URL: process.env.TEST_MFA_URL,
```
norbertschuler, line 124: We do not need the env TEST_MFA_URL for our normal application, but only for testing, so I do not think it is necessary do add it here, do we?

sagararyal: only necessary for testing.

norbertschuler: Instead of this place, TEST_MFA_URL should be added to https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/develop/.env.example.

norbertschuler:
- [ ] Browserstack does not find the button to add a new project:

`NoSuchElementError: no such element: Unable to locate element: {"method":"css selector","selector":"*[id="addProjectBut"]"}`

Issue #779, opened 2021-01-26
Use location data from config API call


norbertschuler:
**Describe the feature you'd like to see included in Plant-for-the-Planet Web App.**
The config API call now contains an approximated user location, e.g.
```
{
  "appVersions": {
    "ios": "1.3.5",
    "android": "1.3.5"
  },
  "clientIp": "xxx",
  "country": "DE",
  "currency": "EUR",
  "cdnMedia": {
    "images": "https://cdn.plant-for-the-planet.org/media/images",
    "cache": "https://cdn.plant-for-the-planet.org/media/cache",
    "pdfs": "https://cdn.plant-for-the-planet.org/media/uploads/pdfs"
  },
  "loc": {
    "latitude": "51.24800",
    "longitude": "7.12280",
    "city": "Wuppertal",
    "postalCode": "42117",
    "countryCode": "DE",
    "regionCode": "NW",
    "timezone": "Europe/Berlin"
  }
}
```
This can be used to show the users location on maps, form et al.

**What are the use cases for this feature?**
Center maps on user location, use user location as default value in forms.

sarveshpro:
Yes, this is very much needed on register trees. I am working on this.
but I can't find it in config
![image](https://user-images.githubusercontent.com/27153515/109926830-6ba2fd00-7ce9-11eb-8eb0-853793f58410.png)

sarveshpro:
location data is missing on app development 
https://app-development.plant-for-the-planet.org/public/v1.2/en/config

sarveshpro:
I think it is because the config is added by a Cloudflare worker only on production

sarveshpro:
@sagararyal can we create a similar worker for development and staging.

norbertschuler:
@sarveshpro we probably would have to create two new workers for staging and develop as the CDN urls are included in the response or add some switches in there to return different CDN urls depending on the url the workers is called with. For a quick fix, you might just code our Planet-App the way it just optionally set the user location if the data is available and test with changing your local API endpoint to production?

sarveshpro:
Sure

sarveshpro:
This isn't tested, to test this on local with Production APIs requires login which is not possible because of Auth0

norbertschuler:
@sarveshpro I added a Cloudflare worker for development (only supporting `en` and `de` and maybe returning wrong data for CDN urls).

sagararyal:
As a part of this redesign, 
[...]

PR #891, opened 2021-03-09
add satellite layer to edit sites map

sarveshpro:
Changes in this pull request -
 - add layer toggle on edit sites map
 - fly to user location(from config API) on register trees


sarveshpro:
#779

norbertschuler:
I added a Cloudflare worker to return the config for the development environment (https://app-development.plant-for-the-planet.org/public/v1.2/de/config and https://app-development.plant-for-the-planet.org/public/v1.2/en/config - no other languages) like for the production system with the location data included (CDN does not seem to work yet) in order to be able to test this. 

- [ ] First test seem to fail: 

Bildschirmfoto 2021-03-10 um 11 19 31

norbertschuler:
- [ ] Not sure if this is related to this PR or a general bug, but the info icon from mapbox overlaps with the zoom controls.

Bildschirmfoto 2021-03-10 um 10 47 23

src/features/user/ManageProjects/components/ProjectSites.tsx
```diff
@@ -487,9 +496,9 @@ function EditSite({ openModal, handleModalClose, changeSiteDetails, siteDetails,
     setGeoJson,
     geoJsonError,
     setGeoJsonError,
-    geoLocation:{
-      geoLatitude:36.96,
-      geoLongitude:-28.5
+    geoLocation: {
+      geoLatitude: 36.96,
```
norbertschuler, line 500: If you search for `36.96` you find a lot of these default values in our code. What is this place?

sarveshpro: That's the default map center I believe, which we use when we aren't looking at anything specific.

norbertschuler: Ok, but where does these coordinates come from - is it Atlantis? *just joking* 😁

norbertschuler:
- [ ]  In my example project I had defined a polygon at a zoom level showing no satellite map when I switched to the satellite layer - but I assume we have to deal with this. After I zoomed out a little bit, the data got loaded
- [ ] Can we give the "Satellite" text a little padding to the left and right?
 
Bildschirmfoto 2021-03-10 um 13 16 26
Bildschirmfoto 2021-03-10 um 13 17 05

norbertschuler:
- [ ] Just a question: Is this the zoom level we want to present the user for registering trees. I think we might zoom in more?

Bildschirmfoto 2021-03-10 um 13 15 13

norbertschuler:
> * [x]  Just a question: Is this the zoom level we want to present the user for registering trees. I think we might zoom in more?

Bildschirmfoto 2021-03-12 um 19 03 21

That's probably a better zoom level to start. I recognized the Copyright remark and info icon are now completely deleted, are they? @sagararyal Is that ok not to have any reference at all?

norbertschuler (approved):
LGFM - please clarify if no map credentials need to be given any more. @sagararyal

PR #893, opened 2021-03-09
Changed 'me' label

rishsingh07:
Changed me label for planet and ttc

### DIFF
--- a/public/tenants/planet/config.ts
+++ b/public/tenants/planet/config.ts
@@ -42,7 +42,7 @@ const config = {
         visible: true,
       },
       me:{
-        title: 'me',
+        title: 'Sign-In',
         onclick: '/me',
         visible: true,
       }
--- a/public/tenants/ttc/config.ts
+++ b/public/tenants/ttc/config.ts
@@ -41,7 +41,7 @@ const config = {
         visible: true,
       },
       me:{
-        title: 'me',
+        title: 'Sign-In',
         onclick: '/me',
         visible: true,
       }


public/tenants/planet/config.ts
```diff
@@ -42,7 +42,7 @@ const config = {
         visible: true,
       },
       me:{
-        title: 'me',
+        title: 'Sign-In',
```
norbertschuler, line 45: Is `Sign-In` changed to something else after the user logged in? Otherwise it can irritate the user and I wouldn't change the menu item name.

public/tenants/ttc/config.ts
```diff
@@ -41,7 +41,7 @@ const config = {
         visible: true,
       },
       me:{
-        title: 'me',
+        title: 'Sign-In',
```
norbertschuler, line 44: Same as above.

public/tenants/planet/config.ts
harshvitra: Maybe we use `Profile` key?

src/features/common/Layout/Navbar/index.tsx
```diff
@@ -125,7 +125,7 @@ export default function NavbarComponent(props: any) {
                     <UserProfileIcon />
                   </div>
                   <p className={router.pathname === SingleLink.onclick ? 'active_icon' : ''}>
-                    {t('common:' + SingleLink.title)}
+                  {isAuthenticated && userInfo ? t('common:' + SingleLink.loggedInTitle) : t('common:' + SingleLink.title)}
```
norbertschuler, line 128: If there are tenants allowing a login which do not have this resource, this could fail. So we better also check if `SingleLink.loggedInTitle` is defined before using it.

norbertschuler (changes requested):

src/features/common/Layout/Navbar/index.tsx
norbertschuler: added a safety check

norbertschuler (approved):

PR #908, opened 2021-03-16
Topbar with information

harshvitra:
Content is to be replaced with real content


src/features/common/Layout/Topbar/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+    const [showTopBar, setshowTopBar] = React.useState(true);
+    const { t, ready } = useTranslation(['common']);
+
+    return (
+        <div className={'topbarContainer'} style={{display:showTopBar ? 'flex' : 'none'}}>
+            <p>
+                {t('common:tobBarText')}
+                <a href="https://waldrekord.plant-for-the-planet.org/" target="_blank">
+                {t('common:clickHere')}
+                </a>
```
norbertschuler, line 20: You should just make the whole text clickable :-)

sagararyal:
Todo:

@sarveshpro 

TopBar notices can be added as json in www5 project.
We could think of integrating this into Webapp at some later time.

www5.planet.../topbar/tenantkey/messages.json

```
{
  "de":{
  "text": "Endspurt beim SAT.1 Waldrekord. Pflanz mit - wir schaffen die Million!",
  "ctaText": "Click Here",
  "ctaLink": "https://example.com"
  },
  "en":{
  "text": "Endspurt beim SAT.1 Waldrekord. Pflanz mit - wir schaffen die Million!",
  "ctaText": "Click Here",
  "ctaLink": "https://example.com"
  }
}
```

sarveshpro:
Not working because of CORS issue

next.config.js
```diff
@@ -118,6 +118,7 @@ module.exports = withPlugins([[withBundleAnalyzer], [withSourceMaps]], {
     VERCEL_URL: process.env.VERCEL_URL,
     SITE_IMAGERY_API_URL: SITE_IMAGERY_API_URL,
     WIDGET_URL: process.env.WIDGET_URL,
+    TOPBAR_API_URL: process.env.TOPBAR_API_URL,
```
norbertschuler, line 121: Please add values to https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/develop/.env.production and https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/develop/.env.development and https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/develop/.env.local.example so we have a default value and do not need to set the env variable in all of our deployments.

sarveshpro: Yes, but this url won't work, will have to find some other solution

sarveshpro: https://30c3f8e4.planet-ads.pages.dev/topbar/ten_I9TW3ncG/messages.json
gives CORS error when fetched in the webapp

norbertschuler: @sarveshpro isn't there a possibility to add a header, e.g. `access-control-allow-origin: *`, to allow fetching it?

PR #909, opened 2021-03-16
Error Context provider

harshvitra:
Error handling using a Global Error handler. 

The global error handler can be triggered in any page of the app. 

**You can use following from the context -** 
- toggleModal: Boolean
- setglobalErrorMessage: value
- setglobalErrorImage: value
- setglobalErrorCloseFunction: Function


norbertschuler (commented):
Asking the user to repeat the action (API call) which failed due to heavy load makes it worse, does it?

Bildschirmfoto 2021-03-16 um 15 02 48

harshvitra:
> Asking the user to repeat the action (API call) which failed due to heavy load makes it worse, does it?

Yes got the same feedback from Sagar. For now we will just ignore the errors, this one we will work on in some time.

PR #923, opened 2021-03-21
added web manifest for tenant ttc and planet

norbertschuler:
Fixes #

Changes in this pull request:
- no [PWA](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps), just adding a [manifest](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Add_to_home_screen)


public/tenants/planet/manifest.json
```diff
@@ -0,0 +1,52 @@
...
+    "name": "Plant for the Planet",
+    "short_name": "Plant",
+    "start_url": "/?utm_source=pwa",
+    "description": "Plant trees around the world - Plant-for-the-Planet",
+    "display": "fullscreen",
+    "background_color": "#fff",
+    "theme_color": "#68b030",
+    "lang": "en",
+    "orientation": "portrait",
+    "icons": [
```
norbertschuler, line 11: The transparent icon version does not work well for iPhone and Android devices:

IMG_9238IMG_9240

public/tenants/planet/manifest.json
```diff
@@ -0,0 +1,52 @@
+{
+    "name": "Plant for the Planet",
+    "short_name": "Plant",
+    "start_url": "/?utm_source=pwa",
+    "description": "Plant trees around the world - Plant-for-the-Planet",
+    "display": "fullscreen",
```
norbertschuler, line 6: `fullscreen` looks best, but with rounded corners and notch, the back arrow for "tree registrations" or "profile editing" is not clickable.

public/tenants/planet/manifest.json
```diff
@@ -0,0 +1,53 @@
+{
+    "name": "Plant for the Planet",
+    "short_name": "Plant",
+    "start_url": "/?utm_source=pwa",
+    "description": "Plant trees around the world - Plant-for-the-Planet",
+    "categories": ["education"],
+    "display": "standalone",
```
norbertschuler, line 7: `standalone` or `fullscreen` looks best, but with rounded corners and notch, the back arrow for "tree registrations" or "profile editing" is not clickable. We have to decide with [display mode](https://developer.mozilla.org/en-US/docs/Web/Manifest/display) is best.

public/tenants/planet/manifest.json
norbertschuler: I tried to add an additional icons set without transparency for the web manifest, but unfortunately iOS seems to prefer taking the app icons from https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/1a4308e94e6438a6c621655c9443ab57adb9fa42/src/features/common/Layout/Header/index.tsx#L51-L54.

norbertschuler:
@sagararyal besides the wrong transparent icons for iOS having this is better than what we currently have :-)

harshvitra:
I think we should go back to our previous icon

sagararyal:
let's use the non transparent icon, and publish them.

public/tenants/planet/manifest.json
norbertschuler: using none transparent icons again

norbertschuler:
> let's use the non transparent icon, and merge this

I will add a new icon set - I could not use my former icon set with white background and some transparent parts as all the transparent parts are shown in black by iOS:

IMG_9458

norbertschuler:
> let's use the non transparent icon, and merge this

Ready to merge if you want to.

PR #938, opened 2021-03-25
Added margin in footer above logos

rishsingh07:
Added margin above footer container.

### DIFF
--- a/src/features/common/Layout/Footer/Footer.module.scss
+++ b/src/features/common/Layout/Footer/Footer.module.scss
@@ -12,6 +12,7 @@
   flex-wrap: wrap;
   margin-bottom: 80px;
   justify-content: center;
+  margin-top: 20px;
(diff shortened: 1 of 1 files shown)


src/features/common/Layout/Footer/Footer.module.scss
```diff
@@ -12,6 +12,7 @@
   flex-wrap: wrap;
   margin-bottom: 80px;
   justify-content: center;
+  margin-top: 20px;
```
harshvitra, line 15: Remove the 20px on the container

and replace button instead of div

harshvitra: This is so that each logo gets the margin 

Screenshot 2021-03-25 at 5 41 32 PM

harshvitra (approved):

PR #950, opened 2021-03-30
Added scroll to top

rishsingh07:
Added scroll to top

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,6 @@
     "react-mapbox-gl": "^5.1.0",
     "react-mapbox-gl-draw": "^2.0.4",
     "react-player": "^2.8.2",
-    "react-use": "^17.2.1",
     "read-more-react": "^1.0.10",
     "styled-jsx": "^3.4.1",
     "uuid": "^8.3.2",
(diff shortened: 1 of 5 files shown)


src/features/common/Layout/index.tsx
```diff
@@ -5,7 +5,8 @@ import CookiePolicy from './CookiePolicy';
 import Header from './Header';
 import Navbar from './Navbar';
 import RedeemPopup from './RedeemPopup';
-
+import Scroll from './BackToTop/backToTop'
+import BackToTop from './BackToTop/backToTop';
```
norbertschuler, line 9: This looks like an unnecessary redundant import.

package.json
```diff
@@ -89,6 +89,7 @@
     "react-mapbox-gl": "^5.1.0",
     "react-mapbox-gl-draw": "^2.0.4",
     "react-player": "^2.8.2",
+    "react-use": "^17.2.1",
```
norbertschuler, line 92: This is another (big) package - is it necessary? https://bundlephobia.com/result?p=react-use@17.2.1

rishsingh07: I'll look for some solution to it.

norbertschuler: Generally it is no problem if we add a package if it is the best solution for the issue and everything else is too difficult. We just should have a short check before, if the problem can not be solved easily without it.

sagararyal:
why do we need this feature?

PR #951, opened 2021-03-30
enabled German translations for showing large number abbreviations

norbertschuler:
Fixes request of @felixfinkbeiner 

Changes in this pull request:
- enabled German translations for showing large number abbreviations

Bildschirmfoto 2021-03-31 um 14 19 52
Bildschirmfoto 2021-03-31 um 14 20 11

- disabled abbreviation for thousands (in all languages):

Bildschirmfoto 2021-03-31 um 13 56 28

### DIFF
--- a/src/utils/getFormattedNumber.ts
+++ b/src/utils/getFormattedNumber.ts
@@ -18,12 +18,11 @@ class NumberParser {
   }
 }
 
-// change 'de-disabled' to 'de' if you want to enable custom abbreviations for German
 const localizedAbbr = {
   'en': {
     'b': 'b', 'm': 'm', 'k': 'k',
   },
-  'de-disabled': {
+  'de': {
     'b': 'Mrd.', 'm': 'Mio.', 'k': 'Tsd.',
   },
 };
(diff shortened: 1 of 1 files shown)


norbertschuler:
@sagararyal would you prefer `10Mio`/`10Mrd.` or `10 Mio`/`10 Mrd.` with or without a space?

sagararyal:
In tree counters, we should disable rounding of millions as well until 99.99 million.

norbertschuler:
> In tree counters, we should disable rounding of millions as well until 99.99 million.

This will require as special formatting function or parameter only for treecounters. As you already merged https://github.com/Plant-for-the-Planet-org/planet-widgets/pull/53, do you also want to merge this first now?

felixfinkbeiner:
@norbertschuler I think there should be a space before Mio/Mrd etc

norbertschuler:
> In tree counters, we should disable rounding of millions as well until 99.99 million.

@sagararyal Show something like this?
Bildschirmfoto 2021-04-02 um 11 00 45

norbertschuler:
The widgets currently cannot deal with high numbers so well - if we add a space before a German abbreviation ` Mio.` the layout event got broken more:

Bildschirmfoto 2021-04-02 um 10 54 44Bildschirmfoto 2021-04-02 um 10 55 21

The tree counter could cope with higher numbers and the German abbreviation, but with the additional space the layout is already quite dense:

Bildschirmfoto 2021-04-02 um 10 59 40Bildschirmfoto 2021-04-02 um 10 59 10

sagararyal:
We should reduce the font size to accommodate larger numbers.
Harsh is aware of this.

PR #962, opened 2021-04-05
Status and Shop links in footer

rishsingh07:
Added status and shop link in footer

### DIFF
--- a/src/features/projects/components/maps/Credits.tsx
+++ b/src/features/projects/components/maps/Credits.tsx
@@ -44,13 +44,13 @@ export default function Credits({ setCurrencyCode }: Props): ReactElement {
                     {`🌐 ${language ? language.toUpperCase() : ''
                         } • ${selectedCurrency}`}
                 </div>
-                <a
+                {(process.env.TENANT === "ttc" || process.env.TENANT === "planet") ? < a 
                     rel="noopener noreferrer"
                     href={`https://www.thegoodshop.org/${userLang}/shop/`}
                     target={'_blank'}
                 >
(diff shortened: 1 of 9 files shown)


src/features/projects/components/maps/Credits.tsx
```diff
@@ -44,6 +44,20 @@ export default function Credits({ setCurrencyCode }: Props): ReactElement {
                     {`🌐 ${language ? language.toUpperCase() : ''
                         } • ${selectedCurrency}`}
                 </div>
+                <a
+                    rel="noopener noreferrer"
+                    href={`https://status.plant-for-the-planet.org/`}
+                    target={'_blank'}
+                >
+                    {t('common:status')}
```
norbertschuler, line 52: The status item is also not part of our bigger footer? Should it or shouldn't it?

Big footer:

Bildschirmfoto 2021-04-05 um 12 16 29

Small footer:

Bildschirmfoto 2021-04-05 um 12 16 40

rishsingh07: @harshvitra

harshvitra: Yes we might add it to the bigger footer also

sagararyal:
@rishsingh07 status should always point to status.pp.eco regardless of tenants.

sagararyal:
@rishsingh07 url is status.pp.eco

This will make sure site is still up even if DNS errors affect -planet.org

Please replace these as well
Screen Shot 2021-04-12 at 3 21 41 PM

PR #972, opened 2021-04-09
Total cost in donate button

rishsingh07:
Added total cost into donate button

### DIFF
--- a/src/features/donations/components/paymentMethods/CardPayments.tsx
+++ b/src/features/donations/components/paymentMethods/CardPayments.tsx
@@ -14,6 +14,7 @@ import AnimatedButton from '../../../common/InputTypes/AnimatedButton';
 import styles from './../../styles/Donations.module.scss';
 // import { payWithCard } from '../components/treeDonation/PaymentFunctions';
 import i18next from '../../../../../i18n';
+import getFormatedCurrency from '../../../../utils/countryCurrency/getFormattedCurrency';
 
 const { useTranslation } = i18next;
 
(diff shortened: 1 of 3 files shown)


src/features/donations/components/paymentMethods/CardPayments.tsx
```diff
@@ -240,6 +244,14 @@ function CardPayments({
           <div onClick={handleSubmit} className={styles.actionButtonsContainerCenter}>
             <AnimatedButton className={styles.continueButton} id='donateContinueButton'>
               {t('common:donate')}
+              <div className={styles.totalCost}
+              style={{ color: styles.light, fontSize: "14px" }}>
+              {getFormatedCurrency(
+                    i18n.language,
+                    currency,
+                    treeCount * treeCost
+                  )}
```
norbertschuler, line 253: This only adds the amount to the enabled donate button (not the disabled version) and only for credit card payment - is that intended?

Bildschirmfoto 2021-04-09 um 10 29 46Bildschirmfoto 2021-04-09 um 10 30 04

harshvitra: @rishsingh07  please add at other places also

sagararyal: @harshvitra @rishsingh07 I don't think this makes sense for `Continue with Sofort`, or `Continue with Giropay` buttons.

harshvitra (approved):

Issue #861, opened 2021-02-27
Replace Country and language selection UI


harshvitra:
### **Replace Country and language selection UI**

The current UI for country selection doesn't allow the user to easily search for their country+currency.
- user have to scroll alot
- the UI doesn't show any visual difference between all the elements (all text), 
- after click it goes to top, etc

<img width="1680" alt="Screenshot 2021-02-27 at 6 31 07 PM" src="https://user-images.githubusercontent.com/25605193/109388011-8c0a3a80-792a-11eb-9afa-f05cb3628067.png">

### **Suggestions -** 

We use country selection from our contact details form for this purpose.

<img width="405" alt="Screenshot 2021-02-27 at 6 32 19 PM" src="https://user-images.githubusercontent.com/25605193/109388099-0470fb80-792b-11eb-96ef-9963d3dd383e.png">

This would solve the need for #790

PR #1011, opened 2021-04-22
Select country bug

### LINKED ISSUE #790: Repeat selected item in country/currency list
**Describe the feature you'd like to see included in Plant-for-the-Planet Web App.**
Selected country should repeat again in the alphabetical list - even if it appears first or second.

Screen Shot 2021-01-27 at 1 52 20 AM

rishsingh07:
Fix #790

Trying to fix the select country bug where country gets vanished from the list after selecting it which ends up messing with the UI


norbertschuler:
I thought we would fix this by completely changing the country selection like described in #861.

PR #1022, opened 2021-04-26
Tree mapper and Account History!!!

rishsingh07:
To Do - 

- [x] Certificates - Account History
- [x] dot markers for sample trees - Treemapper
- [x]  use env variable for CDN: https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1022#discussion_r627960285 or https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1022#discussion_r639569703
- [x] merge conflicts
- [ ] scroll issue with open account record (desktop)
- [ ] verify the backend response with notion doc here - https://www.notion.so/plantfortheplanet/Account-History-API-2ac912ccd19743928b303386ccdae2e7
- [ ] Problem showing profile page with none develop build: https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1022#issuecomment-850434765

To Test - 

- [ ] Register Trees
- [ ] Manage Projects
- [ ] Add Projects
- [ ] logout on all pages
- [ ] visit all pages not logged in


sarveshpro:
- [ ] The text inside the header should not go outside 1200px content width. please refer accounts page.
- [ ] The settings button is missing.
- [ ] The background of the form and the page mismatch. please use $backgroundColor variable everywhere.
- [ ] get rid of the footer but leave some space(equal to the height of the footer) empty at the bottom of the page.
- [ ] use same header component for accounts page

image

sarveshpro (changes requested):
Please make the above changes

src/features/common/Layout/Header/accountHeader.tsx
```diff
@@ -0,0 +1,138 @@
...
+              // any other error
+            }
+          } catch (err) {
+            console.log(err);
+          }
+        } else {
+          loginWithRedirect({
+            redirectUri: `${process.env.NEXTAUTH_URL}/account/history`,
+            ui_locales: localStorage.getItem('language') || 'en',
+          });
```
norbertschuler, line 71: What is the difference between specifying the `redirectUri` directly in the `loginWithRedirect` function call or setting them in our localStorage like at https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/f358ad2e551ba81584c090d0ca88c105c59f0b09/pages/manage-projects/add-project/index.tsx#L53-L54?

norbertschuler: I assume if we want to support a redirectUri for Auth0, we need to add it to all configurations at Auth0 (which could be a lot of work as we already have added quite a lot of applications). Currently `/account/history` is not configured resulting into an error, so we probably need to change this towards using the solution I linked above:

Bildschirmfoto 2021-04-28 um 09 03 48

pages/manage-projects/[id].tsx
```diff
@@ -6,6 +6,10 @@ import GlobeContentLoader from '../../src/features/common/ContentLoaders/Project
 import AccessDeniedLoader from '../../src/features/common/ContentLoaders/Projects/AccessDeniedLoader';
 import Footer from '../../src/features/common/Layout/Footer';
 import { useAuth0 } from '@auth0/auth0-react';
+import AccountHeader from '../../src/features/common/Layout/Header/accountHeader'
+import i18next from '../../i18n';
+import { getLocalUserInfo } from '../../src/utils/auth0/localStorageUtils'
```
Comment (norbertschuler) [inline pages/manage-projects/[id].tsx:11]: Please remove unused imports.

public/static/locales/en/me.json
```diff
@@ -3,6 +3,7 @@
   "target": "Target",
   "share": "Share",
   "registerTrees": "Register Trees",
+  "manageProjects": "Manage Projects",
```
norbertschuler, line 6: We should try to avoid repeating too much duplicate content in translation resources (at least if the translation is within the same resource file). This is already existing as https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/f358ad2e551ba81584c090d0ca88c105c59f0b09/public/static/locales/en/me.json#L11

src/features/common/Layout/Header/Header.module.scss
```diff
@@ -4,3 +4,64 @@
...
+  justify-content: center;
+  align-items: center;
+  background: linear-gradient(
+      to top,
+      rgba(0, 0, 0, 0.5),
+      rgba(0, 0, 0, 0.2),
+      rgba(0, 0, 0, 0),
+      rgba(0, 0, 0, 0)
+    ),
+    url('https://cdn.plant-for-the-planet.org/media/images/app/bg_layer.jpg')
```
norbertschuler, line 21: Is is possible to also try to take the value from our environment variable in an .scss file as in an .tsx file? Or can we rewrite this to load the background image in the related component using this style like we did it for the user profile at https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/f358ad2e551ba81584c090d0ca88c105c59f0b09/src/features/user/UserProfile/screens/IndividualProfile.tsx#L61-L63

src/features/common/Layout/Header/accountHeader.tsx
norbertschuler, line 71: This code block looks like a duplicate the following code - can't we abstract it somehow? https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/f358ad2e551ba81584c090d0ca88c105c59f0b09/src/features/user/Account/screens/index.tsx#L73-L104  A similar code pattern also is used in https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/develop/src/utils/auth0/userInfo.ts.

norbertschuler (changes requested):

src/features/common/Layout/Header/Header.module.scss
rishsingh07: I guess individual profile uses different header. The header component just has back button and settings button which is going to be used in different pages which does not require other features which individual profile has.

src/features/common/Layout/Header/accountHeader.tsx
rishsingh07: Yess I forgot to remove it from accounts

pages/manage-projects/[id].tsx
Comment (rishsingh07) [reply pages/manage-projects/[id].tsx]: done

public/static/locales/en/me.json
rishsingh07: done

src/features/common/Layout/Header/accountHeader.tsx
rishsingh07: @sarveshpro can we remove it from accounts?

public/static/locales/en/me.json
norbertschuler: Please do not just remove `settingManageProject` as it is of course used at https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/b49b0a05d22130b0f9345244937383d84eeecc8c/src/features/user/UserProfile/components/SettingsModal.tsx#L98

rishsingh07: my bad, fixed.

norbertschuler: I did some more fixes and removed the now unused resource `manageProjects`.

norbertschuler:
- [ ] Adding new projects does not yet have a header, e.g. with http://localhost:3000/manage-projects/add-project

src/features/common/Layout/Header/accountHeader.tsx
sarveshpro: I'll suggest you create a userContext. that way you can update it's value from here and use it anywhere in the app.

norbertschuler: I saw that someone added http://localhost:3000/account/history to the valid Callback URLs of our development Auth0 tenant `pftp`, but not for the production Auth0 tenant. Thus this will still fail on production. If we want to allow this URL as a callback URL, it has to be done everywhere!

norbertschuler: There would be at least two production applications on Auth0 to adapt, but this would have to be maintained for every other application (tenant) allowing this URL in the future. I would prefer the code as following:

```
localStorage.setItem('redirectLink','/account/history');
loginWithRedirect({
  redirectUri: `${process.env.NEXTAUTH_URL}/login `,
  ui_locales: localStorage.getItem('language') || 'en',
});
```

src/features/common/Layout/Header/accountHeader.tsx
```diff
@@ -0,0 +1,84 @@
...
+    setEditProfileModalOpen(true);
+  };
+    return (
+      <div>
+      <div className={styles.headerBG}>
+        <div className={styles.accountsHeader}>
+          <div className={styles.navContainer}>
+            <button
+              onClick={() => {
+                router.back();
```
norbertschuler, line 49: Going backwards the browser history can result into an error if going through the Auth0 login process for URLs like http://localhost:3000/manage-projects/add-project or http://localhost:3000/account/history as following:

Bildschirmfoto 2021-04-30 um 16 12 18

norbertschuler (changes requested):

src/features/common/Layout/Header/accountHeader.tsx
norbertschuler: I change it.

pages/_app.tsx
```diff
@@ -153,6 +154,7 @@ export default function PlanetWeb({ Component, pageProps, err }: any) {
       >
         <ThemeProvider>
           <CssBaseline />
+          <UserPropsProvider>
```
norbertschuler, line 157: @sarveshpro can we be sure the user handling in this abstracted HOC does not interfere with any other handling in our code e.g. https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/ec938bc0d35d5ac659fc1710bebb9f65446b2bb2/pages/t/%5Bid%5D.tsx#L90-L112 or https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/ec938bc0d35d5ac659fc1710bebb9f65446b2bb2/src/utils/auth0/userInfo.ts#L7-L46

norbertschuler: At least this change cost us one additional API call to fetch the user's profile. 

- [ ] We already have a duplication at `/login` following even by a third call when redirecting to the user's profile `/t/user` - now this is calling the profile API four(!) times.

Bildschirmfoto 2021-05-03 um 08 23 14

- [ ] The duplication now is calling the profile API call twice being logged in when opening the user's profile.

Bildschirmfoto 2021-05-03 um 08 24 06

sarveshpro:
can someone test everything related to user flow on this branch?

- [ ] login
- [ ] logout
- [ ] signup
- [ ] public profile
- [ ] private profile
- [ ] All authenticated pages and API calls

src/features/common/Layout/Header/AccountHeader.module.scss
```diff
@@ -0,0 +1,90 @@
...
+  justify-content: center;
+  align-items: center;
+  background: linear-gradient(
+      to top,
+      rgba(0, 0, 0, 0.5),
+      rgba(0, 0, 0, 0.2),
+      rgba(0, 0, 0, 0),
+      rgba(0, 0, 0, 0)
+    ),
+    url('https://cdn.plant-for-the-planet.org/media/images/app/bg_layer.jpg')
```
norbertschuler, line 16: I still want to ask if it's possible to set this image within the JavaScript code to allow us to use the env variable CDN_URL for the URL in order to be able to support different environments.

norbertschuler (changes requested):

src/features/common/Layout/Header/Header.module.scss
norbertschuler: close in favor of https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1022#discussion_r627960285

src/features/common/Layout/Header/AccountHeader.module.scss
sarveshpro: Yes, we could do it but the CDN for development doesn't have this image 😅

norbertschuler: @sagararyal can we deploy above image to all CDN (develop, staging, production)?

src/features/common/Layout/Header/accountHeader.tsx
norbertschuler: This should be solved in https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1057

src/features/common/Layout/Header/AccountHeader.module.scss
```diff
@@ -0,0 +1,121 @@
...
+  // position: fixed;
+  // top:0;
+  background: linear-gradient(
+      to top,
+      rgba(0, 0, 0, 0.5),
+      rgba(0, 0, 0, 0.2),
+      rgba(0, 0, 0, 0),
+      rgba(0, 0, 0, 0)
+    ),
+    url('https://cdn.plant-for-the-planet.org/media/images/app/bg_layer.jpg')
```
norbertschuler, line 44: https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1022#discussion_r627960285 is still open, but the code got moved here.

norbertschuler (changes requested):

norbertschuler:
Both Vercel and Cloudflare Pages have build problem throwing this error:


11:03:43.390 | Error: [object Object] is not a PostCSS plugin
-- | --
11:03:43.390 | &gt; Build error occurred
11:03:43.391 | Error: &gt; Build failed because of webpack errors
11:03:43.391 | at /opt/buildhome/repo/node_modules/next/dist/build/index.js:17:924
11:03:43.391 | at runMicrotasks (&lt;anonymous&gt;)
11:03:43.391 | at processTicksAndRejections (internal/process/task_queues.js:97:5)
11:03:43.391 | at async Span.traceAsyncFn (/opt/buildhome/repo/node_modules/next/dist/telemetry/trace/trace.js:6:584)

norbertschuler:
- [ ] I can not login any more currently - anyone else?

Bildschirmfoto 2021-05-28 um 15 51 05

norbertschuler:
> * [x]  After login, the profile page does not load - pages as .e.g https://planet-webapp-9kp2wm9a6-planetapp.vercel.app/account/history can be loaded, so the login succeeded, but the profile e.g. https://planet-webapp-9kp2wm9a6-planetapp.vercel.app/t/norbert-schuler does not get loaded and instead showing an endless spinner:
> 
> Update: This also happening locally when using `npm run build & npm run start -p 3000` - but not if running `npm run dev`!

Still happening to me on Vercel:

111C5545-7248-4942-B108-5598800C1666

sarveshpro:
> > * [x]   After login, the profile page does not load - pages as .e.g https://planet-webapp-9kp2wm9a6-planetapp.vercel.app/account/history can be loaded, so the login succeeded, but the profile e.g. https://planet-webapp-9kp2wm9a6-planetapp.vercel.app/t/norbert-schuler does not get loaded and instead showing an endless spinner:
> > 
> > Update: This also happening locally when using `npm run build & npm run start -p 3000` - but not if running `npm run dev`!
> 
> Still happening to me on Vercel:
> 
> 120145811-6b30da00-c1e4-11eb-8198-e2a682d63f3c

Apparently this is now working on local `npm run build` & `npm run start` but not on vercel deployment

sarveshpro:
failed, moving to other tasks

sarveshpro:
same issue happens on netlify - https://planet-sarvesh.netlify.app/

harshvitra:
Reason why we don't seen anything in console because there is a compile time error

sarveshpro:
This PR is a lost cause, closing it. Don't delete the branch immediately tho. wait until Treemapper and A/c history are merged to develop

PR #1037, opened 2021-04-29
use Punycode URL for tenant ulmpflanzt

norbertschuler:
Changes in this pull request:
- use Punycode URL for tenant ulmpflanzt to avoid error with Stripe API call


norbertschuler:
The url used for Stripe is not the tenantURL from the tenant config, but `NEXTAUTH_URL` from the environment. I changed that in the deployment configuration at Netlify.

PR #1042, opened 2021-05-05
Change texts for single tree count

### LINKED ISSUE #869: Add singular labels for quantities
There are places where we have '1 trees' it should be changed to '1 tree'.

rishsingh07:
Fixed #869


harshvitra (dismissed):
It works
but you should use the plural from i18n

src/features/donations/screens/ThankYou.tsx
```diff
@@ -182,7 +182,17 @@ function ThankYou({
                 />
               </div>
               <div className={styles.donationCount}>
-                {t('donate:myTreesPlantedByOnLocation', {
+                {donation.treeCount == 1 ?
```
norbertschuler, line 185: Please let's use the technology already existing in the i18n module automatically using a plural resource if the given parameter `count` does not equals `1` as e.g. already used at https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/498672c39de909e83f2445f8f47dc988aa794beb/src/features/donations/screens/ContactDetails.tsx#L87-L93 plus the related resources at https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/498672c39de909e83f2445f8f47dc988aa794beb/public/static/locales/en/donate.json#L2-L3
The parameter `count` is only used to decide which resource (singular or plural) to choose while the formatted parameter `treeCount` is actually used to be displayed.

norbertschuler (changes requested):

norbertschuler:
@rishsingh07 can you please check if we should use plural/singular distinction also at https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/3f707dbcdd7020391aa3a9164723c194deb86232/pages/claim/%5Btype%5D/%5Bcode%5D.tsx#L254

rishsingh07:
@norbertschuler  I cannot test this as I do not have redeem code.

norbertschuler:
> @norbertschuler I cannot test this as I do not have redeem code.

@rishsingh07 There seem to be three valid link types at https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/3f707dbcdd7020391aa3a9164723c194deb86232/pages/claim/%5Btype%5D/%5Bcode%5D.tsx#L74

There are some codes in the development backend to test redemptions at https://app-development.plant-for-the-planet.org/backend/redemption, but I am unsure which codes to use combined with one of theses link types, e.g. http://localhost:3000/claim/donor/VH46SX2KO37A seems to work for testing on develop (not on production), but please **do not finish** claiming it as otherwise it's used until the develop backend database got reset.

I am sorry, I have not found any example of a code to claim only 1 tree to test the singular case.

norbertschuler (approved):
LGFM

harshvitra (approved):

harshvitra (approved):

PR #1051, opened 2021-05-11
Font size converted in variables

rishsingh07:
These are the font sizes I've created as per the major number of useage.

fontXXSmall: '10px',
    fontXSmall: '12px',
    fontSmall: '14px',
    fontSixteen: '16px',
    fontMedium: '18px',
    fontLarge: '22px',
    fontXLarge: '26px',
    fontXXLarge: '36px',
    fontXXXLarge: '48px'

There are still different font-sizes present like:
130, 75, 20, 34, 100, 150, 11, 8, 27, 24, 40, 38, 28, 15, 52, 30, 42, 72, 80

Please let me know how to proceed.

### DIFF
--- a/src/features/common/ContentLoaders/Donations/PaymentProgress.module.scss
+++ b/src/features/common/ContentLoaders/Donations/PaymentProgress.module.scss
@@ -11,12 +11,12 @@
   font-family: $primaryFontFamily;
   font-weight: 800;
   margin-top: 36px;
-  font-size: 36px;
+  font-size: $fontDoubleExtraLarge;
 }
 
 .progressBackdropText {
   font-family: $primaryFontFamily;
   font-weight: 400;
   margin-top: 8px;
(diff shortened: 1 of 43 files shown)


norbertschuler (approved):
I did a small test of common dialogs and designs and it looks ok for me (no unreadable or cut texts). 

After we merge this, nobody should use any `font-size` with a number anymore. I do not know how to enforce this by software, so we must communicate this to all developers and maybe add it to one [contribution guideline](https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/develop/CONTRIBUTING.md). In the end it just requires developers discipline (the same as not committing console.log statements) :-)

We could need a short documentation page on notion.io with a design guide to know which font sizes are available to use and where the best should be used!

sagararyal (approved):
Let's document this, and add it where it is accessible by all.

PR #1066, opened 2021-05-20
Active icon bug for different tenants

rishsingh07:
This was the issue. 
image


sarveshpro (approved):
LGTM ✔ tested on salesforce tenant

norbertschuler (approved):
We should not try to check for any menu titles strings which can change in the future. We have tenants with different title strings, so checking for the key name of the link is much more reliable. I tested this with various tenants including Salesforce, but it would be nice to get another review as well.

sarveshpro (approved):
Works for planet and salesforce tenant ✔

PR #1077, opened 2021-05-27
added _redirects file for Netlify

norbertschuler:
Fixes #

Changes in this pull request:
- added _redirects file for Netlify as redirects specified in netlify.toml do not work


sarveshpro:
Shouldn't we remove redirects from `netlify.toml` now. or else will have to maintain at 2 places.

sarveshpro (approved):

PR #1096, opened 2021-06-04
Height fixed in explore

harshvitra:
Height fixed in explore toggle of safari


norbertschuler (approved):

norbertschuler:
@harshvitra be aware I changed the base to develop again as we are so close to a regular release we do not need to do this as a hot fix, but can just merge it to develop and ship it with everything else 😁

PR #1097, opened 2021-06-06
Fix payment status

harshvitra:
Fixed payment status

### DIFF
--- a/src/features/donations/components/PaymentFunctions.ts
+++ b/src/features/donations/components/PaymentFunctions.ts
@@ -250,7 +250,7 @@ export async function payDonationFunction({
   }
 
   try {
-    const paidDonation = await payDonation(payDonationData, donationID, token);
+    const paidDonation = await payDonation(payDonationData, donationID, token);    
 
     if (paidDonation) {
       if (paidDonation.code === 400 || paidDonation.code === 401) {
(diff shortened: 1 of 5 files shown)


src/features/donations/components/PaymentFunctions.ts
```diff
@@ -271,13 +271,15 @@ export async function payDonationFunction({
...
-      } else if (paidDonation.paymentStatus === 'success' || paidDonation.paymentStatus === 'pending' ||
-      paidDonation.data.status === "success" || 
-      paidDonation.data.status === "paid" ||
-      paidDonation.data.paymentStatus === "paid") {
+        return paidDonation;
+      } else if (paidDonation.paymentStatus === 'success'
+        || paidDonation.paymentStatus === 'pending'
+        || paidDonation.paymentStatus === 'paid'
+        || paidDonation.status === 'success'
+        || paidDonation.status === 'pending'
```
norbertschuler, line 279: @harshvitra I added this just to be sure as `paidDonation.paymentStatus === 'pending'` was also requested before -but I am not sure about the status model currently used for payments.

norbertschuler (approved):
This works for me, but currently on the development backend, successful donations are still marked as "not finalized", so we probably are not yet finished - maybe this can be solved in the backend?

Bildschirmfoto 2021-06-06 um 15 56 46

src/features/donations/components/PaymentFunctions.ts
norbertschuler: Credit card payment on staging (not development) backend are getting finalized now, Paypal not.

sagararyal: This is prob because of missing web hooks to staging from PayPal.

PR #1115, opened 2021-06-14
Update profile after redemption fix

### LINKED ISSUE #1107: Update profile after redemption
**Describe the bug**
I was able to test redeeming a code on develop (taking the first Coupon code from the backend). It seems to work, but my counter on the profile is not updated after the API calls are done (it updates when reloading the page):

Bildschirmfoto 2021-06-08 um 10 56 13

Also in the confirmation dialog for redemption the button is a little too small for the German translation of the text:

Bildschirmfoto 2021-06-08 um 10 55 08

develop

**To Reproduce**
1. Login and redeem a code.
2. See error

**Expected behavior**
User profile should get updated (as API call is rather old, maybe we have to refresh the whole page?)

**What actually happens**
User profile (counter) is not updated.

- OS: latest MacOS

rishsingh07:
Fixes #1107

### DIFF
--- a/src/features/user/UserProfile/components/RedeemModal.tsx
+++ b/src/features/user/UserProfile/components/RedeemModal.tsx
@@ -31,7 +31,7 @@ export default function RedeemModal({
 
   const config = tenantConfig();
 
-  const { user, contextLoaded, token } = React.useContext(UserPropsContext);
+  const { user, contextLoaded, token, setUser } = React.useContext(UserPropsContext);
 
   const imageRef = React.createRef();
   const sendRef = () => imageRef;
@@ -119,6 +119,8 @@ export default function RedeemModal({
           setCodeRedeemed(true);
           setIsUploadingData(false);
           setCodeValidated(false);
+          const newUser = {...user, score:{personal: res.schemata.treeCounter.countPlanted}}
+          setUser(newUser);
         }
       });
     }


src/features/user/UserProfile/components/RedeemModal.tsx
```diff
@@ -119,6 +119,8 @@ export default function RedeemModal({
           setCodeRedeemed(true);
           setIsUploadingData(false);
           setCodeValidated(false);
+          const newUser = {...user, score:{personal: res.schemata.treeCounter.countPlanted}}
```
norbertschuler, line 122: I would explicitly mark this code with a comment to remind ourselves that it needs to be updated once the old API gets replaced with something new.

norbertschuler:
- [ ] I used one of the develop codes to add 0.2 trees (chocolate coupon) to my profile formerly having 90.6 trees on my counter and got this result:

Bildschirmfoto 2021-06-18 um 10 55 08

After I used reload in my browser to load my profile again the 0.2 trees got added and 90.8 trees and my goal are displayed again:

Bildschirmfoto 2021-06-18 um 10 56 23

src/features/user/UserProfile/components/RedeemModal.tsx
```diff
@@ -119,6 +119,8 @@ export default function RedeemModal({
           setCodeRedeemed(true);
           setIsUploadingData(false);
           setCodeValidated(false);
+          const newUser = {...user, score:{personal: res.schemata.treecounter.countPlanted, received: user.score.received}}
```
norbertschuler, line 122: I assume you should use countPersonal here. And we should somehow still show the goal.

norbertschuler (changes requested):

PR #1121, opened 2021-06-18
fix single project page error

sarveshpro:


norbertschuler:
Exception still happens on this pull request:
Bildschirmfoto 2021-06-18 um 20 42 21

PR #1157, opened 2021-07-26
TreeMapper Feature Release

sarveshpro:
- [ ] depends on backend release!

To test

- [x] Donations
- [x] Project Pages
- [x] Maps Functionality
- [x] Login
- [x] Profile Pages
- [x] Register Trees
- [x] Manage Projects
- [x] Gifts

Deployments to test -
https://ttc-staging.netlify.app/ - TTC tenant, production, Netlify
https://dev.pp.eco/ - Planet tenant, staging, Vercel (for testing donations, gifts)
https://planet-app-dev.herokuapp.com/ - Planet tenant, production, Heroku

To test plant locations (TreeMapper) check Yucatan Project on production

UI issues found:

- [x] White polygons rarely visible on map with a lot of streets on it https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1157#issuecomment-887851014
- [x] Text labels missing on iOS and Safari on MacOS https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1157#issuecomment-887852041
- [x] Tree mappings rarely visible on satellite view with green areas https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1157#issuecomment-888457842
- [x] Not possible to switch to satellite view on mobile devices https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1157#issuecomment-888619538

PR #1161, opened 2021-07-28
only used the date part for conversion without time

norbertschuler:
Fixes https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1157#issuecomment-887852041

Finally I found the reason: Safari can not convert a value of kind “2021-07-23 00:00:00” into a date with new Date(string) as it needs a timezone or just a date like “2021-07-23" without the time value. So I truncated the value to the first 10 characters.

We should check our whole codebase for such errors. If we receive such date values from the API, we should not just try to convert them into Date objects!


sarveshpro:
Working for me
image

sarveshpro (approved):

PR #1162, opened 2021-07-28
set map style to satellite if no treemapper data found

sarveshpro:


norbertschuler (commented):
Difficult to test without data on staging, so I tested with production API :-)
Looks o.k. for me, but this feature might be irritating for the users if no tree mapping data exists for the first area shown but only for other areas of the project not visible at first sight.

sarveshpro:
addition to this also added a fallback to default style if user zooms to zoom level more than 15 while on satellite view.
avoiding this
image

sarveshpro:
fixes #1163

norbertschuler:
> fixes #1163

You removed this again, right? So I unlinked the issue again too.

norbertschuler (approved):
Works for me. Not fixing "fallback to default style if user zooms to zoom level more than 15 while on satellite view".

norbertschuler (approved):

sarveshpro:
Yes, we decided to not fix that because in some areas we do have more detailed imagery, and we don't want it to be missed out. There's a fix for this we will define max zoom level for each project on the backend.

PR #1167, opened 2021-08-04
Treemapper improvements

sarveshpro:
1. display HID
2. satellite button on phone

### DIFF
--- a/src/features/projects/components/PlantLocation/PlantLocationDetails.tsx
+++ b/src/features/projects/components/PlantLocation/PlantLocationDetails.tsx
@@ -125,6 +125,7 @@ export default function PlantLocationDetails({
       {plantLocation && (
         <div className={'singleProjectDetails'}>
           <div className={styles.treeCount}>
+            <div>
             {plantLocation.type === 'multi' && (
               <>
                 <span>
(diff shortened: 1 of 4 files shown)


src/features/projects/components/PlantLocation/PlantLocationDetails.tsx
```diff
@@ -146,6 +147,10 @@ export default function PlantLocationDetails({
             )}
             {plantLocation.type === 'single' && <span>{t('1Tree')} </span>}
             {plantLocation.type === 'sample' && <span>{t('sampleTree')} </span>}
+            </div>
+            <div>
+              {plantLocation.hid?plantLocation.hid.substring(0, 3) + "-" + plantLocation.hid.substring(3):null}
```
norbertschuler, line 152: What does HID mean and is it for sure that it has always 3 or more characters (so this code does not fail)?

norbertschuler (approved):
Works for me. Just above request for clarifying HID.

src/features/projects/components/PlantLocation/PlantLocationDetails.tsx
sarveshpro: HID is Hex ID. should always be 6 characters.

PR #1173, opened 2021-08-10
Landing Video

harshvitra:
Status:

- [x]  Prevent video from loop
- [x] Fix video not playing on the first load on many mobile devices.


src/features/common/LandingVideo/index.tsx
```diff
@@ -27,34 +27,34 @@ function VideoContainer({ setshowVideo }: Props): ReactElement {
       switch (localStorage.getItem('language')) {
         case 'de':
           setvideoURL(
-            `${process.env.CDN_URL}/landing-vid/planet/en-intro-mobile-planet.mp4`
+            `https://a.plant-for-the-planet.org/wp-content/videos/en-intro-mobile-planet.mp4`
```
norbertschuler, line 30: How does this fit with https://github.com/Plant-for-the-Planet-org/plant-for-the-planet-org.github.io/pull/22/files?

sagararyal: @norbertschuler file size limits, still testing.

sagararyal: Problem is mainly Safari iOS, video does not load automatically.

sagararyal:
- [ ] would be lovely to have support for `canplaythrough` HTML event and disable video where the response is false.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/canplaythrough_event

- [ ] Based on user agent googlebot; etc mark the video disabled.

PR #1179, opened 2021-08-13
fix missing sample tree name

### LINKED ISSUE #1169: Sample trees without scientific name are not selectable in area list
**Describe the bug**
There are some sample trees in the current production data from Yucatán which do have `"scientificName":null` in the API response. This leads to not showing a name in the list of sample trees for an area and therefore also not showing the link to look at the details of the sample tree. If viewing the sample tree by the point on the map, one also sees the field for scientific name stays empty in the details.

develop & production

**To Reproduce**
1. Go to Yucatán 
2. Click on area with HID 58X-1C3
3. See error

**Expected behavior**
Show some placeholder text for missing scientific names, e.g. "unknown species" with the link to the detail page where you also show "unknown species".

**What actually happens**
The name is missing and thus a link is missing too.

**Screenshots**
Bildschirmfoto 2021-08-05 um 08 30 15
Bildschirmfoto 2021-08-05 um 08 31 39

- OS: latest MacOS

sarveshpro:
Fixes #1169

### DIFF
--- a/src/features/projects/components/PlantLocation/PlantLocationDetails.tsx
+++ b/src/features/projects/components/PlantLocation/PlantLocationDetails.tsx
@@ -280,7 +280,7 @@ export default function PlantLocationDetails({
                   <span>
                     {plantLocation.scientificName
                       ? plantLocation.scientificName
-                      : plantLocation.scientificSpecies}
+                      : plantLocation.scientificSpecies?plantLocation.scientificSpecies:t('unknown')}
                   </span>
                 </div>
               </div>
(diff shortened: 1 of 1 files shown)


norbertschuler (changes requested):
The list is successfully fixed, but the space below the scientific name label on the detail screen is still an empty void :-) - shouldn't we put some 'Unknown' there too?

Bildschirmfoto 2021-08-13 um 08 11 04

sarveshpro:
> The list is successfully fixed, but the space below the scientific name label on the detail screen is still an empty void :-) - shouldn't we put some 'Unknown' there too?
> 
> Bildschirmfoto 2021-08-13 um 08 11 04

Fixed

norbertschuler (approved):

PR #1181, opened 2021-08-15
fix map not available issue in satellite mode

### LINKED ISSUE #1163: Disable satellite view when Zoom level increases
Screenshot 2021-07-28 at 11 47 21 PM

sarveshpro:
Fixes #1163

Changes in this pull request:
- change satellite map source to esri wayback which does not fail when map data is not available


sagararyal (approved):
LGFM

PR #1189, opened 2021-08-18
Fix satellite map when switching between map modes

sarveshpro:
reset map zoom when changing modes


sagararyal:
Some issues still remain.
Uploading Screen Recording 2021-08-19 at 10.07.23 PM.mov…

sagararyal (changes requested):
check video.

sarveshpro:
@sagararyal check zoom buttons now

norbertschuler:
This is how zoom buttons look like on landscape mode on an iPhone X overlapping with another button, but it’s probably an edge case:
399CF2C9-9518-4B46-B362-0A937B4BA45B

sarveshpro:
> This is how zoom buttons look like on landscape mode on an iPhone X overlapping with another button, but it’s probably an edge case:
> 

Not sure why the project details is taking full width in landscape mode.
It should be more like this. Although this is not specific to this PR. You can create a new issue for this.
image

norbertschuler:
> Not sure why the project details is taking full width in landscape mode.

The iPhone X simulation in my Chrome browser behaves different from my real iPhone X. It's more like on an iPhone 5 simulation in Chrome where the projects are shown with full width.

> It should be more like this. Although this is not specific to this PR. You can create a new issue for this.

I create a new issue #1204.

PR #1196, opened 2021-08-20
changed BrowserNotSupported screen

norbertschuler:
Changes in this pull request:
- simplified browser not support screen (as flex layout is also not supported on some older browsers)
- added a link with https://urlopen.link/ to open from InAppBrowser into another browser on the device

This does not work for me tested with an InAppBrowser in an QR Code scanner app on my Nexus 7 tablet as well as with the UC browser of a OnePlus 7 :-(


norbertschuler:
Closed as not working.

PR #1199, opened 2021-08-24
Fix : sample tree tag null

### LINKED ISSUE #1198: Don’t show null for missing sample tree tags
**Describe the bug**
For missing sample tree tags we should not show `null` as a value, but just hide it. 

production 

**To Reproduce**
1. View plant locations 
2. See error

**Screenshots**

image

image

**Smartphone (please complete the following information):**
- Device: iPhone X
- OS: latest iOS
- Browser: Chrome

sarveshpro:
Fixes #1198


norbertschuler (approved):
LGFM

PR #1200, opened 2021-08-24
set default country as DE if country code is XX or T1

### LINKED ISSUE #1188: Check if country from config is supported
Like requested at https://github.com/Plant-for-the-Planet-org/planet-donations/issues/101 we also should check if the country given with the config API call is supported in our code by testing at https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/9d78483cf5d7857cf3d2cb854f0cc337b49be0d0/src/utils/storeConfig.ts#L18-L24 if the country is included in our own country list https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/develop/src/utils/countryCurrency/countriesData.json and defaulting to country = "DE" otherwise.

tejassonar:
Fixes #1188 

@sarveshpro

### DIFF
--- a/src/utils/storeConfig.ts
+++ b/src/utils/storeConfig.ts
@@ -16,8 +16,13 @@ export async function storeConfig() {
     .then(async (res) => {
       const config = await res.json();
       localStorage.setItem('config', JSON.stringify(config));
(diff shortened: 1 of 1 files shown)


sarveshpro (approved):
LGFM

PR #1208, opened 2021-08-25
change skip button style

### LINKED ISSUE #1206: Improve skip button
- [x] Skip Button: White with green outline to make it less prominent. We wan’t to discourage people from skipping. The current button is so prominent that everyone will want to skip. Especially on mobile it should be a lit smaller.

- [x] Next to “Explore” button (to the left), add a similarly styled button “How does this work?“, where the icon is a standard triangular play button, which initiates a replay of the intro video.

- [ ] Make the skip button on mobile even smaller?

- [ ] Padding is missing between the play button and the explore button in the Safari version

- [ ] This button should not just be the icon button, but also “How does this work?” text next to it

sarveshpro:
Fixes #1206


src/features/common/LandingVideo/styles.module.scss
```diff
@@ -97,3 +97,40 @@
         transform: rotate(360deg);
     }
 }
+
+
+.playButton {
+    position: fixed;
```
norbertschuler, line 103: Giving this button a fixed position of course can not adapt the position to the changing width of the explore button in different languages.

norbertschuler (changes requested):

src/features/common/LandingVideo/styles.module.scss
norbertschuler: I added some space to the right site of the play button to support also longer translations.

norbertschuler (approved):
If we agree to not have a dark mode for the Skip button, this PR could be merged.

PR #1219, opened 2021-08-27
Fixed mobile donations popup

### LINKED ISSUE #1197: Closing donation dialog on mobile device touching the wrong places
**Describe the bug**
The current version of https://www1.plant-for-the-planet.org/ (also https://www.trilliontreecampaign.org/) on my iPhone X again shows the behavior of closing the donation dialog when touching the screen on areas without elements like text or forms if you just wanted to scroll it up a little bit.

This seems to happen only if opening the donation dialog from the project list and not from the project details page. 

production

**To Reproduce**
1. Open donate dialog for a project
2. Scroll up and down touching the screen on various places
3. See error

**Video**
https://user-images.githubusercontent.com/1532418/130435731-36b68c5c-f0b4-43d3-843b-9af7fcc7c666.MP4

**Smartphone (please complete the following information):**
- Device: iPhone X
- OS: latest iOS

harshvitra:
Fixes #1197

### DIFF
--- a/src/features/projects/screens/Projects.tsx
+++ b/src/features/projects/screens/Projects.tsx
@@ -164,15 +164,6 @@ function ProjectsList({
         <div
           style={{ transform: `translate(0,${scrollY}px)` }}
           className={'container'}
-          onTouchMove={(event) => {
-            if (isMobile) {
-              if (event.targetTouches[0].clientY < (screenHeight * 2) / 8) {
-                setScrollY(event.targetTouches[0].clientY);
-              } else {
-                setScrollY((screenHeight * 2) / 9);
-              }
-            }
-          }}
         >
           <div
             className={'header'}


norbertschuler:
This pull request prevents setting the `scrollY` state of the projects list and thus rendering `ProjectsList` again and loosing the `open` state for the `ProjectSnippet`, but other events like swiping down can still force a rendering of `ProjectsList` and thus closing the donation modal. I wonder if it's possible to disable any state changes or the rendering of the component at all while the donation modal is open?

norbertschuler:
Maybe we can test https://github.com/willmcpo/body-scroll-lock if it helps to prevent the elements behind the modal to scroll?

norbertschuler:
> Maybe we can test https://github.com/willmcpo/body-scroll-lock if it helps to prevent the elements behind the modal to scroll?

I tested it, but it won't work as calling `enableBodyScroll()` in `ProjectSnippet` also forces a rendering of `ProjectsList` and thus closing the donation modal. So almost anything I tried to modify `ProjectsList` from `ProjectSnippet` to not listen to events forced a rendering.

src/features/projects/screens/Projects.tsx
```diff
@@ -164,15 +164,6 @@ function ProjectsList({
         <div
           style={{ transform: `translate(0,${scrollY}px)` }}
```
norbertschuler, line 165: I assume the whole state `scrollY` is not used any more now and could be removed everywhere?

norbertschuler (approved):
This at least improves the current production version for mobile users. So we should push it until we found some other solution.

PR #1220, opened 2021-08-30
Suggest API

rishsingh07:
Added suggest API in contacts form.


sagararyal:
@rishsingh07 can you clean up console logs, and also implement this for signup page and edit profile as well?

norbertschuler:
- [ ] I recognized the suggestions showing an address + zip + city + country, but we are not using the country. Is it possible to limit the suggestion to the chosen country?

norbertschuler:
- [ ] Suggestions are not supporting dark mode:

Bildschirmfoto 2021-09-04 um 17 03 18

norbertschuler (approved):
LGFM

PR #1221, opened 2021-08-30
fixed changing of user profile

### LINKED ISSUE #1168: Changing profile type not possible
**Describe the bug**
In the latest version of the website editing the profile allows to change the profile type between individual, school and company. This was not possible before, but now a dropdown is shown to change the profile type. But as the profile type is not send with the API request to change the profile information, the profile type change is actually not done when submitting the form.

develop and production

**To Reproduce**
1. Go to your profile
2. Click the settings menu and choose to change your profile
3. Change your profile type from the dropdown and save the profile
4. Load the profile again (full reload of website)
5. See error

**Expected behavior**
Either do not show the dropdown to change the profile type like before and support the change of the profile type within the API call by adding the profile type as an additional parameter in the request and verify the backend supports it.

**Screenshots**
Bildschirmfoto 2021-08-04 um 13 05 06

- OS: latest MacOS

tejassonar:
Fixes #1168

Changes:
- send type to API for profile changes, but not if type is TPO as then the API would not accept this parameter
- correctly preselect type from loading user profile
- small dark mode fix for MuiAlert message text

@norbertschuler

### DIFF
--- a/src/features/user/UserProfile/components/EditProfileModal.tsx
+++ b/src/features/user/UserProfile/components/EditProfileModal.tsx
@@ -165,14 +165,14 @@ export default function EditProfileModal({
     setIsUploadingData(true);
     const bodyToSend = {
       ...data,
+      type: type,
       country: country,
     };
     if (contextLoaded && token) {
       try {
         putAuthenticatedRequest(`/app/profile`, bodyToSend, token)
           .then((res) => {
-            console.log(res);
-            if(res.code !== 400) {
+            if (res.code !== 400) {
               setSeverity('success');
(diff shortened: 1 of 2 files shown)


norbertschuler:
- [ ] This works on develop backend, but there is another error if you are a company or school (type organization or education) and trying to edit your profile in the way the dropdown is alway shown as "Individual" and the school or company name is not shown. This only happens when freshly loading the profile (it works if you just have changed the profile type):

Bildschirmfoto 2021-09-01 um 09 21 57

Maybe we can fix this error as well.

norbertschuler (approved):
fixed some more stuff - see description

PR #1238, opened 2021-09-08
Added classification of project

rishsingh07:
Removed comments to show classification of the project.
image


norbertschuler:
- [ ]  Did we also want to be able to filter the project list for a classification? (Or maybe as a technical solution just add the ability to use the - translated - classification string as a search term?)

sarveshpro:
Can you increase the minimized height of the container so that the back button does not overlap on the classification text
https://user-images.githubusercontent.com/27153515/132483170-62b1df91-90d8-4dcb-b043-092b8c9d85da.mp4

src/features/projects/components/projectDetails/getProjectClassification.ts
```diff
@@ -1,21 +1,21 @@
...
       break;
     case 'agroforestry':
       parseClassification = 'Agroforestry';
       break;
     default:
       parseClassification = '';
-  }
+  } 
   return parseClassification;
 }
```
sarveshpro, line 21: This needs to be translated. you don't really need a switch case just make `t(project.classification)` and add all possible values to translation file. you might want to check if `project.classification` exists first.

sarveshpro (changes requested):

public/static/locales/en/donate.json
```diff
@@ -96,5 +96,11 @@
...
   "orgNamePublishedTax": "Company name is visible on our leaderboard and app. Tax-deduction receipts will be issued for the same name.",
   "orgNamePublished": "Company name is visible on our leaderboard and app.",
-  "aboutProject": "About Project"
+  "aboutProject": "About Project",
+  "natural-regeneration": "Natural Regeneration",
+  "managed-regeneration": "Managed Regeneration",
+  "large-scale-planting": "Large Scale Planting",
+  "agroforestry": "Agroforestry",
+  "urban-planting": "Urban Planting",
+  "other-planting": "Other Planting"
```
sagararyal, line 105: @rishsingh07 I suggest following one standard here.

sagararyal:
can you add
case "mangroves"  = Mangroves

PR #1247, opened 2021-09-09
Hide play button

harshvitra:
Hiding play button for other tenants


norbertschuler (approved):
LGFM (tested locally with Salesforce tenant vs. Planet tenant)

PR #1319, opened 2021-11-10
NextJS 12 update

sarveshpro:
Todo - 
- [ ] migrate react-i18next to version 8
- [ ] i18next-browser-languagedetector doesn't work with the new version, check for updates


i18n/index.js
```diff
@@ -1,34 +0,0 @@
-import path from 'path';
-import LanguageDetector from 'i18next-browser-languagedetector';
-import NextI18Next from 'next-i18next';
-import { initReactI18next } from 'react-i18next';
-import getConfig from '../tenant.config';
-
-const config = getConfig();
-
-export default new NextI18Next({
```
norbertschuler, line 9: I still found some `import i18next` in our codebase. Shouldn't they all be removed?

next-i18next.config.js
```diff
@@ -0,0 +1,36 @@
+// const getConfig = require('./tenant.config');
+const path = require('path');
+const LanguageDetector = require('i18next-browser-languagedetector');
+const { initReactI18next } = require('react-i18next');
```
norbertschuler, line 4: This does not seem to be used anymore. Probably these npm modules can be removed (also from package.json).

norbertschuler (changes requested):

norbertschuler:
This is an important architecture issue if we ever want to be able to upgrade to newer next versions.

norbertschuler:
I tried to continue this PR, but wondered if [react-i18next ](https://react.i18next.com/) is suited if we want to support not only SSR at Vercel, but also a static version of the website to be deployed at Heroku and at Netlify?

norbertschuler:
I closed this as I think keeping it could be dangerous as this code is not working.

PR #1320, opened 2021-11-10
Mega Menu 

rishsingh07:
To do: 

- [x] Translation to be added.
- [x] Direct user to appropriate href.
- [x] 
image


public/tenants/planet/config.ts
```diff
@@ -36,13 +36,50 @@ const config = {
         title: 'aboutUs',
         onclick: 'https://a.plant-for-the-planet.org/',
         visible: true,
+        subMenu: [
+          {
+            title: 'Overview',
```
norbertschuler, line 41: This solution does not consider translating resources. Please use the `title` as a key for a translation resource, e.g. `overview` -> `Overview` as at https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/ef8e93859407c7d9a2cf91a26c6120e3b1c001da/src/features/common/Layout/Navbar/index.tsx#L162

norbertschuler (changes requested):

src/features/common/Layout/Navbar/index.tsx
```diff
@@ -216,7 +216,10 @@ export default function NavbarComponent(props: any) {
                         <a
                           key={submenu.title}
                           className={'menuRow'}
-                          href={submenu.onclick}
+                          href={i18n.language === 'en' ? submenu.onclickEN :
+                          i18n.language === 'de' ? submenu.onclickDE :
+                          i18n.language === 'es' ? submenu.onclickES :
+                        ''}
```
norbertschuler, line 222: This leads for wrong links (just https://planet-webapp-8c27paq4u-planetapp.vercel.app/) for French, Italian and Portuguese. There should be a default link without a locale if no i18n.language matches (as maybe we add further languages in the future).

norbertschuler (changes requested):
You should talk to @sarveshpro and find a decent common solution :-)

tejassonar:
image

On small device, icons are overlapping

rishsingh07:
> image
> 
> On small device, icons are overlapping

It's happening because of language. Can be solved in a different PR I guess.

src/features/common/Layout/Navbar/index.tsx
```diff
@@ -1,30 +1,59 @@
...
+    return () => window.removeEventListener('resize', handleResize);
+  }, [handleResize]);
+  return width;
+};
 const { useTranslation } = i18next;
 const config = tenantConfig();
 export default function NavbarComponent(props: any) {
   const { t, ready, i18n } = useTranslation(['common']);
   const router = useRouter();
+  const lang_path = {
```
norbertschuler, line 31: Please use https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/develop/src/utils/constants/wpLanguages.ts which @sarveshpro added for his implementation of language dependent Wordpress links (we need to have a common software architecture and do not build stuff twice).

norbertschuler (changes requested):

norbertschuler:
- [ ] The Mega menu is not dark mode compatible yet:

Bildschirmfoto 2021-11-17 um 08 35 00

norbertschuler (approved):
LGFM (tests are failing, but I think not because of this pull request)

sarveshpro (changes requested):
Translation seems to be broken for some elements
image

sarveshpro:
About us button doesn't have `cursor: pointer`, please add it to show that the button is clickable.
image

sarveshpro (approved):

PR #1321, opened 2021-11-10
add principles and standards section

sarveshpro:
image

### DIFF
--- a/src/tenants/planet/LeaderBoard/components/Stories.tsx
+++ b/src/tenants/planet/LeaderBoard/components/Stories.tsx
@@ -14,7 +14,7 @@ export default function Stories({}: Props): ReactElement {
       {ready ? (
         <div className={styles.wrapper}>
           <div className={styles.container}>
-            <a className={styles.storyCard} href={`https://a.plant-for-the-planet.org/${lang_path[i18n.language]}/principles`}>
+            <a className={styles.storyCard} href={`https://a.plant-for-the-planet.org/${lang_path[i18n.language]?lang_path[i18n.language]:'en'}/principles`}>
               <img
                 src={'/tenants/planet/images/leaderboard/restoreTrees.svg'}
               />
--- a/src/tenants/planet/LeaderBoard/components/Stories.tsx
+++ b/src/tenants/planet/LeaderBoard/components/Stories.tsx
@@ -28,7 +28,7 @@ export default function Stories({}: Props): ReactElement {
                 </div>
               </div>
             </a>
-            <a className={styles.storyCard} href={`https://a.plant-for-the-planet.org/${lang_path[i18n.language]}/standards`}>
+            <a className={styles.storyCard} href={`https://a.plant-for-the-planet.org/${lang_path[i18n.language]?lang_path[i18n.language]:'en'}/standards`}>
               <img
                 src={'/tenants/planet/images/leaderboard/selectProjects.svg'}
               />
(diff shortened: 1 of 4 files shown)


src/tenants/planet/LeaderBoard/components/Stories.tsx
```diff
@@ -8,30 +8,40 @@ interface Props {}
...
-            <div>
-              <img src={'/tenants/planet/images/leaderboard/playIcon.svg'} />
-              <p>{t('planet:watchVideo')}</p>
-            </div>
+  return (
+    <>
+      {ready ? (
+        <div className={styles.wrapper}>
+          <div className={styles.container}>
+            <a className={styles.storyCard} href="https://a.plant-for-the-planet.org/principles">
```
norbertschuler, line 16: If we ever want to localize these pages, it would be better to link them like `https://a.plant-for-the-planet.org/${i18n.language}/principles` (needs i18n to get loaded) - at least this page is also available in German https://a.plant-for-the-planet.org/de/principles.

src/tenants/planet/LeaderBoard/components/Stories.tsx
```diff
@@ -8,30 +8,40 @@ interface Props {}
...
+                <h2>{t('planet:restoreTrees')}</h2>
+                <div>
+                  <img
+                    src={'/tenants/planet/images/leaderboard/mapIcon.svg'}
+                  />
+                  <p>{t('planet:learnMore')}</p>
+                </div>
+              </div>
+            </a>
+            <a className={styles.storyCard} href="https://a.plant-for-the-planet.org/standards">
```
norbertschuler, line 30: Same as above, but there is no German page for https://a.plant-for-the-planet.org/de/standards existing yet (but maybe will be added later on).

norbertschuler (changes requested):

src/tenants/planet/LeaderBoard/components/Stories.tsx
norbertschuler: (You can compare with my solution in https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1302.)

src/tenants/planet/LeaderBoard/components/Stories.tsx
```diff
@@ -1,37 +1,48 @@
...
-              <img src={'/tenants/planet/images/leaderboard/playIcon.svg'} />
-              <p>{t('planet:watchVideo')}</p>
-            </div>
+  const { i18n, t, ready } = useTranslation(['planet']);
+  return (
+    <>
+      {ready ? (
+        <div className={styles.wrapper}>
+          <div className={styles.container}>
+            <a className={styles.storyCard} href={`https://a.plant-for-the-planet.org/${lang_path[i18n.language]}/principles`}>
```
norbertschuler, line 17: This leads to a link like https://a.plant-for-the-planet.org/undefined/principles for Portuguese :-( It works as WP is handling this case, but it does not look very nice.

norbertschuler (changes requested):
You should talk with @rishsingh07 and you both might find a common decent solution.

src/tenants/planet/LeaderBoard/components/Stories.tsx
sarveshpro: I solved it for principles and standards

norbertschuler (approved):
Works for me

Issue #1310, opened 2021-10-27
TPO can not save project without donation amount 'treeCost'


norbertschuler:
**Describe the bug**
A TPO can not save a project without specifying a donation amount `treeCost` as the API call POST /app/projects rejects such a request with a 400 error - we need to decide if this function is still wanted and allow a missing `treeCost` parameter again or change the frontend to make the tree costs mandatory.
```
curl 'https://app-staging.plant-for-the-planet.org/app/projects' \
  --data-raw '{"name":"Home","slug":"home","classification":"agroforestry","geometry":{"type":"Point","coordinates":[10.124513002818723,50.79083339388848]},"countTarget":1000,"website":"epublica.de","description":"test","acceptDonations":false,"treeCost":null,"currency":"EUR","visitorAssistance":false,"publish":false}' \
```

**Is it on a Production or Development Build?**
staging

**To Reproduce**
1. Try to add a new project not adding tree costs
2. See error

**Screenshots*
<img width="1572" alt="Bildschirmfoto 2021-10-27 um 12 40 14" src="https://user-images.githubusercontent.com/1532418/139050724-d71d2b48-3110-4604-92eb-51453e340029.png">
*

PR #1341, opened 2021-12-07
treecost optional in add project

sarveshpro:


norbertschuler:
Does this solve #1310?

PR #1348, opened 2021-12-10
Shop Icon in Menu Header

rishsingh07:
image


norbertschuler (changes requested):
There is not enough place on small devices for another navigation item (e.g. iPhone 5). (Maybe we can omit the shop icon in the mobile view where the navigation is at the bottom?)

Bildschirmfoto 2021-12-10 um 08 13 18

norbertschuler (approved):
Works for me. The solution is hiding the shop icon when you shrink the window and not showing it again when the window size got increased in my Chrome browser, but that is a minor issue for me.

PR #1349, opened 2021-12-10
Cypress test fix

rishsingh07:
Fixing test cases failure.
1. Added test id for country
2. Added test id for pending donations


norbertschuler:
There was a 400 error in the staging backend when running Cypress tests - maybe it was temporary, so we should run the tests again:

Bildschirmfoto 2021-12-10 um 12 39 41

norbertschuler:
Cypress tests are currently not running very reliable :-(

PR #1357, opened 2021-12-13
add missing translations and remove 0

sarveshpro:
FIxes -

image

Also added missing resource for `bouquet-donation`


norbertschuler (approved):
- [ ] Pending offline funding donation and also successful funding donation looking good:

Bildschirmfoto 2021-12-13 um 18 15 38
Bildschirmfoto 2021-12-13 um 18 15 07

PR #1360, opened 2021-12-15
use plant date instead of registration date

sarveshpro:
Screenshot 2021-12-15 at 3 29 53 PM


norbertschuler:
- [ ] Yes, it's showing the planting date now, but due to the "load more" button functionality the sorting order of the tree mappings is not correct after loading more:

Bildschirmfoto 2021-12-27 um 13 11 09Bildschirmfoto 2021-12-27 um 13 12 03

tejassonar (approved):

PR #1380, opened 2022-01-07
fix video loading even when skip set to true

sarveshpro:
Test if landing video components load when showVideo(localstorage) is set to false

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -135,16 +135,40 @@ export default function PlanetWeb({ Component, pageProps, err }: any) {
 
   const [showVideo, setshowVideo] = React.useState(true);
 
+  // if localShowVideo is undefined
+  // set localShowVideo is true and show the video
+  // if localShowVideo is true show the video
+  // if localShowVideo is false hide the video
+
+  const [localShowVideo, setLocalShowVideo] = React.useState(false);
+
   React.useEffect(() => {
-    if (typeof window !== 'undefined') {
-      if (localStorage.getItem('hidePreview')) {
-        setshowVideo(false);
+    if (router.pathname === '/') {
+      if (typeof window !== 'undefined') {
+        if (localStorage.getItem('showVideo')) {
+          if (localStorage.getItem('showVideo') === 'true') {
+            setLocalShowVideo(true);
+          } else {
+            setLocalShowVideo(false);
+          }
+        } else {
+          localStorage.setItem('showVideo', 'true');
+          setLocalShowVideo(true);
+        }
       }
-      if (router.pathname !== '/') {
-        setshowVideo(false);
-      }
+    } else {
+      setLocalShowVideo(false);
     }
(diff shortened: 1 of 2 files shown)


sagararyal:
LFTM

norbertschuler (approved):
This works for me, but I do not see what's improved compared to the existing version.

pages/_app.tsx
```diff
@@ -135,16 +135,40 @@ export default function PlanetWeb({ Component, pageProps, err }: any) {
...
-      if (router.pathname !== '/') {
-        setshowVideo(false);
+    if (router.pathname === '/') {
+      if (typeof window !== 'undefined') {
+        if (localStorage.getItem('showVideo')) {
+          if (localStorage.getItem('showVideo') === 'true') {
+            setLocalShowVideo(true);
+          } else {
+            setLocalShowVideo(false);
+          }
```
ankitecd, line 153: You can directly do `setLocalShowVideo(localStorage.getItem('showVideo') === 'true')`, no need for conditional loops

pages/_app.tsx
```diff
@@ -135,16 +135,40 @@ export default function PlanetWeb({ Component, pageProps, err }: any) {
...
+      setLocalShowVideo(false);
     }
   }, []);
+
+  React.useEffect(() => {
+    if (localShowVideo) {
+      setshowVideo(true);
+    } else {
+      setshowVideo(false);
+    }
```
ankitecd, line 169: You can directly do setshowVideo(localShowVideo), no need for conditional loops. 

Also if we already have one variable why use new variable `localShowVideo, setLocalShowVideo` ?

ankitecd (changes requested):
Can optimize the code

pages/_app.tsx
sarveshpro: reduced to setshowVideo(localShowVideo),

To handle the case where component loads before local storage is loaded.

pages/_app.tsx
sarveshpro: local storage values are returned as string.

pages/_app.tsx
```diff
@@ -28,6 +27,11 @@ import ProjectPropsProvider, {
 import UserPropsProvider from '../src/features/common/Layout/UserPropsContext';
 import PlayButton from '../src/features/common/LandingVideo/PlayButton';
 import ErrorHandlingProvider from '../src/features/common/Layout/ErrorHandlingContext';
+import dynamic from 'next/dynamic';
+
+const VideoContainer = dynamic(() => import('../src/features/common/LandingVideo'), {
+  ssr: false,
+});
```
norbertschuler, line 34: If this intended to avoid loading the video, it did not work for me - does it work for you (at least the video is loaded from the cache if it was loaded before):

Bildschirmfoto 2022-01-13 um 17 29 51

PR #1399, opened 2022-01-25
fix project details and remove currency from donation URL

sarveshpro:
Project Details were not visible for conservation projects because of change in API response.


ankitecd (approved):
LGFM

PR #1422, opened 2022-02-26
support external plant locations and improve get requests

sarveshpro:
### Changes
- adapt `getRequest()` to accept version as a parameter
- query params request object in `getAuthenticatedRequest()`
- abstract `getQueryString()`
- use version `1.0.4` in plantLocation APIs to get external plant locations

### DIFF
--- a/src/features/user/TreeMapper/index.tsx
+++ b/src/features/user/TreeMapper/index.tsx
@@ -37,7 +37,7 @@ function TreeMapper({ }: Props): ReactElement {
     if (next && links?.next) {
       const response = await getAuthenticatedRequest(links.next, token, {},
         handleError,
-        '/profile');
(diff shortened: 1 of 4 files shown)


norbertschuler:
Have I correctly understood, you just want to use the endpoints `/treemapper/plantLocations`, `/app/plantLocations` with a version 1.0.4 header?

You now also added an explicit usage of a version 1.0.3 header for all the other GET requests (authorized and non authorized) which did not use such a header before. It seems to work, but it is a change you made worth to note! Have you ask @jmiridis about it?

norbertschuler (approved):
LGFM, but please verify with backend developers that using a `x-accept-versions: 1.0.3` on all other API calls except the ones to use version 1.0.4. is ok to do.

PR #1437, opened 2022-03-10
fix: edit donation modal

sunilsabatp:
Fixes #

Changes in this pull request:
- unified the size of edit donation modal

@


tejassonar:
Giving a fixed height won't work. Can we find a better solution here?
image

PR #1455, opened 2022-03-22
edited terms & condition feature

sunilsabatp:
Fixes #

Changes in this pull request:
- added feature of Terms and condition  system on the complete sign up page
-created a new state for term and condition 
-the user cannot create account without accepting terms and condition


mariahosfeld:
@sunilsabatp Please also include the German translations of the new text components, at the moment it is only in English (same for the error message, not included in the screenshot).
translation_problem

public/static/locales/en/editProfile.json
```diff
@@ -30,5 +30,9 @@
   "publicAccount": "Public Account",
   "accountPrivacyChangeText": "In order to create a profile widget you need to have a public account. By clicking continue you change your privacy settings to public.",
   "continue": "Continue",
-  "cancel": "Cancel"
+  "cancel": "Cancel",
+  "termAndConditionPart1":"I agree to the",
```
norbertschuler, line 34: From the perspective of a translator this is difficult and using one resource with a link placeholder would be better. But I am not sure if we use i18n already in a way to support this e.g. with a Trans component.

norbertschuler:
- [ ] The red error message already gets displayed when you switch e.g. the private checkbox: 

image

And after registration a red error popped up, but I don't have a console on my phone browser to see the related error:

image

norbertschuler:
- [ ] And after registration a red error popped up and this new user cannot access its own profile - the API call POST https://app-staging.plant-for-the-planet.org/app/profile to register the user replied with an 400 error:

```
{"code":400,"message":"Validation Failed","errors":{"errors":
   ["Dieses Formular sollte keine zus\u00e4tzlichen Felder enthalten.","","Extra fields: terms"],
"children":{"type":{},"oAuthAccessToken":{},"name":{},"gender":{},"title":{},"firstname":{},"lastname":{},"address":{},"zipCode":{},"city":{},"country":{},"isPrivate":{},"getNews":{},"currency":{},"locale":{},"url":{},"urlText":{},"bio":{},"target":{},"imageFile":{},"tin":{},"mayContact":{}}}}
```

160249792-699bdeed-8f3a-4313-98c2-83e7b4a4897d

tejassonar (approved):

src/features/user/CompleteSignup/index.tsx
```diff
@@ -501,17 +522,49 @@ export default function CompleteSignup() {
...
           </div>
 
+          <div className={styles.isPrivateAccountDiv}>
+            <div className={styles.mainText}>
+              <label htmlFor={'terms'} style={{ cursor: 'pointer' }}>
+                {t('editProfile:termAndConditionPart1')}
+                <a
+                  className={styles.termsLink}
+                  href={`https://pp.eco/legal/${i18n.language}/terms`}
+                >
```
norbertschuler, line 545: Maybe this link should get opened in a new tab by adding `rel="noopener noreferrer"` and `target={'_blank'}`?

norbertschuler (changes requested):

public/static/locales/en/editProfile.json
norbertschuler: Change this towards using the i18n Trans component to support links.

norbertschuler:
> Please implement interpolation of translation key, so multiple translation keys aren't added for one sentence.

I added a usage of the Trans component from react-i18next: https://react.i18next.com/latest/trans-component.

norbertschuler (approved):

norbertschuler (approved):

sagararyal (approved):

PR #1473, opened 2022-04-19
fix issue of redirecting to profile after login

sunilsabatp:
#issue
Given a user has an account, when they use a link to access a page that requires authentication via log in, then they are redirected to that page after providing their credentials and not to /profile

#Fix
1.fixed the issue by setting redirecting link to localstorage in userLayout


tejassonar:
image

If the route doesn't exist, you can remove it.

norbertschuler (approved):
LGFM

tejassonar (approved):

PR #1482, opened 2022-05-12
terms and condition locale changes

tejassonar:


norbertschuler:
Please do these changes also at LingoHub as otherwise with the next export they will get overwritten and lost again.

norbertschuler:
> Please do these changes also at LingoHub as otherwise with the next export they will get overwritten and lost again.

I updated these German resources in LingoHub now.

PR #1489, opened 2022-05-25
fix the terms and condition missing part

sunilsabatp:
Issue : Terms and Condition part was missing on complete sign up page


mariahosfeld:
I have checked the functionality and it works like expected.

Shreyaschorge (approved):

PR #1492, opened 2022-06-01
fix error message of donation for conservation form

sunilsabatp:
issue: wrong error message  for donation field (i.e cost per tree) for conservation form
fix : change the error message for donation field  (i.e cost per m2) for conservation from


mariahosfeld:
I have tested the functionality, it works as expected.

Shreyaschorge (approved):

PR #1497, opened 2022-06-06
fix the issue of picture of a sample tree does not change

sunilsabatp:
issue: the pictures of a sample tree does not change when selecting different ones.

projectID: ferncliffe-forest-wilding-1

step 1 : go to view mode of project using above projectID
step2 : select different sampletrees and check whether the image in the left side of slide updating or not


tejassonar (approved):

norbertschuler (approved):
Works for me.

mariahosfeld:
Also checked the functionality, works as expected.

PR #1499, opened 2022-06-09
fix the validation error message for employeeCount

sunilsabatp:
issue :  if we fail to put data for employee field and if try submit a form then it was not displaying the validation error for employee field

step: 1 go to any project and open detail analysis form 
2. try to submit form without filling the   employee count  field
3. check whether the error message is displaying or not


norbertschuler (approved):
Bildschirmfoto 2022-06-11 um 10 26 08

It works now, but all the fieldName values should use translation keys instead of raw strings:

```JavaScript
                      message: t('manageProjects:validation', {
                        fieldName: 'Area Protected',
                      }),
```

I started to fix this in #1503 - not sure if other field validation error messages have this error too.

mariahosfeld:
Tested the functionality, works as expected.

PR #1506, opened 2022-06-14
added validation to myspecies

sunilsabatp:
issue: Missing Validation to prevent empty form submission

step 1 : go to tree mapper and select my species
step 2 : then click on add button without filling and selecting any value for the text field
step 3 : check whether it displaying validation error or not


norbertschuler (approved):
LGFM

PR #1507, opened 2022-06-15
Integrate updated subscription API

Shreyaschorge:
Fixes

 -  Breaking recurrent payments page after recurrent top-up of PlanetCash

PR #1513, opened 2022-06-25
fix crash showing recurring donations

norbertschuler:
Fixes crash on production

Changes in this pull request:
- changed project to destination for subscription type


norbertschuler:
Duplicate of #1507

PR #1529, opened 2022-07-28
fix: make config url available in next config

mohitb35:
Fixes issue of /config not being called.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -120,6 +120,7 @@ module.exports = withPlugins([[withBundleAnalyzer]], {
     VERCEL_URL: process.env.VERCEL_URL,
     SITE_IMAGERY_API_URL: SITE_IMAGERY_API_URL,
     WIDGET_URL: process.env.WIDGET_URL,
+    CONFIG_URL: process.env.CONFIG_URL,
   },
   trailingSlash: false,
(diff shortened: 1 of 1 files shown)


next.config.js
```diff
@@ -120,6 +120,7 @@ module.exports = withPlugins([[withBundleAnalyzer]], {
     VERCEL_URL: process.env.VERCEL_URL,
     SITE_IMAGERY_API_URL: SITE_IMAGERY_API_URL,
     WIDGET_URL: process.env.WIDGET_URL,
+    CONFIG_URL: process.env.CONFIG_URL,
```
norbertschuler, line 123: Please also put examples of CONFIG_URL now in our .env files in the repository:
- [.env.development](https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/develop/.env.development)
- [.env.example](https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/develop/.env.example)
- [.env.local.example](https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/develop/.env.local.example)
- [.env.production](https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/develop/.env.production)

Shreyaschorge (approved):

PR #1531, opened 2022-07-29
fix: better 400 error handling for GET requests

mohitb35:
Addresses silent error if a GET request fails with a 400 error.

### DIFF
--- a/src/utils/apiRequests/api.ts
+++ b/src/utils/apiRequests/api.ts
@@ -43,7 +43,7 @@ const handleApiError = (
     }
     console.error('Error 403: Forbidden');
   } else if (error === 400) {
-    if (!result['error_code']) {
+    if (!result || !result['error_code']) {
       if (errorHandler) {
         errorHandler({
           type: 'error',


src/utils/apiRequests/api.ts
```diff
@@ -43,7 +43,7 @@ const handleApiError = (
     }
     console.error('Error 403: Forbidden');
   } else if (error === 400) {
-    if (!result['error_code']) {
+    if (!result || !result['error_code']) {
```
norbertschuler, line 46: Do you know how some "wrong" 400 error without a `result` will be treated in the UI now? What will happen?

mohitb35: You will see "Validation Error" in the error popup that appears on the bottom right, just like earlier (before bulk codes was merged to develop). 

I had added `result` as an argument to handleApiError in order to handle certain error scenarios while creating bulk codes (e.g. insufficient balance) for which Jorgo had modified the API response on my request. An example of such a response is below.

```
{
	"error_type": "payment_error",
	"error_code": "planet_cash_insufficient_credit",
	"parameters": {
		"available_credit": 977100
	}
}
```
The changes in this PR are intended as a temporary fix so that we don't get a silent error that leaves the user wondering.

norbertschuler (approved):
LGFM

PR #1543, opened 2022-08-08
fix: fixes map marker positioning

mohitb35:
- Inline style={{ left: '28px' }} was applied to marker, and was not functional earlier
- `style` prop support was added in `react-map-gl 5.3.19`
- As a result, map markers had moved 28px to the right
- This PR corrects that by removing the inline style


norbertschuler (approved):
LGFM

PR #1545, opened 2022-08-09
fix: remove "Powered by" text from locale files

mohitb35:
Minor text update. Removes "Powered by Plant-for-the-Planet" from locale files as it is not intended to be translated.


mariahosfeld (approved):
LGFM

norbertschuler (approved):

PR #1548, opened 2022-08-11
Minor fixes and updates

mohitb35:
Changes
1. Opens first item in payment history by default
2. Removes intermediate redirect from `/t/userSlug` to `/profile` when the user clicks on the "Me" navbar link


mariahosfeld:
Tested the functionality, works as described.

norbertschuler (approved):
LGFM

PR #1554, opened 2022-08-18
Feature/redeem endpoint

sunilsabatp:
to do : change reedem end-point

previous endpoint : https://app-staging.plant-for-the-planet.org/api/v1.3/de/validateCode

change to : https://app.plant-for-the-planet.org/app/redeem

steps : 1. go to /profile
           2. click on redeem and enter redeem code 
           3. click on validate code


sagararyal:
#### 📝 Changed routes:


| [**/claim/\[type\]/\[code\]**](https://planet-webapp-apud8xs6c-planetapp.vercel.app/claim/[type]/[code]) |
|:-:|
|<a href="https://planet-webapp-apud8xs6c-planetapp.vercel.app/claim/[type]/[code]">Screenshot of /claim/[type]/[code]</a><br /><sup><a href="https://images.weserv.nl?url=https://sjc.microlink.io/LCENmcfGXKD4N6m3kUZLH58XOmvOyxnXYWbI2bjdesA1gsWB_cPTAMElhUDsKeBA45B4urqNk0ZjuujplCVK-w.png">(view full size)</a> |







Commit 520bc1943f21b3de6a14b544b447bfda646884ef (https://planet-webapp-apud8xs6c-planetapp.vercel.app).

mariahosfeld:
Receiving "Clientside-exception" error message after logging in.

mariahosfeld:
Function works for me, please change the following thing:

- [ ] Change text on button from "Validate Code" to "Redeem Code"

mariahosfeld:
Functionality works for me

mohitb35:
1. Treecounter isn't updated after successful redemption.

Screenshot 2022-08-29 at 3 37 26 PM

2. If the code is already redeemed, I am shown an error message, and there is no way to update the code.

Screenshot 2022-08-29 at 3 39 09 PM

public/static/locales/en/me.json
```diff
@@ -231,5 +231,7 @@
   "apiKeyMessage1": "API Key is like a digital password that allows you to act as your profile in any application. It also allows applications to donate on your behalf, use PlanetCash and view transactions.",
   "apiKeyMessage2": "Once you generate an API Key, we will require a two factor authentication for your login. This you’ll be able to setup next time you log in.",
   "apiKeyMessage3": "Regenerating an API Key will invalidate your existing API Key immediately.",
-  "regenerateKey": "Regenerate Key"
+  "regenerateKey": "Regenerate Key",
+  "alreadyRedeemed": "The code has already been redeemed",
+  "invalidCode": "This code is invalid"
```
mohitb35, line 236: Since these are errors specific to redemption, and we already are using redeem.json, perhaps these keys should be moved to redeem.json

src/features/user/Profile/components/RedeemModal.tsx
```diff
@@ -16,6 +16,7 @@ import { UserPropsContext } from '../../../common/Layout/UserPropsContext';
 import { ErrorHandlingContext } from '../../../common/Layout/ErrorHandlingContext';
 import ShareOptions from '../../../common/ShareOptions/ShareOptions';
 import { styled } from '@mui/material';
+import { useEffect } from 'react';
```
mohitb35, line 19: combine this with other imports from 'react'

src/features/user/Profile/components/RedeemModal.tsx
```diff
@@ -103,7 +101,7 @@ export default function RedeemModal({
   async function redeemCode() {
```
mohitb35, line 101: This is not used any more, and can be removed.

src/features/user/Profile/components/RedeemModal.tsx
```diff
@@ -266,17 +273,25 @@ export default function RedeemModal({
           <div className={styles.modal}>
             {codeValidated && validCodeData ? (
               <>
-                <div className={styles.codeTreeCount}>
+                <div
+                  className={styles.codeTreeCount}
+                  style={{ fontSize: '2rem' }}
```
mohitb35, line 278: Avoid using inline styles (esp. if there's already a class defined).

src/features/user/Profile/components/RedeemModal.tsx
```diff
@@ -74,23 +77,18 @@ export default function RedeemModal({
   async function validateCode(data: any) {
```
mohitb35, line 77: We were earlier separating validate and redeem into 2 steps, and are now combining it into one step. Should the `validateCode` function be renamed?

pages/claim/[type]/[code].tsx
```diff
@@ -94,20 +94,21 @@ function ClaimDonation({}: Props): ReactElement {
...
           setErrorMessage(res.message);
           setIsUploadingData(false);
-        } else if (res.status === 'error') {
-          setErrorMessage(res.errorText || t('me:wentWrong'));
+        } else if (res.error_code === 'already_redeemed') {
+          setErrorMessage(res.error_code || t('me:wentWrong'));
           setIsUploadingData(false);
-        } else if (res.status === 'success') {
+        } else if (res.error_code === 'invalid_code') {
+          setErrorMessage(res.error_code);
```
Comment (mohitb35) [inline pages/claim/[type]/[code].tsx:110]: errorMessage text needs to be translated. Currently the key (e.g. 'already_redeemed') will display instead of the desired text.

pages/claim/[type]/[code].tsx
```diff
@@ -94,20 +94,21 @@ function ClaimDonation({}: Props): ReactElement {
...
         token,
         handleError
       ).then((res) => {
         if (res.code === 401) {
           setErrorMessage(res.message);
           setIsUploadingData(false);
-        } else if (res.status === 'error') {
-          setErrorMessage(res.errorText || t('me:wentWrong'));
+        } else if (res.error_code === 'already_redeemed') {
+          setErrorMessage(res.error_code || t('me:wentWrong'));
```
Comment (mohitb35) [inline pages/claim/[type]/[code].tsx:107]: Earlier, t('me:wentWrong') was used as a catch all message in case no errorText was set.

The error handling logic has changed now, so the way this is used will change as well.

src/features/user/Profile/components/RedeemModal.tsx
```diff
@@ -74,23 +77,18 @@ export default function RedeemModal({
...
-          setErrorMessage(res.message);
+      postAuthenticatedRequest(`/app/redeem`, submitData, token).then((res) => {
+        if (res.error_code === 'already_redeemed') {
+          setErrorMessage(t('me:alreadyRedeemed'));
           setIsUploadingData(false);
-        } else if (res.status === 'error') {
-          setErrorMessage(res.errorText || t('me:wentWrong'));
+        } else if (res.error_code === 'invalid_code') {
+          setErrorMessage(t('me:invalidCode'));
           setIsUploadingData(false);
```
mohitb35, line 90: 1. Both errors have similar handling, so you could combine them into one block. 
2. What happens if an unanticipated error is returned? It seems to me that the app will not inform the user about the error in that case.

src/features/user/Profile/components/RedeemModal.tsx
```diff
@@ -138,10 +136,19 @@ export default function RedeemModal({
     }
   }
 
+  useEffect(() => {
+    if (errorMessage || validCodeData) {
+      setAfterValidateCodeButton(true);
+    }
+  }, [errorMessage, validCodeData]);
+
```
mohitb35, line 144: I think `afterValidateCode` isn't required to be set in the state. It is only being used to store the value of `errorMessage || validCodeData`. If you need a variable to store this, you can just use a regular variable (instead of storing it in state and triggering a rerender of the component).

src/features/user/Profile/components/RedeemModal.tsx
```diff
@@ -266,17 +273,25 @@ export default function RedeemModal({
...
-                      count: Number(validCodeData.treeCount),
+                      count: Number(validCodeData.units),
                     })}
                   </span>
                 </div>
+                <div>
+                  <span className={styles.codeTreeCount}>
+                    {t('redeem:successfullyRedeemed')}
+                  </span>
+                </div>
```
mohitb35, line 294: It is a little confusing to use `styles.codeTreeCount` here, as that suggests that the data shown here is tree count. Either use a more general classname, or create a separate class.

src/features/user/Profile/components/RedeemModal.tsx
```diff
@@ -288,69 +303,89 @@ export default function RedeemModal({
...
+                  >
+                    {isUploadingData ? (
+                      <div className={styles.spinner}></div>
+                    ) : (
+                      t('redeem:exitToProfile')
+                    )}
+                  </button>
+                ) : (
+                  ''
+                )}
```
mohitb35, line 321: When you modify existing code, make sure that the naming used makes sense in this context. For example, the `id={'redeemModalCont'}` does not make sense here, as this is not a CONTINUE action any more. It may make sense to change that to `id={'redeemModalClose'}`

src/features/user/Profile/components/RedeemModal.tsx
```diff
@@ -41,8 +42,10 @@ export default function RedeemModal({
   const [errorMessage, setErrorMessage] = React.useState();
   const [codeValidated, setCodeValidated] = React.useState(false);
   const [isUploadingData, setIsUploadingData] = React.useState(false);
-  const [validCodeData, setValidCodeData] = React.useState();
+  const [validCodeData, setValidCodeData] = React.useState('');
   const [codeRedeemed, setCodeRedeemed] = React.useState(false);
```
mohitb35, line 46: Does `codeRedeemed` have a purpose any more?

src/features/user/Profile/components/RedeemModal.tsx
sunilsabatp: sure

mariahosfeld:
I checked the newly added functionalities, they work as expected for me.

mariahosfeld:
The functionality works as expected.

mariahosfeld:
The functionality works as expected.

src/features/user/Profile/components/RedeemModal.tsx
mohitb35: Please remove this completely, don't leave it commented. If you need to go back to check it out, version control holds the history of the file.

src/features/user/Profile/components/RedeemModal.tsx
mohitb35: removed

mohitb35: removed

mohitb35:
Please make the changes made for the redeem modal in `pages/claim/[type]/[code].tsx`
 as well. There are 2 places that code redemption can happen. Currently some of the code here is the old functionality.

For example, the `redeemCode` function is still present here and calls the old endpoint.

```
  async function redeemCode(code: any, type: any) {
...
postAuthenticatedRequest(
        `/api/v1.3/${userLang}/convertCode`,
        submitData,
        token,
        handleError
      ).then(...
...
}
```

src/theme/global.scss
```diff
@@ -269,6 +269,25 @@ video {
   cursor: pointer;
 }
 
+.combineButton {
```
mohitb35, line 272: I think this class is unused. Remove if not needed

src/theme/global.scss
```diff
@@ -269,6 +269,25 @@ video {
...
+  display: flex;
+}
+
+.exitToProfile{
+  max-width: 150px;
+  margin-top: 24px;
+ }
```
mohitb35, line 279: I think this class is unused. Remove if not needed

src/theme/global.scss
```diff
@@ -269,6 +269,25 @@ video {
...
+
+.redeemCode {
+  max-width: 200px;
+  margin-top: 24px;
+}
+
+.redeemAnotherCode{
+  max-width: 220px;
+  margin-top: 24px;
+}
```
mohitb35, line 289: Why are we using 2 different classes? It may be better to use a common class in both places and prescribe a padding for the button instead of a max width.

Also, since these classes are specific to the Redeem Modal, perhaps they should be used in the modular scss file RedeemModal.scss

src/features/user/Profile/components/RedeemModal.tsx
```diff
@@ -288,69 +327,73 @@ export default function RedeemModal({
...
+                    id={'redeemCodeModal'}
+                    onClick={handleSubmit(redeemCode)}
+                    className="primaryButton redeemCode"
+                  >
+                    {isUploadingData ? (
+                      <div className={styles.spinner}></div>
+                    ) : (
+                      t('redeem:redeemCode')
+                    )}
+                  </button>
```
mohitb35, line 395: Why are we using 2 different classes? It may be better to use a common class in both places and prescribe a padding for the button instead of a max width.

src/tenants/common/Home/index.tsx
```diff
@@ -14,9 +14,14 @@ const { useTranslation } = i18next;
 interface Props {
   leaderboard: any;
   tenantScore: any;
+  updateTenantScore: string;
```
mohitb35, line 17: I'm not clear how this is used. I don't see any place in the app that uses updateTenantScore

src/features/user/Profile/components/RedeemModal.tsx
```diff
@@ -266,17 +283,39 @@ export default function RedeemModal({
...
+                    {t('redeem:successfullyRedeemed')}
+                  </span>
+                </div>
+
+                <button
+                  className="primaryButton redeemAnotherCode"
+                  onClick={afterSuccessfulOfOneCode}
+                >
+                  {t('redeem:redeemAnotherCode')}
+                </button>
```
mohitb35, line 318: If possible, I think we could have a common structure here and from lines 333-365

mohitb35 (commented):
@sunil-sabat Please see my comments.

In addition, it seems like there is a lot of duplication within the RedeemModal file. Time permitting, it may make sense to address that now. If not, it can be a point for the future.

src/types/Interface.tsx
```diff
@@ -0,0 +1,41 @@
...
+  email: string;
+  displayName: string;
+  score: Score;
+  isReviewer: boolean;
+  isPrivate: boolean;
+  getNews: boolean;
+  address: Address;
+  locale: string;
+  planetCash: PlanetCash | null;
+}
```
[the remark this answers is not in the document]
mohitb35: I see that this has been moved to `redeem.d.ts`. However, the same interfaces/types were already present in `user.d.ts`. I can't see a reason to have this in 2 different places. Please use/adapt the interfaces already defined in `user.d.ts`.

norbertschuler:
> steps : 1. go to /profile 2. click on redeem and enter redeem code 3. click on validate code

- [ ] I tested this with 2 chocolate codes and I wondered why the codes got instantly redeemed after entering them. Before I was shown what the code resolve to and then hit another button to decide to redeem them. Now this has happened in one step, so you cannot decide against redeeming a code any more after seeing what it resolves to (e.g. if you accidentally enter the  code of someone else).

mohitb35:
The layout is not consistent across the 3 different states of the modal: input form, invalid/used code, success.

Screenshot 2022-09-26 at 3 47 23 PM
Screenshot 2022-09-26 at 3 47 30 PM
Screenshot 2022-09-26 at 3 47 39 PM

src/features/user/Profile/styles/RedeemModal.module.scss
```diff
@@ -21,30 +21,27 @@
 }
 
 .fixModal {
-  max-height: 279px;
+  max-height: 260px;
```
mohitb35, line 24: I think the `max-height` can be removed. Keeping this has the effect of reducing the button height when there is an error message below the text field.

src/features/user/Profile/components/RedeemModal.tsx
```diff
@@ -100,113 +100,103 @@ export default function RedeemModal({
...
-                  {t('common:tree', {
-                    count: Number(validCodeData.units),
+        {isCodeRedeemed && validCodeData ? (
+          // after successful redeeemed code
+          <div className={`${styles.modal} ${styles.fixModal}`}>
+            <div className={styles.crossButtonDiv}>
+              <button className={styles.crossButton} onClick={closeRedeem}>
+                <CancelIcon color={styles.primaryFontColor} />
+              </button>
+            </div>
```
mohitb35, line 110: This is common for both true and false state of the condition `isCodeRedeemed && validCodeData` - you could move the condition below this line and remove the repetitive code.

mohitb35 (changes requested):
@sunilsabatp Have requested minor changes. Once done, this is good to go.

mohitb35 (approved):

mohitb35:
Looks good to me.

pages/claim/[type]/[code].tsx
```diff
@@ -145,34 +145,34 @@ function ClaimDonation({}: Props): ReactElement {
...
-          setIsUploadingData(false);
-        } else if (res.response.status === 'success') {
-          setCodeRedeemed(true);
-          setIsUploadingData(false);
-          setCodeValidated(false);
-        }
-      });
-    }
-  }
+  // async function redeemCode(code: any, type: any) {
```
Comment (norbertschuler) [inline pages/claim/[type]/[code].tsx:148]: I think we can remove unused code. We have git if we want to retrieve it again.

PR #1563, opened 2022-09-07
New Message and Error for Delete Profile

prachigarg19:
Displays error  if user tries to delete their account with active subscriptions.

Changes in this pull request:
- Display message to make sure user have cancelled all subscriptions before delete button.
- Create modal which pops up when user continues deleting their account with active subscriptions: 
  If continue is clicked, the user is redirected to profile/recurrency. 
- After cancelling all subscriptions, user can delete the account successfully.

### DIFF
--- a/src/features/common/Layout/CustomModal/index.tsx
+++ b/src/features/common/Layout/CustomModal/index.tsx
@@ -5,8 +5,8 @@ import styles from './CustomModal.module.scss';
 
 interface Props {
   isOpen: boolean;
-  onClick1: () => void;
-  onClick2: () => void;
+  handleContinue: () => void;
+  handleCancel: () => void;
   buttonTitle: string;
   modalTitle: string;
   modalSubtitle: string;
@@ -16,8 +16,8 @@ const { useTranslation } = i18next;
 
 export default function CustomModal({
   isOpen,
-  onClick1,
-  onClick2,
+  handleContinue,
+  handleCancel,
   buttonTitle,
   modalTitle,
   modalSubtitle,
@@ -40,14 +40,14 @@ export default function CustomModal({
           <button
             id={'Continue'}
             className={styles.continueButton}
-            onClick={onClick1}
+            onClick={handleContinue}
           >
             {buttonTitle}
           </button>
           <button
             id={'Cancel'}
             className={styles.cancelButton}
-            onClick={onClick2}
+            onClick={handleCancel}
           >
             {t('editProfile:cancel')}
           </button>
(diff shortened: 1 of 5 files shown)


mariahosfeld:
- [ ] Please change the wording of "Before proceeding, make sure you've deleted all subscriptions before ." to "Before proceeding, make sure you've cancelled all subscriptions."

mariahosfeld:
The basic functionality works, but please make the error message when a Subscription is still there more prominent, e.g. in red and bold type, so that the user will spot it immediately.

norbertschuler (commented):
LGFM, first I was puzzled if "continue" means continue to delete my profile anyway, but then reading the text I understood continue to be redirected to the subscriptions to be able to cancel them. Maybe the button could be renamed to "show my subscriptions"?

mariahosfeld:
Functionality works as expected for me.

src/features/common/Layout/CustomModal/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+          <div className={styles.modal__subtitle}>{modalSubtitle}</div>
+        </div>
+        <div
+          className={styles.buttonContainer}
+          style={{ justifyContent: 'center' }}
+        >
+          <button
+            id={'Continue'}
+            className={styles.continueButton}
+            onClick={onClick1}
```
Shreyaschorge, line 43: @prachigarg19 Please give it a meaningful name, maybe something like `handleContinue`

src/features/user/Settings/DeleteProfile.tsx
```diff
@@ -18,25 +19,36 @@ export default function DeleteProfile({}: any) {
...
   const [canDeleteAccount, setcanDeleteAccount] = React.useState(false);
 
   const handleDeleteAccount = () => {
     setIsUploadingData(true);
     deleteAuthenticatedRequest('/app/profile', token, handleError).then(
       (res) => {
-        if (res !== 404) {
-          logoutUser(`${process.env.NEXTAUTH_URL}/`);
-        } else {
+        if (res == 400) {
```
Shreyaschorge, line 29: @prachigarg19 There could be multiple factors that might result into a 400 response. Before opening the Cancel Subscriptions modal, along with 400 response please check the response (i.e error_code, error_type) returned from the server. 
image

src/features/common/Layout/CustomModal/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+          >
+            {buttonTitle}
+          </button>
+          <button
+            id={'Cancel'}
+            className={styles.cancelButton}
+            onClick={onClick2}
```
Shreyaschorge, line 50: @prachigarg19 Please give it a meaningful name, maybe something like `handleCancel`

Shreyaschorge (changes requested):
Functionally it works. There are only minor changes that needs to be handled.

Shreyaschorge (approved):
LGFM. Only thing is we need to manage our errors in a better way.

PR #1564, opened 2022-09-09
Makes the toggle Project Details/List button translatable

mohitb35:
Makes the toggle Project Details/List button translatable

### DIFF
--- a/public/static/locales/de/maps.json
+++ b/public/static/locales/de/maps.json
@@ -41,5 +41,9 @@
   "today": "Heute",
   "yesterday": "Gestern",
   "clickForDetails": "Klicke für Details",
-  "treeMapper": "TreeMapper"
+  "treeMapper": "TreeMapper",
+  "showProjectList": "Projektliste anzeigen",
+  "showProjectDetails": "Projektdetails anzeigen",
+  "hideProjectList": "Projektliste ausblenden",
+  "hideProjectDetails": "Projektdetails ausblenden"
 }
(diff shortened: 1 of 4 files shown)


norbertschuler (approved):
LGFM, the button is shown capitalized due to https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/4d0c24fbef624473fa478849683d9459617c8c22/src/features/common/InputTypes/MuiButton.tsx#L7 - so the German verb is written capitalized which is wrong, but I do not know what else would be affected if we change the button style.

Bildschirmfoto 2022-09-18 um 10 59 08

mohitb35:
> LGFM, but the button text is shown capitalized due to
> 
> https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/4d0c24fbef624473fa478849683d9459617c8c22/src/features/common/InputTypes/MuiButton.tsx#L7
> 
> - so the German verb is written capitalized which is wrong, but I do not know what else would be affected if we change the button style.
> Bildschirmfoto 2022-09-18 um 10 59 08

The two buttons to toggle Project Details and Project List are the only ones that will be immediately affected by this. What is the correct way to show this text?

norbertschuler:
> > LGFM, but the button text is shown capitalized due to
> 
> The two buttons to toggle Project Details and Project List are the only ones that will be immediately affected by this. What is the correct way to show this text? 

Just don't use the style capitalization and write the text correctly in the resource file.

PR #1567, opened 2022-09-13
Add beta tag for Bulk Codes submenu

mohitb35:
- Shows tags for sub menu items
- Adds beta tag for Bulk Codes sub menu


mariahosfeld (approved):
Looks good to me.

PR #1570, opened 2022-09-19
Center map

prachigarg19:


mariahosfeld:
Tested the functionality, works for me as expected.

src/features/projects/components/ProjectsMap.tsx
```diff
@@ -161,6 +161,17 @@ export default function ProjectsMap(): ReactElement {
...
+      const newViewport = {
+        ...viewport,
+        latitude: 36.96,
+        longitude: 0,
+      };
+      setViewPort(newViewport);
+    }
+  }, [showProjectList, router]);
+
   return (
```
mohitb35, line 175: @prachigarg19 
1. Not clear why `router` is a dependency here.
2. Please check the combination of `embed` and `showProjectList` as `showProjectList` is not supposed to affect anything without `embed = true`

mohitb35 (changes requested):
@prachigarg19 I've taken a look and suggested some minor changes.

Also, I think the new viewport gets lost if the user goes to a single project page and comes back (the map no longer remains centered in this case)

src/features/projects/components/ProjectsMap.tsx
prachigarg19: ` router` works for initial render.

mariahosfeld:
I checked again, the project_list=false context is lost when going back from a single project view.

mariahosfeld:
Functioniality now works for me.

norbertschuler (approved):
This works, but make the map even more Europe centered as before as the map always get reseted to these coordinates when hiding or showing projects even if you currently looking at another spot of the world.

PR #1572, opened 2022-09-21
fix: removes text transform from MuiButton text

mohitb35:
- shows text from translation file as is
- so that German text is not incorrectly capitalized

Screenshot 2022-09-21 at 11 38 33 AM
Screenshot 2022-09-21 at 11 38 05 AM


mariahosfeld:
Checked the functionality, works as expected.

mariahosfeld (approved):

PR #1574, opened 2022-09-23
Implement Re-measurement data in the frontend

prachigarg19:
Changes in this pull request:
- Add “Latest measurements” below sample trees heading
- If **`status = dead`** → Show only status and value of status  reason,  separated with the dot
- Implement the same for the sample tree detail view as well.

@


mariahosfeld:
Functionality works as expected.

norbertschuler:
Any good idea how to best test this? I have no RO having dead trees in the staging backend.

Shreyaschorge (approved):
LGFM!

Issue #59, opened 2020-08-13
Scroll up and show only cost button in single project details page


harshvitra:
On project details page if user scrolls up the header should be shown like the below

![Screen Shot 2021-07-13 at 2 11 52 PM](https://user-images.githubusercontent.com/5401575/125449562-61f3bbb0-59d4-4299-afd9-768486cd835e.png)

PR #1599, opened 2022-10-27
Scroll up and show only cost button in single project details page

prachigarg19:
Solves https://github.com/Plant-for-the-Planet-org/planet-webapp/issues/59

Changes in this pull request:
-
-
-

@


mariahosfeld:
Generally the functionality works, in some cases the scrolling stops to early, so that the last lines are not completely visible. Please make sure that everything will be visible when the donor scrolls.

mariahosfeld:
The issue where a part of the container for the information about a project has been cut off is now solved.
Another issue:
When I am in the wrong part of the description (container with the donate button), and start scrolling, it just stops and does not compress anything. Ideal would be - if possible - I the pointer is above one of the containers I can scroll and always have the same behavior: The top container compresses and the lower scrolls.

PR #1606, opened 2022-11-08
Add texts to inform RO's to create projects in English only.

prachigarg19:
Add texts to inform RO's to create projects in English only.


mariahosfeld:
Looks good to me.

norbertschuler (approved):
Works for me, but for this change we really should at least wait for the German translation before doing a release.

PR #1607, opened 2022-11-09
Feature/show top projects list

mohitb35:
1. Enables the top projects list (formerly featured list) if the env variable NEXT_PUBLIC_SHOW_TOP_PROJECTS is true. 
2. Also updates the logic to populate the list to use `isApproved` and `isTopProject` instead of `isFeatured`. List is not shown if 0 projects satisfy the condition.


prachigarg19 (approved):

src/features/projects/screens/Projects.tsx
```diff
@@ -38,15 +37,14 @@ function ProjectsList({
   const [hideSidebar, setHideSidebar] = React.useState(isEmbed);
   const { t, ready } = useTranslation(['donate', 'country', 'maps']);
 
-  const featuredList = process.env.NEXT_PUBLIC_FEATURED_LIST;
+  const topProjectsList = process.env.NEXT_PUBLIC_TOP_PROJECTS_LIST;
 
-  const showFeaturedList =
-    featuredList === 'false' || featuredList === '0' ? false : true;
+  const showTopProjectsList = topProjectsList === 'true' ? true : false;
```
norbertschuler, line 42: The former env `NEXT_PUBLIC_FEATURED_LIST` resulted into showing the list as default if not explicitly disabled by this env. This new env `NEXT_PUBLIC_TOP_PROJECTS_LIST`  is hiding the list on default and only showing it if the env is set to True.

This is fine for me, but please think about checking the environment setup of all our tenants on Netlify and Heroku to make sure the configuration is as wanted, e.g. for https://trees-weareams.netlify.app/ (Netlify) and https://forest.lacoquetakids.com/ (Heroku) the list is currently shown as no NEXT_PUBLIC_FEATURED_LIST was set or set to True.

norbertschuler (approved):
LGFM, please read my comment regarding deployment.

src/features/projects/screens/Projects.tsx
mariahosfeld: I have checked, the list should be shown by default.

prachigarg19 (approved):

prachigarg19 (approved):

src/features/projects/screens/Projects.tsx
mohitb35: @norbertschuler We've kept the same logic in place as before. `NEXT_PUBLIC_TOP_PROJECTS_LIST` must be `true` to display the list. This is being done only for `planet` and `ttc` for now, so we've made sure to set the variable in those environments.

mariahosfeld (commented):
Functionality looks good to me

PR #1617, opened 2022-11-18
tenant config language

sunilsabatp:
fix -> applied tenant preference languages

### DIFF
--- a/src/features/common/Layout/QueryParamsContext.tsx
+++ b/src/features/common/Layout/QueryParamsContext.tsx
@@ -59,7 +59,10 @@ const QueryParamsProvider: FC = ({ children }) => {
       const userBrowserLanguage = navigator.language ?? navigator.languages[0];
       // checks whether tenant supported locale matches the user browser preference locale
       const languageMatched = tenantSupportedLocale.find((locale) => {
-        return locale === userBrowserLanguage[0] + userBrowserLanguage[1];
+        return (
+          locale[0] + locale[1] ===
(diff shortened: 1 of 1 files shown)


src/features/common/Layout/QueryParamsContext.tsx
```diff
@@ -53,38 +55,24 @@ const QueryParamsProvider: FC = ({ children }) => {
...
-      setLanguage(query.locale);
-    } else {
-      if (localStorage?.getItem('language') === null) {
-        const userPreferenceLanguage =
-          navigator.language ?? navigator.languages;
+    if (localStorage.getItem('language') === null) {
+      const userBrowserLanguage = navigator.language ?? navigator.languages;
+      // checks whether tenant supported locale matches the user browser preference locale
+      const languageMatched = tenantSupportedLocale.find((locale) => {
+        return locale === userBrowserLanguage;
```
norbertschuler, line 62: This does not work as navigator.language can return `de-DE` and navigator.languages an array of locales, compare  https://developer.mozilla.org/en-US/docs/Web/API/Navigator/languages.

Try something like 
```
const userBrowserLanguage = (navigator.language && navigator.language.substring(0,2)) ?? (navigator.languages && navigator.languages[0].substring(0,2));
```

norbertschuler (changes requested):

norbertschuler (approved):
Works for me. (And I do not think we need locale support in our URLs.)

Issue #1629, opened 2022-12-02
NOT URGENT: Stories for components dependent on next-i18next do not render


mohitb35:
After upgrading to next 12, we are directly using the `useTranslation` hook from `next-i18next`. This does not seem to work correctly with storybook and stories dependent on translations are not rendering currently.

PR #1630, opened 2022-12-02
Fixes failed storybook build process/Chromatic 

mohitb35:
- updates storybook and addons to latest stable versions
- updates styled-jsx from 3.4.7 to 5.0.0

While this fixes the storybook deployment issue on Chromatic, the stories are currently not rendering correctly wherever translation is involved using `next-i18next`. I've created an issue #1629 that can be tackled when we need to use storybook.


.github/workflows/chromatic.yml
```diff
@@ -9,7 +9,7 @@ on:
     branches:
       - develop
       - main
-      - feature/bulk-code-generation
+      - feature/fix-storybook-module-not-found
```
norbertschuler, line 12: Testing this on GitHub as I do not have the token: ✖ Missing project token

norbertschuler (approved):
Seems to run again.

PR #1634, opened 2022-12-08
Hotfix: Update bulk code docs - replace csv template with xlsx

mohitb35:
Allow both csv and xlsx files to be uploaded for bulk codes (import method) 

Context: 
Microsoft excel does not automatically support csv in german encoding and also the german characters are messed with csv most of the time


mariahosfeld:
Functionality of xlsx upload works for me.

PR #1635, opened 2022-12-09
fix route

sunilsabatp:
2 issues have been covered in this pr
1`.profile/history/` when click to (`Restoring Hope for the Atlantic forest of`) it takes the user to `/savingnaturebrazil?site=site_dhGZUIZnE4vStzW` . then after clicking back button it takes the user to landing page instead of  profile/history/

2.route→` profile/treemapper/impor`t add cursor(pointer) on hover over delete icon

note-> login using RO account

### DIFF
--- a/src/features/projects/screens/SingleProjectDetails.tsx
+++ b/src/features/projects/screens/SingleProjectDetails.tsx
@@ -17,7 +17,6 @@ import { ProjectPropsContext } from '../../common/Layout/ProjectPropsContext';
 import ProjectTabs from '../components/maps/ProjectTabs';
 import PlantLocationDetails from '../components/PlantLocation/PlantLocationDetails';
 import { ParamsContext } from '../../common/Layout/QueryParamsContext';
-import { useAuth0 } from '@auth0/auth0-react';
 
 const TimeTravel = dynamic(() => import('../components/maps/TimeTravel'), {
   ssr: false,
@@ -35,7 +34,6 @@ const ImageSlider = dynamic(
 
 function SingleProjectDetails({}: Props): ReactElement {
   const router = useRouter();
-  const { isAuthenticated } = useAuth0();
 
   const { t, ready } = useTranslation(['donate', 'common', 'country', 'maps']);
   const {
@@ -105,7 +103,7 @@ function SingleProjectDetails({}: Props): ReactElement {
                   ? `?embed=true&callback=${callbackUrl}`
                   : '?embed=true'
               }`
-            : ``
+            : `${document.referrer === '' ? '' : window.history.go(-2)}`
         }`
       );
     }
--- a/src/features/projects/screens/SingleProjectDetails.tsx
+++ b/src/features/projects/screens/SingleProjectDetails.tsx
@@ -81,11 +81,6 @@ function SingleProjectDetails({}: Props): ReactElement {
     plantLocation: hoveredPl ? hoveredPl : selectedPl,
   };
 
-  const routeContainProfile = localStorage
-    .getItem('redirectLink')
-    ?.split('/')
-    .some((element) => element === 'profile');
-
   const goBack = () => {
     if (selectedPl || hoveredPl) {
       setHoveredPl(null);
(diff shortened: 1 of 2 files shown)


src/features/projects/screens/SingleProjectDetails.tsx
```diff
@@ -101,6 +100,8 @@ function SingleProjectDetails({}: Props): ReactElement {
         }`
       );
     } else {
+      const link = localStorage.getItem('redirectLink');
+      const previousPath = link?.replace(link[0], '');
```
norbertschuler, line 104: I visited https://planet-webapp-lc2mzz1t5-planetapp.vercel.app/profile/treemapper/import resulting into storing this value into the locale storage value `redirectLink` (took this as an extreme example to better make my point). When I now visit https://planet-webapp-lc2mzz1t5-planetapp.vercel.app/science-forest-planbe?site=site_aufcHOggZ6setis and click the back link, I get redirected to https://planet-webapp-lc2mzz1t5-planetapp.vercel.app/profile/treemapper/import - I don't think that's what we want.

norbertschuler (changes requested):
Added use case example where this solution does not really work well.

src/features/projects/screens/SingleProjectDetails.tsx
sunilsabatp: I did this  to redirect the user back to that page from where he visited to  /:projectID

src/features/projects/screens/SingleProjectDetails.tsx
```diff
@@ -109,7 +103,7 @@ function SingleProjectDetails({}: Props): ReactElement {
                   ? `?embed=true&callback=${callbackUrl}`
                   : '?embed=true'
               }`
-            : ''
+            : `/profile`
```
norbertschuler, line 106: I don't think this is a good solution as well. Of course from "/profile/history" jumping to a project this PR will return to "/profile" by this change. But the back link on a project site normally should navigation back to the home page of planet-webapp (just "/") and not forcing someone to login to show a profile. Especially for people visiting the site not having any profile at all this is extremely irritating!

norbertschuler (changes requested):

src/features/projects/screens/SingleProjectDetails.tsx
```diff
@@ -109,7 +110,7 @@ function SingleProjectDetails({}: Props): ReactElement {
                   ? `?embed=true&callback=${callbackUrl}`
                   : '?embed=true'
               }`
-            : ''
+            : `${isAuthenticated ? (routeContainProfile ? 'profile' : '') : ''}`
```
norbertschuler, line 113: Sorry, but it still does not make sense for me to default the navigation for logged in users to their profile if visiting a project from the homepage and then use the button to navigate back:

https://user-images.githubusercontent.com/1532418/212468061-910ea137-04d1-45ed-a3f5-0e0262e09755.mp4

norbertschuler (changes requested):

src/features/projects/screens/SingleProjectDetails.tsx
sunilsabatp: I have removed of mine  previous implementation , now the back button will redirect user only   to`' / `' for user coming from using any route

norbertschuler (approved):
This solution works for me in latest Chrome on MacOS and iOS.

mariahosfeld:
@sunilsabatp , this is no longer working.

sunilsabatp:
> @sunilsabatp , this is no longer working.

@mariahosfeld  I have tested this feature using  firefox , it's working fine for me

mariahosfeld:
> > @sunilsabatp , this is no longer working.
> 
> @mariahosfeld I have tested this feature using firefox , it's working fine for me

@sunilsabatp , it still takes me to /undefined when using the back button in firefox. Using Edge it is working as expected.

mariahosfeld:
> > > @sunilsabatp , this is no longer working.
> > 
> > 
> > @mariahosfeld I have tested this feature using firefox , it's working fine for me
> 
> @sunilsabatp , it still takes me to /undefined when using the back button in firefox. Using Edge it is working as expected.

Tested again, on Firefox this problem persists, when testing on Edge it is working accordingly.

mariahosfeld:
I no longer see the undefined issue. But at first I only see the zoom out to the map before the URL changes and takes me back where it should. Maybe we can speed this up a bit, as it was - at least to me - a bit confusing what was happening.

norbertschuler (approved):
In my latest Chrome on MacOS it works this way.

PR #1637, opened 2022-12-09
update german translations for bulk codes

mohitb35:
Updates german translations for bulk codes after adding import excel functionality


mariahosfeld (approved):
Looks good to me.

PR #1640, opened 2022-12-12
Consolidate MUI themes 

prachigarg19:
The theme and styles are shifted from muiThemeNew to themeStyles.ts


src/theme/themeProperties.ts
```diff
@@ -11,6 +11,10 @@ const themeProperties = {
     fontXLarge: '26px',
     fontXXLarge: '36px',
     fontXXXLarge: '48px',
+    fontRemLarge: '2.25rem',
+    fontRemMedium: '1.5rem',
+    fontRemSmall: '1.125rem',
+    fontRemXSmall: '0.875rem',
```
mohitb35, line 17: Could you add a comment here as a reminder to update this later and use only one system (either px or rem)? Perhaps we should use rem everywhere as that is a relative unit.

src/theme/themeStyles.ts
```diff
@@ -6,18 +6,93 @@ const materialTheme = createTheme({
...
+        containedSizeSmall: {
+          padding: '5px 10px',
+          borderRadius: 14,
+        },
+        text: {
+          padding: 0,
+          borderRadius: 0,
+          fontSize: 'inherit',
+          fontWeight: theme.fontWeight,
+        },
```
mohitb35, line 75: There seems to be a change in the styling of text buttons seen in Recurring Donations on clicking Edit Donation and viewing the calendar picker. This was also seen for Treemapper>Import data

*Before*
Screenshot 2022-12-16 at 11 47 27 AM

*After*
Screenshot 2022-12-16 at 11 47 31 AM

src/theme/muiThemeNew.ts
```diff
@@ -6,7 +6,7 @@ declare module '@mui/material/styles/createPalette' {
     base?: string;
   }
 }
```
mohitb35, line 8: Don't forget to move this code before deleting this file

mohitb35 (changes requested):

mohitb35:
@prachigarg19 Mostly looks fine, there are small changes suggested.

Also, please replace the Mui theme in `.storybook/preview.js` as well.

src/theme/themeStyles.ts
mohitb35: @prachigarg19 This likely comes from the button style overrides added to themeStyle.ts - inspect the elements and try to figure out where the styles are coming from.

```
MuiButton: {
      styleOverrides: {
        root: {
          borderRadius: 28,
          textTransform: 'capitalize',
          padding: '12px 30px',
        },
        containedSizeSmall: {
          padding: '5px 10px',
          borderRadius: 14,
        },
        text: {
          padding: 0,
          borderRadius: 0,
          fontSize: 'inherit',
          fontWeight: theme.fontWeight,
        },
        
```

mohitb35: This is done for the EditModal, but is pending for other places where MuiDatePicker is used.

mohitb35 (changes requested):
Looks good except for 2 points.

1. The space below the Cancel / OK buttons used in MobileDatePicker throughout the application should be uniform. EditModal has been updated, but other places (like RegisterTrees, SampleTreeCard etc. are pending)
2. MuiButton is used in 2 places - Project List and Project Details - please adapt this to the common theme (replace themeProperties with the Mui theme, and match styles to before). While doing this, make sure that we are not changing the way Mui buttons look elsewhere.

**Now**
Screenshot 2023-01-30 at 11 45 25 AM
**Before**
Screenshot 2023-01-30 at 11 44 29 AM

mohitb35 (approved):

PR #1645, opened 2022-12-20
feature: added the map key to the my forest section

sunilsabatp:
feature -> added map key to my forest section 
visit to` /profile` and should have either register trees or donated trees for that login account

### DIFF
--- a/src/features/user/Profile/components/MyTrees/MyTrees.tsx
+++ b/src/features/user/Profile/components/MyTrees/MyTrees.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement } from 'react';
+import React from 'react';
 import styles from '../../styles/MyTrees.module.scss';
 import dynamic from 'next/dynamic';
 import {
@@ -23,7 +23,7 @@ interface Props {
 }
 
 export default function MyTrees({ profile, authenticatedType, token }: Props) {
-  const { t, i18n, ready } = useTranslation(['country', 'me']);
+  const { t, ready } = useTranslation(['country', 'me']);
   const [contributions, setContributions] = React.useState();
   const { handleError } = React.useContext(ErrorHandlingContext);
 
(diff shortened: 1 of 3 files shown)


mariahosfeld:
Please change "Donated trees" to "Donated Trees". Other than that it looks good to me.

norbertschuler (approved):
LGFM, I pushed some commit to add a German translation, but as the branch name used uppercase characters `Feature/type_of_contributions` and my push used the standard lowercase name, GitHub could not correctly associate it with this PR.

src/features/user/Profile/components/MyTrees/MyTrees.tsx
```diff
@@ -85,6 +85,20 @@ export default function MyTrees({ profile, authenticatedType, token }: Props) {
...
+              <p className={styles.donatedTrees}> {t('me:donatedTrees')} </p>
+              <div className={styles.registeredTreesContainer}>
+                <div style={{ marginLeft: '2px' }}>
+                  <TreesIcon color="#3D67B1" />
+                </div>
+                <p className={styles.registeredTrees}>
+                  {t('me:registeredTrees')}
+                </p>
+              </div>
+            </div>
```
mohitb35, line 101: @sunilsabatp This can be structured a little better.

1. Logically, `donatedTreesContainer` and `registeredTreesContainer` are at the same level of hierarchy. But in this code, `registeredTreesContainer` is a child of `donatedTreesContainer`. 
    a. I'd suggest keeping a parent container element for both donatedTreesContainer and registeredTreesContainer. 
    b. You could keep a common structure for both donatedTreesContainer and registeredTreesContainer - each containing the icon and the label. 
2. Perhaps the naming could be better. For example, you could name the parent container element something like myTreesLegend.
3. Avoid inline styles like marginLeft: 2px. As you're using flexbox for the parent element, adjust spacing between donatedTreesContainer and registeredTreesContainer using the `gap` properties - you do not need to specify a px value.
4. Try to use common styles for donatedTreesContainer and registeredTreesContainer - the only thing that really differs is the colour.

mohitb35 (approved):
Looks good to merge.

PR #1646, opened 2022-12-21
Update Croatian currency to Euros

mohitb35:
Update Croatian currency to Euros. To be released on January 1, 2023 when Croatia joins the Eurozone and the Croatian Kuna (HRK) is deprecated in favor of the Euro.


norbertschuler (approved):
LGFM.

While testing this I switched my currency with the switcher in my profile and navigated back to the home page where the new currency is not shown in the footer. Probably still some bug.

Bildschirm­foto 2022-12-26 um 11 41 24

PR #1657, opened 2023-01-11
Add branch name checks using husky

mohitb35:
Introduces pre-commit hooks using husky to check that the branch name conforms to the following rules:

1. Permanent branches - `main` `develop`
2. Temporary branches start with
     a. `feature/` - most changes will use this.
     b. `hotfix/` - changes that need to be merged directly to main will use this.
3. Permitted characters (after /) - lowercase alphabets, numbers, hyphen/underscore as separator

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,5 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+# Validate the branch name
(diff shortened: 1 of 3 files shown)


norbertschuler (approved):
LGFM. Old PR's with other names probably should not be merging develop into it (once this is merged into develop) or have to be renamed then to be continued.

Issue #1600, opened 2022-11-02
Minor Code Refactors/Cleanup for Gift Funds


mohitb35:
These are a few minor refactors that could be done to the code for Gift Fund.

- [x] https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1597#discussion_r1011746989 - Separate the render logic in [src/features/user/GiftFunds/Details.tsx](https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1597/files/4f2fc5439c848c0751959893e8d7378047872175#diff-908982083223d15b152962dc54c401351477b6ab09c3bc54d5946bec6fa71bd7)
- [ ] https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1597#discussion_r1011744173 - Use a common component like `SingleColumnView` instead of an entirely new component `Details` in [src/features/user/GiftFunds/index.tsx](https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1597/files/4f2fc5439c848c0751959893e8d7378047872175#diff-3020e8d602785a7a4d5d76a542e9f306a4fd663a74186603a4ac1109b849e1d3)
- [ ] https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1597#discussion_r1011741966 - Used styled() to provide styling instead of modular scss

mariahosfeld:
closed, as #1601 has been merged into develop.

PR #1661, opened 2023-01-15
Gift Funds Cleanup

prachigarg19:
https://github.com/Plant-for-the-Planet-org/planet-webapp/issues/1600


src/features/user/GiftFunds/Details.tsx
```diff
@@ -1,41 +1,77 @@
...
+      gap: 15,
+
+      div: {
+        flex: 1,
+      },
+    },
+    hr: {
+      border: '1px solid #DDDBDA',
+    },
+  }));
```
mohitb35, line 46: A small change can be made here to ensure that the styles for AccountDetails (PlanetCash), BankAccountDetails (Manage Payouts) and StyledContainer are common. 

Compare the files - 
src/features/user/PlanetCash/components/AccountDetails.tsx
src/features/user/ManagePayouts/components/BankAccountDetails.tsx
src/features/user/GiftFunds/Details.tsx

src/features/user/GiftFunds/Details.tsx
```diff
@@ -1,41 +1,77 @@
...
 import { UserPropsContext } from '../../common/Layout/UserPropsContext';
-import styles from './GiftFunds.module.scss';
 import { useTranslation } from 'next-i18next';
+import { GiftFunds } from '../../common/types/user';
+import { styled } from '@mui/material';
 
 const Details = (): ReactElement | null => {
   const { user } = useContext(UserPropsContext);
   const { t, ready } = useTranslation('giftfunds');
+  const [validGiftFunds, setValidGiftFunds] = useState<GiftFunds[] | null>(
```
mohitb35, line 10: Would it make sense to move this logic and the mapping logic to the index file (src/features/user/GiftFunds/index.tsx)? That way, this component can be used to show a single GiftFund Details component, and be dependent only on props.

mohitb35 (changes requested):

src/features/user/GiftFunds/Details.tsx
prachigarg19: Since the component is not being used anywhere else yet, it won't bring much change logically. Though after the changes, the code would look cleaner and it'll bring more uniformity to the code as we've used props in `DashBoardView` children at other places, so I'll make the respective changes.

mohitb35 (approved):
Pushed some minor naming/prop interface changes. Looks good to merge.

PR #1662, opened 2023-01-22
hotfix:overlapping buttons of sample trees form

sunilsabatp:
visit to the following route ->` profile/treemapper/import`
then visit to 2nd form (sample trees form) and checkout the` continue` and `skip` button

### DIFF
--- a/src/features/user/TreeMapper/Import/components/SampleTrees.tsx
+++ b/src/features/user/TreeMapper/Import/components/SampleTrees.tsx
@@ -255,11 +255,7 @@ export default function SampleTrees({
       </div>
       <div className={`${styles.formField}`}>
         <div className={styles.formFieldHalf}>
-          <button
-            onClick={handleSubmit(onSubmit)}
-            className="primaryButton"
-            style={{ width: '188px' }}
-          >
+          <button onClick={handleSubmit(onSubmit)} className="primaryButton">
             {isUploadingData ? (
               <div className={styles.spinner}></div>
             ) : (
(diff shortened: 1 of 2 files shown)


src/features/user/TreeMapper/Import/components/SampleTrees.tsx
```diff
@@ -255,16 +255,24 @@ export default function SampleTrees({
       </div>
       <div className={`${styles.formField}`}>
         <div className={styles.formFieldHalf}>
-          <button onClick={handleSubmit(onSubmit)} className="primaryButton">
+          <button
+            onClick={handleSubmit(onSubmit)}
+            className="primaryButton"
+            style={{ width: '188px' }}
+          >
```
mohitb35, line 262: 1. Avoid using inline styles. It would be better to add styles in `Import.module.scss` to handle this styling. In this case, you can nest the button style inside the styles for `.formFieldHalf` along with `input` (which is already nested).
2. Please test your changes against different screen sizes (always!!). Currently, the buttons continue to overlap in mobile screens. I find it easier to style for mobile first, and then make adjustments for other screen sizes if needed.
3. Related to point 2 >>> It may be better to allow the button to take the full container width (100%) instead of specifying width = 188px. If you specify a pixel value for the width, you need to take care of other screen sizes as well by specifying a different pixel value. 

**Mobile: Width 100%**
Screenshot 2023-01-23 at 10 33 37 AM

**Mobile: Width 188px**
Screenshot 2023-01-23 at 10 41 17 AM

mohitb35 (changes requested):

mohitb35:
@sunilsabatp 

For the future (don't make changes now) 
> use "hotfix/.." branch names only when you're intending to merge directly to `main`. If merging to `develop`, please use "feature/.." branch names.

mohitb35 (approved):

mohitb35 (approved):
LGTM

PR #1663, opened 2023-01-25
Prevent husky install script running on production environments

mohitb35:
Updates prepare script to prevent `husky install` on production environments.


sarveshpro (approved):
LGTM ✅

PR #1670, opened 2023-01-30
fix: change the label

sunilsabatp:
to do-> change the label for` in-progress`  in` /profile/history` to `Pending`
image

### DIFF
--- a/src/features/user/Account/History.tsx
+++ b/src/features/user/Account/History.tsx
@@ -141,7 +141,7 @@ export default function History({
               <div className={styles.filterGrid}>
                 {accountingFilters &&
                   Object.entries(accountingFilters).map((item) => {
-                    return (
+                    return item[0] !== 'in-progress' ? (
                       <div
                         key={item[0]}
                         className={`${styles.filterButton} ${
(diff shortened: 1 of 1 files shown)


src/features/user/Account/History.tsx
```diff
@@ -151,6 +151,16 @@ export default function History({
...
                       </div>
+                    ) : (
+                      <div
+                        key={item[0]}
+                        className={`${styles.filterButton} ${
+                          filter === item[0] ? styles.selected : ''
+                        }`}
+                        onClick={() => handleSetFilter(item[0])}
+                      >
+                        {t('me:pending')}
```
norbertschuler, line 162: You seem to have duplicated the whole div-tag looking like the same code, is it? Why don't you just make a decision on what translation resource to use, e.g. `{t(item[0] !== 'in-progress' ? item[0] : 'me:pending')}`?

norbertschuler (changes requested):

src/features/user/Account/History.tsx
norbertschuler: Btw. this looks like quite a hack for the backend returning a wrong status, is it? We should add a comment about this for some future developer looking at it!

sunilsabatp: it is going to be fix in the backend , as in frontend we are mapping the response to display  button

sunilsabatp:
issue fixed in the backend

PR #1676, opened 2023-02-07
Add partner signup info for planet cash / bulk codes

mohitb35:
Updates main description (subtitle) for planet cash and bulk codes to include partner signup info.

### DIFF
--- a/public/static/locales/en/planetcash.json
+++ b/public/static/locales/en/planetcash.json
@@ -1,5 +1,6 @@
 {
   "title": "PlanetCash",
+  "partnerSignupInfo": "Our Planet Cash/Bulk Codes feature is a feature specifically for Plant-for-the-Planet partners. Feel free to contact us at <2>partner@plant-for-the-planet.org</2> if you are interested.",
   "description": "PlanetCash is a developer friendly tool that extends the Plant-for-the-Planet Platform with APIs and built in tools. PlanetCash balance is a contribution to Plant-for-the-Planet that can be dedicated by the donor to different projects on the platform.",
   "termsText": "View Terms & Conditions.",
   "learnMoreText": "Learn more about PlanetCash.",
--- a/public/static/locales/en/planetcash.json
+++ b/public/static/locales/en/planetcash.json
@@ -1,6 +1,6 @@
 {
   "title": "PlanetCash",
-  "partnerSignupInfo": "Our Planet Cash/Bulk Codes feature is a feature specifically for Plant-for-the-Planet partners. Feel free to contact us at <2>partner@plant-for-the-planet.org</2> if you are interested.",
+  "partnerSignupInfo": "Use of this feature by Companies is subject to partnership with Plant-for-the-Planet. Please contact <2>partner@plant-for-the-planet.org</2> for details.",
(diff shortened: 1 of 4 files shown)


public/static/locales/en/planetcash.json
```diff
@@ -1,5 +1,6 @@
 {
   "title": "PlanetCash",
+  "partnerSignupInfo": "Our Planet Cash/Bulk Codes feature is a feature specifically for Plant-for-the-Planet partners. Feel free to contact us at <2>partner@plant-for-the-planet.org</2> if you are interested.",
```
norbertschuler, line 3: I wonder if as the key and text is the same, the resource will only to have been translated once or twice in LingoHub. 🤔

mohitb35: I'm guessing twice? Since it's contained in different files..

norbertschuler: Then I would like to raise the question if it might be possible to only use one translation for this Trans component?

mohitb35: I'm trying to keep the text for different features in separate files as far as possible. Even though there may be some duplication, I feel it keeps the code clearer, and I do not have to import another translation file for a single string.

We do maintain some common strings, but this text is only going to be used in these 2 places - so I think it may be ok to duplicate it.

norbertschuler: Using translation memory in LingoHub will probably compensate for this then.

src/features/user/PlanetCash/index.tsx
```diff
@@ -146,26 +146,41 @@ export default function PlanetCash({
...
+              href="https://www.plant-for-the-planet.org/planetcash/"
+              target="_blank"
+              rel="noreferrer"
+            >
+              {t('learnMoreText')}
+            </a>
+            <br />
+            <a
+              className="planet-links"
+              href="https://www.plant-for-the-planet.org/terms-and-conditions/"
```
norbertschuler, line 176: Please check if it is ok to use this link and not the pp.eco links as e.g. https://pp.eco/legal/en/terms - also not sure if your link would show German content automatically, for the footer we dynamically insert a `locale` into the URL?

src/features/user/PlanetCash/index.tsx
```diff
@@ -146,26 +146,41 @@ export default function PlanetCash({
...
+                partner@plant-for-the-planet.org
+              </a>{' '}
+              if you are interested.
+            </Trans>
+          </p>
+          <p>
+            {t('description')}{' '}
+            <a
+              className="planet-links"
+              href="https://www.plant-for-the-planet.org/planetcash/"
```
norbertschuler, line 167: Same as the other link.

src/features/user/PlanetCash/index.tsx
mohitb35: Have updated the links here - https://github.com/Plant-for-the-Planet-org/planet-webapp/commit/3c72bbc74f3b889f8a41a7c37473a9a59c4daebc

src/features/user/PlanetCash/index.tsx
mohitb35: Have updated the links here - https://github.com/Plant-for-the-Planet-org/planet-webapp/commit/3c72bbc74f3b889f8a41a7c37473a9a59c4daebc

norbertschuler (approved):
LGFM, https://www.plant-for-the-planet.org/de/planetcash/ does not show German content yet, but that's now an issue for this PR.

PR #1680, opened 2023-02-15
fix: restricted double click of redeem code button

sunilsabatp:
Fixes # 
image

enter the code and double click redeem code button and make sure it only send 2 requests ,and try to checkout this behaviour in dev.pp.eco before testing


mariahosfeld:
Functionality works for me, only the first click triggers a request, the other ones are "ignored".

norbertschuler (approved):
LGFM

Issue #1649, opened 2022-12-29
In home page, the selected currency-country defaults to Germany - Euro when the page loads


mohitb35:
Ref: 

https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1646#pullrequestreview-1230021989

PR #1701, opened 2023-03-14
fix default country-currency

prachigarg19:
https://github.com/Plant-for-the-Planet-org/planet-webapp/issues/1649


norbertschuler (approved):
LGFM

mariahosfeld:
Functionality also looks good to me.

PR #1707, opened 2023-03-20
Prevent multiple clicks

prachigarg19:
Create a context to prevent multiple clicks/ multiple api requests on a button.


mariahosfeld:
@prachigarg19 , can you add the description to the PR?

norbertschuler (commented):
Code seems valid to me. I currently have no redeem voucher for testing.

sunilsabatp (approved):

sunilsabatp (commented):
resolve the merge conflicts and code factor  errors

mariahosfeld:
Functionality seems fine, tested redeeming a code.

mohitb35:
@prachigarg19 IMO, keeping the `disable` state in `UserPropsContext` isn't advisable because
1. `disable` does not have anything to do with a `User` - it indicates the state of a `button`
2. Keeping this in context can lead to conflicts when multiple components try to access or update `disable`. It seems to me like local state (or button component state - but that would need a change in structure in multiple places) would be a more suitable place to store this value.

mohitb35:
Closed as we've decided not to take this approach.

Ref - https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1707#issuecomment-1539370905

PR #1714, opened 2023-03-28
fix: resolves hidden banner text on small screens

mohitb35:
Fixes hidden banner text on salesforce home page (smaller screens)


sunilsabatp (approved):

mariahosfeld (commented):
Functionality seems fine.

PR #1727, opened 2023-04-04
Moves translations between namespaces

mohitb35:
Moves a few translations from `common.json` to `treemapperAnalytics.json` (for German)


mariahosfeld (approved):
Works now correctly.

PR #1706, opened 2023-03-17
Update subscriptions

prachigarg19:
1. Omit the option to edit the date, just leave Amount and Frequency
2. When the user selects a date for the cancellation, the Frontend Client needs to calculate how many times we will still deduct the money via the selected payment method from their account. This way we want to ensure that the donor has all necessary information right away and can keep track.

PR #1729, opened 2023-04-04
Update subscriptions

prachigarg19:
1. When the user selects a date for the cancellation, the Frontend Client needs to calculate how many times we will still deduct the money via the selected payment method from their account. This way we want to ensure that the donor has all necessary information right away and can keep track.
2. Omit the option to edit the date, just leave Amount and Frequency

Note: For reference, the old PR is #1706


public/static/locales/en/me.json
```diff
@@ -249,6 +249,9 @@
   "apiKeyMessage3": "Regenerating an API Key will invalidate your existing API Key immediately.",
   "regenerateKey": "Regenerate Key",
   "giftFund": "GiftFund",
+  "donatedTrees": "Donated Trees",
+  "remainingPaymentsText": "{{remainingPayments}} donations would take place before cancellation.",
+  "continue": "Continue"
```
sagararyal, line 254: @nirali25parekh comma is missing here.

prachigarg19: @sagararyal The deployment failed even after fixing the error.

mohitb35:
Adjust the position of the popup shown before confirming cancellation. It should be in the centre of the screen.

Screenshot 2023-04-28 at 11 05 19 AM
Screenshot 2023-04-28 at 11 05 15 AM

mohitb35:
This typo (seen while reactivating a subscription) should be corrected. It should say "billed" or "next billed", not "built". The date format may also need to be adjusted.

Screenshot 2023-04-28 at 11 05 59 AM

mohitb35:
After cancelling a subscription on a future date, this is the UI seen. While this is technically correct information, we should also continue to show the next payment date (if it exists). 

Screenshot 2023-04-28 at 11 03 34 AM

src/features/user/Account/EditModal.tsx
mohitb35, line 1: @prachigarg19 Please clean up the commented code that isn't needed any more.

src/features/common/Layout/CustomModal/index.tsx
```diff
@@ -6,10 +6,11 @@ import styles from './CustomModal.module.scss';
 interface Props {
   isOpen: boolean;
   handleContinue: () => void;
-  handleCancel: () => void;
+  handleCancel?: () => void;
   buttonTitle: string;
   modalTitle: string;
   modalSubtitle: string;
+  isCancel: boolean;
```
mohitb35, line 13: Is `isCancel` actually needed?

You can check whether `handleCancel` is undefined instead.

If you choose to keep this, rename `isCancel` to a more relevant name e.g. `isCancellable`, or `shouldAllowCancel` ?

src/features/user/Account/CancelModal.tsx
mohitb35, line 1: While cancelling the donation on a future date, I don't see any loading/processing state on the button like for the other actions (e.g. reactivating the subscription).

src/features/user/Account/Recurrency.tsx
```diff
@@ -134,6 +134,7 @@ export default function Recurrency({
                 handleCancelModalClose={handleCancelModalClose}
                 record={currentRecord}
                 fetchRecurrentDonations={fetchRecurrentDonations}
+                handleCancelModalOpen={() => setcancelModalOpen(true)}
```
mohitb35, line 137: I'm not clear on why this is needed. Let's discuss this.

mohitb35 (changes requested):

mariahosfeld:
Will be implemented differently.

PR #1733, opened 2023-04-10
Component unification (Api Key page)

prachigarg19:
Unify mui components in the webapp.


mariahosfeld:

mariahosfeld:
Please check, if the correct button style is used.

src/features/user/Settings/ApiKey/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+    <DashboardView
+      title={t('apiKey')}
+      subtitle={
+        <div>
+          <p>
+            {t('me:apiKeyMessage1')}
+            <br />
+            <br />
+            {t('me:apiKeyMessage3')}
+          </p>
```
mohitb35, line 21: Avoid using line breaks. Perhaps we can use `<p>` tags instead to separate the text like done in src/features/user/Widget/DonationLink/index.tsx

src/features/user/Settings/ApiKey/ApiKeyForm.tsx
```diff
@@ -76,19 +86,10 @@ export default function ApiKey({}: any) {
...
-          {t('me:apiKeyMessage1')}
-          {/* <br /><br />
-                   {t('me:apiKeyMessage2')} */}
-          <br />
-          <br />
-          {t('me:apiKeyMessage3')}
-        </p>
+    <StyledForm>
+      <div className="inputContainer">
         <div className={styles.apiKeyContainer}>
```
mohitb35, line 91: Would it be possible to replace the `<div className={styles.apiKeyContainer}>` with an `<InlineFormDisplayGroup>`? 

You may need to make some adjustments (e.g. to incorporate the hide/view key button - this could be useful https://mui.com/material-ui/react-text-field/#input-adornments)

The reason I'm suggesting this is for consistency - I can see some slight differences in the 2 places. One example is the copy button which has slightly different dimensions inspite of using a common component.

Remember to remove unused styles from ApiKey.module.scss if this is done.

src/features/user/Settings/ApiKey/ApiKeyForm.tsx
```diff
@@ -37,36 +46,37 @@ export default function ApiKey({}: any) {
   const getApiKey = async () => {
     setIsUploadingData(true);
     try {
-      const res = await getAuthenticatedRequest(
-        '/app/profile/apiKey',
-        token,
-        logoutUser
-      );
+      const res: ApiKeyResponse | ApiCustomError | undefined =
+        await getAuthenticatedRequest('/app/profile/apiKey', token, logoutUser);
```
mohitb35, line 50: While this works, `getAuthenticatedRequest` is a generic function and accepts a type.

https://www.typescriptlang.org/docs/handbook/2/generics.html

For an example, refer 

src/features/user/TreeMapper/Analytics/index.tsx Line 20 - 24

src/features/user/Settings/ApiKey/ApiKeyForm.tsx
```diff
@@ -37,36 +46,37 @@ export default function ApiKey({}: any) {
...
-      );
-      setIsUploadingData(false);
-      setApiKey(res.apiKey);
+      const res: ApiKeyResponse | ApiCustomError | undefined =
+        await putAuthenticatedRequest(
+          '/app/profile/apiKey',
+          undefined,
+          token,
+          logoutUser
+        );
```
mohitb35, line 72: While this works, `getAuthenticatedRequest` is a generic function and accepts a type.

https://www.typescriptlang.org/docs/handbook/2/generics.html

For an example, refer 

src/features/user/TreeMapper/Analytics/index.tsx Line 20 - 24

mohitb35 (changes requested):
@prachigarg19 please go through my feedback here.

src/features/user/Settings/ApiKey/ApiKeyForm.tsx
mohitb35: Made a minor change to pass `type=other` to `InlineFormDisplayGroup` as it does not contain only text fields. This vertically centers the children contained within `InlineFormDisplayGroup`.

mohitb35 (approved):

PR #1734, opened 2023-04-11
Adds space to columns of project info

mohitb35:
1. Adds space to columns of project info
2. Refactors some code to clean it up, removing unnecessary `<div>`s and other html tags, inline styles, using `flex` and `gap` instead of margin to make future code changes easier.

**Before**:
Screenshot 2023-04-11 at 1 00 22 PM

**After**:
Screenshot 2023-04-11 at 1 00 02 PM


mariahosfeld:
Looking good in both Firefox and Edge.

sunilsabatp (approved):

PR #1748, opened 2023-04-24
refactor: profile section redeem code modal

sunilsabatp:
code clean up for redeem code modal in profile section

### DIFF
--- a/src/features/common/RedeemMicro/RedeemCode.tsx
+++ b/src/features/common/RedeemMicro/RedeemCode.tsx
@@ -5,7 +5,7 @@ import CancelIcon from '../../../../public/assets/images/icons/CancelIcon';
 import styles from '../../../../src/features/user/Profile/styles/RedeemModal.module.scss';
 import { useTranslation } from 'next-i18next';
 import MaterialTextField from '../InputTypes/MaterialTextField';
-import CircularProgress from '@mui/material/CircularProgress';
+import Button from '@mui/material/Button';
 import { RedeemedCodeData } from '../types/redeem';
 import { SerializedError } from '@planet-sdk/common';
 
@@ -29,7 +29,7 @@ export interface SuccessfullyRedeemed {
   closeRedeem: () => void;
 }
 
-export const InputRedeemCode = ({
+export const EnterRedeemCode = ({
   setInputCode,
   inputCode,
   changeRouteCode,
@@ -72,15 +72,15 @@ export const InputRedeemCode = ({
       )}
 
       <div style={{ paddingTop: '30px' }}>
-        <button className={'primaryButton'} onClick={changeRouteCode}>
+        <Button variant="contained" onClick={changeRouteCode}>
           {t('redeem:redeemCode')}
-        </button>
+        </Button>
       </div>
     </div>
   );
 };
 
-export const RedeemCodeFailed = ({
+export const RedeemFailed = ({
   errorMessages = null,
   code,
   redeemAnotherCode,
@@ -99,9 +99,9 @@ export const RedeemCodeFailed = ({
       <div className={styles.RedeemTitle}>{code}</div>
 
       <div className={styles.redeemAnotherCodeDiv}>
-        <button className="primaryButton" onClick={redeemAnotherCode}>
+        <Button variant="contained" onClick={redeemAnotherCode}>
           {t('redeem:redeemAnotherCode')}
-        </button>
+        </Button>
       </div>
     </div>
   );
@@ -136,9 +136,9 @@ export const SuccessfullyRedeemed = ({
       </div>
 
       <div className={styles.redeemAnotherCodeDiv}>
-        <button className="primaryButton" onClick={redeemAnotherCode}>
+        <Button variant="contained" onClick={redeemAnotherCode}>
           {t('redeem:redeemAnotherCode')}
-        </button>
+        </Button>
       </div>
     </div>
   );
--- a/src/features/common/RedeemMicro/RedeemCode.tsx
+++ b/src/features/common/RedeemMicro/RedeemCode.tsx
@@ -99,7 +99,11 @@ export const RedeemFailed = ({
       <div className={styles.RedeemTitle}>{code}</div>
 
       <div className={styles.redeemAnotherCodeDiv}>
-        <Button variant="contained" onClick={redeemAnotherCode}>
+        <Button
+          variant="contained"
+          onClick={redeemAnotherCode}
+          sx={{ textTransform: 'none' }}
+        >
           {t('redeem:redeemAnotherCode')}
         </Button>
       </div>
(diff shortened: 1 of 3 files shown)


mariahosfeld:
Generally: Functionality looks good.
But: After changing the button, the capitalization has changed (at least for German), this needs to be reverted to the old capitalization.
Old Button Style (correct version):
grafik
New Button Style:
grafik

sunilsabatp:
> @sunilsabatp, still facing the same capitalization issue as before. Checked on other branch, there it is working correctly

sunilsabatp:
> @sunilsabatp, still facing the same capitalization issue as before. Checked on other branch, there it is working correctly.

this is the dev branch sample 
image

this is the sample for this branch 
image

mariahosfeld:
The issue for this particular button is resolved, but as there more, we should find a more centralized solution.

src/features/common/RedeemMicro/RedeemCode.tsx
mohitb35, line 1: I suggest separating the components into 3 different files. You can use an index file to export them together so that imports do not need to be updated elsewhere.

src/features/common/RedeemMicro/RedeemCode.tsx
```diff
@@ -64,23 +64,21 @@ export const InputRedeemCode = ({
...
-        </button>
+        </Button>
       </div>
     </div>
   );
 };
 
-export const RedeemCodeFailed = ({
+export const RedeemFailed = ({
   errorMessages = null,
```
mohitb35, line 82: Remove `errorMessages` if it is not going to be used.

src/features/common/RedeemMicro/RedeemCode.tsx
```diff
@@ -64,23 +64,21 @@ export const InputRedeemCode = ({
...
-      {errors.code && (
-        <span className={styles.formErrors}>{errors.code.message}</span>
-      )}
-
       <div style={{ paddingTop: '30px' }}>
-        <button className={'primaryButton'} onClick={changeRouteCode}>
+        <Button variant="contained" onClick={handleSubmit(changeRouteCode)}>
           {t('redeem:redeemCode')}
-        </button>
+        </Button>
```
mohitb35, line 75: The padding above the button looks a little too much.

src/features/common/RedeemMicro/RedeemCode.tsx
mohitb35, line 1: Can inline styles be removed while we're refactoring this?

src/features/common/RedeemMicro/RedeemCode.tsx
```diff
@@ -64,23 +64,21 @@ export const InputRedeemCode = ({
           placeholder="XAD-1SA-5F1-A"
           label=""
           variant="outlined"
+          error={errors.code}
+          helperText={errors.code && errors.code.message}
         />
```
mohitb35, line 69: Should we move to the field styles similar to Edit Profile? i.e. use TextField instead of MaterialTextField? We're moving away from MaterialTextField

mohitb35 (changes requested):
Hey Sunil, please see my comments below.

mariahosfeld:
Buttons show up correctly now.

src/features/common/RedeemMicro/RedeemCode.tsx
mariahosfeld: We decided that in this case we are going to use the message in both places, please take this into consideration.

src/features/user/Profile/styles/RedeemModal.module.scss
```diff
@@ -84,6 +84,12 @@
   min-height: 275px;
   min-width: 438px;
   padding: 20px 56px;
+  >div:nth-child(5) {
+    padding-top: 15px;
+  }
```
mohitb35, line 89: What is this style for? If this is for the padding above the button, it will cause issues - what happens if I change the order of the div elements, or I add another div element above the button.

Please change this.

src/features/common/RedeemCode/EnterRedeemCode.tsx
```diff
@@ -0,0 +1,65 @@
...
+  setInputCode: React.Dispatch<React.SetStateAction<string | null>>;
+  inputCode: string | null;
+  changeRouteCode: () => void;
+  closeRedeem: () => void;
+}
+
+const EnterRedeemCode = ({
+  setInputCode,
+  inputCode,
+  changeRouteCode,
```
mohitb35, line 19: `changeRouteCode` >> what does this do? Please give this a more understandable name.

src/features/common/RedeemCode/EnterRedeemCode.tsx
```diff
@@ -0,0 +1,65 @@
...
+          variant="outlined"
+          error={errors.code}
+          helperText={errors.code && errors.code.message}
+        />
+      </div>
+
+      <div>
+        <Button variant="contained" onClick={handleSubmit(changeRouteCode)}>
+          {t('redeem:redeemCode')}
+        </Button>
```
mohitb35, line 59: There should be some loading state

sunilsabatp: to implement the loading state there is need to create state variable for it. So once this [https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1707/commits/78399f602c1b51f93cbe1a28799af135eb70f771](PR) get merge into develop I can get a state variable (`disable`) which is there in userPropContext

mohitb35: Why would you use a variable from userPropContext for disabling the action? 

It seems more suitable to use a local state variable, which gets before and after the redeem code api call. Using context indiscriminately makes the code tougher to read, and can also lead to conflicts (for example, if one component updates the value of `disable` while it's in use somewhere else).

Use context when we need to store data that is used for multiple components, and pass it through multiple levels of a component tree.

src/features/user/Profile/styles/RedeemModal.module.scss
mohitb35: @sunilsabatp This is done. One minor change -> I noticed that

.redeemAnotherCodeDiv and .redeemCodeButtonContainer have different styles. Should these be the same (for consistency in spacing)? Also, should one class satisfy the purpose?

mohitb35: Styles applied were not reflected due to nesting / application of incorrect class. Corrected here - cc264bd68674b084d92c95552b95f7af6adda7a4

src/features/common/RedeemCode/EnterRedeemCode.tsx
mohitb35: Minor adjustments made here -  032160ef1253396972a4677a31d46fc5a58aaa71

mohitb35 (approved):
Made minor adjustments - LGFM now.

mariahosfeld:
Functionality works.

PR #1753, opened 2023-04-28
Updates webpack from 5.74.0 to 5.76.0

mohitb35:
Updates webpack in package.json to ^5.76.0 i.e. min version 5.76.0

Webpack in package-lock.json is already at 5.80.0, so nothing needed to be done there.


Shreyaschorge (approved):
LGTM

PR #1769, opened 2023-05-09
Adds enable_intro query param to enable intro video for embedded site

mohitb35:
- updates QueryParamsContext
1. initializes query param related state at one go
2. adds `enableIntro` 
3. adds `isContextLoaded`  to detect context initialization

- if embed=true and enable_intro=false is passed in url
1. does not show video
2. hides play button

Testing scenarios
- [ ] No `embed` param / `embed=true` is not passed >> video shown on first load, play button visible on map and clickable
- [ ] `embed=true` is passed, `enable_intro=true` is not passed (empty or any other value) >> video and play button not shown
- [ ] `embed=true` and `enable_intro=true` is passed >> video shown on first load, play button visible on map and clickable

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -202,43 +202,43 @@ const PlanetWeb = ({ Component, pageProps }: any) => {
   } else {
     return (
       <ErrorHandlingProvider>
-        <div>
-          <div
-            style={
-              showVideo &&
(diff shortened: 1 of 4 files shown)


mariahosfeld (approved):
Functionality works correctly.

sunilsabatp (approved):

PR #1773, opened 2023-05-12
Sign up button multiple clicks

prachigarg19:
Prevent multiple clicks on signup button

### DIFF
--- a/src/features/user/CompleteSignup/index.tsx
+++ b/src/features/user/CompleteSignup/index.tsx
@@ -43,7 +43,7 @@ export default function CompleteSignup(): ReactElement | null {
   const { i18n, t, ready } = useTranslation(['editProfile', 'donate']);
   const { setErrors, redirect } = React.useContext(ErrorHandlingContext);
   const [addressSugggestions, setaddressSugggestions] = React.useState([]);
-  const [disable, setDisable] = React.useState<boolean>(false);
+  const [isProcessing, setIsProcessing] = React.useState<boolean>(false);
   const geocoder = new GeocoderArcGIS(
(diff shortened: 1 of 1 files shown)


src/features/user/CompleteSignup/index.tsx
```diff
@@ -43,6 +43,7 @@ export default function CompleteSignup(): ReactElement | null {
   const { i18n, t, ready } = useTranslation(['editProfile', 'donate']);
   const { setErrors, redirect } = React.useContext(ErrorHandlingContext);
   const [addressSugggestions, setaddressSugggestions] = React.useState([]);
+  const [disable, setDisable] = React.useState<boolean>(false);
```
mohitb35, line 46: Could you rename `disable` to `isProcessing` to indicate it's purpose better? It indicates that the request is processing, that's why the button is disabled. Update this throughout.

P.S. boolean values should usually be named beginning with `is`, `can`, `should`, `has`. Think of it like asking a question.

src/features/user/CompleteSignup/index.tsx
```diff
@@ -165,6 +166,7 @@ export default function CompleteSignup(): ReactElement | null {
 
   const sendRequest = async (bodyToSend: any) => {
     setRequestSent(true);
+    setDisable(true);
```
mohitb35, line 169: You're setting `disable` to `true`. In case of success, the user will be redirected away from this page, so there isn't any need to set it as  `false`. 

The failure scenario is not handled, however. The button will remain disabled if the request fails, preventing the user from attempting the request again (which is desirable).

mohitb35 (approved):

PR #1786, opened 2023-05-22
Updates node version and package-lock after dep updates 

mohitb35:
1. Updates node version to 18.x as 16.x is reaching end of life in Sep 2023, and being deprecated in Aug 2023.
2. Also updates npm version to 9.x as that is being used on vercel.
3. Syncs package lock to the dependency updates #1782 #1783 #1784


Shreyaschorge (approved):
LGTM

PR #1801, opened 2023-06-13
Update react-hook-form to v7

mohitb35:
Changes in this pull request:
- updates react-hook-form to 7.44
- Wrap TextField/MaterialTextField/other MUI inputs in Controllers, and update existing Controller syntax where applicable. Use of `register` has been removed.
- code cleanup to address related TS warnings
- set up form data types and set default values in most places.
- take form `errors` from `formState`


prachigarg19:
This comment is not related to the changes in PR, but while testing impersonate-user out, I saw the button has no spinner to denote the request is being processed. Along with this, it allowed sending multiple api requests.
image

src/features/user/TreeMapper/Import/components/SpeciesAutoComplete.tsx
```diff
@@ -14,14 +14,27 @@ import { ErrorHandlingContext } from '../../../../common/Layout/ErrorHandlingCon
...
+}
+
+export default function SpeciesSelect<
+  TFieldValues extends FieldValues,
+  TName extends FieldPath<TFieldValues>
+>({
+  label,
+  name,
+  width,
+  // mySpecies,
```
sunilsabatp, line 35: remove the commented part , If it not require

mohitb35: Leaving the comments in case this is needed for future reference. Will remove later if it is clear that this will not be needed.

sunilsabatp (approved):

PR #1808, opened 2023-06-20
Remove @mui/styles (incompatible with React 18)

mohitb35:
Removes @mui/styles throughout, and updates styles to maintain status quo.

- replaces `ServerStyleSheets` with `EmotionCache` for SSR support
- Updates dependencies for `emotion`, and adds `@emotion/server` as a dependency
- removes instances of `withStyles` / `makeStyles` and replaces with `styled()` or modular styles if needed
- removes `AutoCompleteCountry` (unused component) and renames `AutoCompleteCountryNew` to `AutoCompleteCountry`


mariahosfeld (commented):
Did functionality testing, so far it looks good to me.

src/theme/themeStyles.ts
```diff
@@ -99,6 +99,13 @@ const materialTheme = createTheme({
         },
       },
     },
+    MuiAutocomplete: {
+      styleOverrides: {
+        option: {
+          fontSize: '0.875rem',
```
Shreyaschorge, line 105: The sizes specified in rem (root em) units should not change for different browsers. The rem unit is relative to the root element's font size, which is typically set to a base value of 16 pixels (or another value defined in the CSS). As long as the root element's font size remains consistent across browsers, the rem units will yield the same results.

Different browsers may have slightly different default font sizes or zoom levels, which could indirectly affect the rendering of rem units. To ensure consistency, it's recommended to set a base font size explicitly.

Currently css in web-app
html {
  font-size: 100%
}

body {
   font-size: 0.875rem;
}

So there is no base font size set currently.

Shreyaschorge (changes requested):
LGTM!
Jst one thing.

Shreyaschorge (approved):

src/theme/themeStyles.ts
mohitb35: I've created an issue #1809  to make font size and related properties consistent across the app.

PR #1811, opened 2023-07-07
ProjectPropsContext Refactor

mohitb35:
- sets up `ProjectPropsContextInterface` and associated types, including `PlantLocation`
- updates `ProjectPropsContext` types
- removes unrequired default value setup in ProjectPropsContext
- creates `useProjectProps` hook to access context (instead of directly)
- replaces import of `ProjectPropsContext` with `useProjectProps` hook throughout project
- refactors _app.tsx (as ProjectPropsContext is not accessible)
- sets up `MapHolder` component to render `MapLayout` and `PlayButton`
- minor update to eslint rules to allow use before define for types/interfaces

PR #1815, opened 2023-07-17
Feature/use planet sdk project types

mohitb35:
1. Updates the webapp code to use project types provided in planet-sdk
2. Resolves TS warnings related to projects across the codebase
3. Resolves most TS warnings/errors in src/features/projects
4. removes certain types to avoid duplication, and reuses common types
5. adds stronger type checks
6. Removes unused code, variables, props etc.

Note: this builds upon #1811, and should be merged to develop only after #1811 is also merged.


Shreyaschorge (approved):
LGTM!

I was just wondering why do we need the an extra `/build` folder while importing types from planet-sdk?

mohitb35:
> LGTM!
> 
> I was just wondering why do we need the an extra `/build` folder while importing types from planet-sdk?

I'm just trying to import from the lowest possible level to keep bundle size low. I'm not sure if it makes a difference though.

PR #1820, opened 2023-07-21
Fixes filter toggles on map

mohitb35:


mariahosfeld (commented):
Functionality looks good.

norbertschuler:
Any idea why you can only switch between top projects and all projects when the first filter "Large Scale Restoration" is selected?

Bildschirmfoto 2023-07-22 um 10 11 45

Here is no switching possibility and when you have been selected top projects before you might end up seeing no projects in the list, but still some projects on the map:

Bildschirmfoto 2023-07-22 um 10 14 39

norbertschuler (approved):
Tested locally, works for me.

mohitb35:
> Any idea why you can only switch between top projects and all projects when the first filter "Large Scale Restoration" is selected?
> 
> Bildschirmfoto 2023-07-22 um 10 11 45
> Below is no switching possibility and when you have been selected top projects before you might end up seeing no projects in the list, but still some projects on the map:
> 
> Bildschirmfoto 2023-07-22 um 10 14 39
> On https://trees.salesforce.com/ the switching possibility between top projects and all projects is not available at all.

Nothing to be done here, as the top projects list is not shown on trees.salesforce.com (which is the only tenant which has filters)

Shreyaschorge (approved):
LGTM!

PR #1822, opened 2023-07-24
Resolve cookiePolicy and redeemPolicy popup

prachigarg19:
When we try to run localhost:3000, the cookiePolicy and redeemPopup are not visible.

**Reason:**
The useEffect responsible for setting values for the above runs twice in development mode. 

This PR deals with preventing the useEffect from re-executing code.


sunilsabatp (commented):
could you resolve the  conflict ?

Shreyaschorge (commented):
I dont see the cookiePolicy popup in my local machine after running `npm run dev`

image

src/features/common/Layout/CookiePolicy/index.tsx
```diff
@@ -15,19 +15,31 @@ export default function CookiePolicy() {
...
+      if (!prev) {
+        setShowCookieNotice(true);
+      } else {
+        setShowCookieNotice(prev === 'true');
+      }
     }
   }, []);
 
   useEffect(() => {
     localStorage.setItem('cookieNotice', showCookieNotice);
```
norbertschuler, line 35: I think the problem is this line executes with `showCookieNotice` as `false` even after `setShowCookieNotice(true);` in called the useEffect above, so the localStorage value and the React state differ.

Another solution not waiting for any mounts could be the attached code setting state and localStorage at the same time:

```TypeScript
import React, { useEffect, useState } from 'react';
import CloseIcon from '../../../../../public/assets/images/icons/CloseIcon';
import styles from './CookiePolicy.module.scss';
import { useUserProps } from '../UserPropsContext';
import { useTranslation } from 'next-i18next';

export default function CookiePolicy() {
  const [showCookieNotice, setShowCookieNotice] = useState(false);
  const { t, ready } = useTranslation(['leaderboard']);
  const { user, contextLoaded } = useUserProps();

  const setCookieNoticeLocalStorage = (show: boolean) => {
    setShowCookieNotice(show);
    localStorage.setItem('cookieNotice', show ? 'true' : 'false');
  };

  useEffect(() => {
    if (contextLoaded && user) {
      setCookieNoticeLocalStorage(false)
    }
  }, [contextLoaded, user]);

  useEffect(() => {
    const prev = localStorage.getItem('cookieNotice');
    if (!prev) {
      setCookieNoticeLocalStorage(true)
    } else {
      setCookieNoticeLocalStorage(prev === 'true')
    }
  }, []);

  return ready && showCookieNotice ? (
    <div className={styles.cookieContainer}>
      <button
        id={'cookieCloseButton'}
        className={styles.closeButton}
        onClick={() => setCookieNoticeLocalStorage(false)}
      >
        <CloseIcon color={styles.primaryColor} />
      </button>
      <div className={styles.cookieContent}>
        {t('common:privacyPolicyNotice')}{' '}
        <a href="https://www.plant-for-the-planet.org/en/footermenu/privacy-policy">
          {t('common:privacyPolicy')}
        </a>
      </div>
    </div>
  ) : null;
}
```

Shreyaschorge:
> I dont see the cookiePolicy popup in my local machine after running `npm run dev`
> 
> image

I tried this on the browser window where local-storage already had a value.
Everything works as expected in a fresh incognito window

Shreyaschorge (approved):
LGTM!

Shreyaschorge (approved):
LGTM!

PR #1826, opened 2023-07-25
Fix widgets layout

prachigarg19:
Current look: 
image

This PR adds margin to the top of widgets container.

### DIFF
--- a/src/features/common/Layout/UserLayout/UserLayout.module.scss
+++ b/src/features/common/Layout/UserLayout/UserLayout.module.scss
@@ -40,6 +40,15 @@
 .profilePageWrapper {
   width: 100%;
   min-height: 100vh;
+  .widgetsContainer {
+    padding: 0px; 
+    margin-top: 130px;
+  }
+    @include xsPhoneView {    
+      & .widgetsContainer {
+          padding-top: 80px !important;
+        }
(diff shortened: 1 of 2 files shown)


mohitb35 (commented):
@prachigarg19 This works, but it may be better to just keep a separate class for this div instead of using `profilePage` - as we would aim to remove `profilePage` throughout.

src/features/common/Layout/UserLayout/UserLayout.module.scss
```diff
@@ -40,6 +40,15 @@
 .profilePageWrapper {
   width: 100%;
   min-height: 100vh;
+  .widgetsContainer {
+    padding: 0px; 
+    margin-top: 130px;
+  }
+    @include xsPhoneView {    
+      & .widgetsContainer {
+          padding-top: 80px !important;
```
norbertschuler, line 49: Not sure if this is necessary testing with Chrome some mobile device views:

www1.plant-for-the-planet.org:

Bildschirmfoto 2023-07-27 um 09 18 13


After your change having some additional white space:

Bildschirmfoto 2023-07-27 um 09 17 51

norbertschuler (approved):
LGFM

PR #1831, opened 2023-08-01
Component Unification (Import Data)

prachigarg19:


mariahosfeld:
If possible, these two buttons should also have the new layout
image

mariahosfeld:
Functionality and layout are good, just the change above mentioned should happen to unify everything (if possible).

src/features/user/TreeMapper/Import/components/SampleTreeCard.tsx
```diff
@@ -294,7 +281,7 @@ export default function SampleTreeCard({
                     }
                   }
                 )}
-              </MaterialTextField>
+              </TextField>
             )}
           />
           {/* <SpeciesSelect label={t('treemapper:species')} name={`scientificSpecies`} width='300px' control={control} /> */}
```
sunilsabatp, line 287: could you please remove the commented part  ?

sunilsabatp (approved):

PR #1841, opened 2023-08-20
Hotfix/remove impersonation data

sunilsabatp:
Fix -> cleaned up local storage of Impersonation data if impersonated  account get deleted


mariahosfeld:
Functionality works as expected.

PR #1838, opened 2023-08-07
Release August II

mariahosfeld:
**Includes:**

- #1826 
- #1827 
- #1835 
- #1816 
- #1842

PR #1842, opened 2023-08-21
Cleanup for #1838

mohitb35:
Changes
- removes unused types in `project.d.ts`
- deletes non-required `ConfigType` (both definition and use)
- removes types for array method callback args, if not needed, in a. `Navbar/index.tsx` b. `UserLayout.tsx`
- resolves `any` warning in `UserLayout.tsx`


Shreyaschorge (approved):
LGTM!

PR #1840, opened 2023-08-19
use some existing label for projects on profile page

norbertschuler:
Fixes  a bug

Changes in this pull request:
- renamed and used existing unused label for `Projects` on profile page to allow a proper translation

PR #1833, opened 2023-08-03
Displays new project properties (map/project details)

mohitb35:
**Project Details**
1. Renames "First Tree Planted" to "Restoration Started", "Planting Seasons" to "Restoration Seasons"
2. Displays Main Interventions (if data exists)
3. Also makes changes to allow restoration/activity seasons to occupy full column width.
4. Shows ecosystem for the project
5. Shows per unit cost based on `unitType` instead of project `purpose`
6. Shows unit type for total units donated (instead of "Trees" by default).

**Map (List and Popup on hover)**
1. Shows ecosystem for the project
2. Shows per unit cost based on `unitType` instead of project `purpose`
3. Shows unit type for total units donated (instead of "Trees" by default).

**Create/Edit Project (Detailed Analysis)**
1. Changes label from "First Tree Planted" to "Restoration Started"

Screenshot 2023-08-03 at 10 54 05 AM
Screenshot 2023-08-03 at 10 52 40 AM
Screenshot 2023-08-03 at 10 52 34 AM
Screenshot 2023-08-03 at 10 51 13 AM
Screenshot 2023-08-03 at 10 54 25 AM
Screenshot 2023-08-03 at 10 54 20 AM

Pending
- [ ] Showing total donations / units > API change may be needed, and corresponding front end changes as well

PR #1704, opened 2023-03-15
refactor: project creation

sunilsabatp:
refactor ->1. ordering the components of project creation and following the same stylings similar to` e.g planetCash`
                  2. define types 
                  3. changed the layout of forms

PR #1843, opened 2023-08-22
Release All-Methods Projects

mariahosfeld:
**Includes:**

- #1704 
- #1840 
- #1833 
- #1839


public/static/locales/en/donate.json
```diff
@@ -5,7 +5,7 @@
   "yourTreesPlantedByOnLocation_plural": "Your {{treeCount}} trees will be planted by {{projectName}} in {{location}}.",
   "myTreesPlantedByOnLocation": "My {{treeCount}} tree is being planted in {{location}}.",
   "myTreesPlantedByOnLocation_plural": "My {{treeCount}} trees are being planted in {{location}}.",
-  "PROJECTS": "PROJECTS",
```
prachigarg19, line 8: similar changes for `locales/cs/donate.json` has not been made. Why is that?

mariahosfeld: Only English is handled via the code here, the rest is via LingoHub. If something in other languages is missing, or we need to make this change as well, we can do that in a subsequent release.

prachigarg19 (approved):

PR #1863, opened 2023-09-14
Limits projects renderable in public profile

mohitb35:
Prevents rendering of non-tree/conservation projects, and the resultant crash due to absence of `metadata`


norbertschuler (approved):
LGFM

mariahosfeld:

PR #1867, opened 2023-09-18
Handles error due to invalid raster data

mohitb35:
This resolves the error scenario where `https://raster.plant-for-the-planet.org/api/v1/project/` returned `{"message":"Something went wrong"}`  resulting in an app crash while viewing any project.

This has been resolved so that the app does not crash any more by checking for validity of the response (it should contain an `imagery` object). If the response is invalid, it is treated as an error condition, and the tabs to select different map modes will not be seen in this case.

Note: raster URL for this branch has been configured to always result in an error scenario

### DIFF
--- a/src/features/projects/components/maps/Project.tsx
+++ b/src/features/projects/components/maps/Project.tsx
@@ -56,7 +56,10 @@ export default function Project({
     let result2;
 
     try {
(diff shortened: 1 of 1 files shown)


norbertschuler (approved):
Works for me

PR #1856, opened 2023-08-31
feat:types for map

sunilsabatp:
fix -> types error for maps

PR #1871, opened 2023-09-21
Resolve TS errors: ManageProjects

prachigarg19:
This PR resolves ts errors for all files in` ManageProjects`

### DIFF
--- a/src/features/common/types/project.d.ts
+++ b/src/features/common/types/project.d.ts
@@ -109,25 +109,25 @@ export interface Project {
   revisionPeriodicityLevel: any;
 }
 
-export interface CertificateScopeProjects extends ProfileProject {
+export type CertificateScopeProjects = ProfileProject & {
   _scope: string;
   certificates: Certificate[];
-}
+};
 
-export interface ImagesScopeProjects extends ProfileProject {
+export type ImagesScopeProjects = ProfileProject & {
   _scope: string;
   images: UploadImage[];
-}
+};
 
-export interface ExpensesScopeProjects extends ProfileProject {
+export type ExpensesScopeProjects = ProfileProject & {
   _scope: string;
   expenses: ProjectExpense[];
-}
+};
 
-export interface SitesScopeProjects extends ProfileProject {
+export type SitesScopeProjects = ProfileProject & {
   _scope: string;
   sites: Site[];
-}
+};
 interface Units {
   tree?: number;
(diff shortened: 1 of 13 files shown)


src/features/user/ManageProjects/components/ProjectMedia.tsx
mohitb35, line 1: While fetching images for the project, you're using the `Project` type. I think this should be handled similar to `ProjectCertificates` - where you've defined a `CertificateScopeProjects` interface.

```
const fetchImages = async () => {
    try {
      // Fetch images of the project
      if (projectGUID && token) {
        const result = await getAuthenticatedRequest<Project>(
```

src/features/common/types/project.d.ts
mohitb35, line 1: You've made a change here to `ProjectSitesProps`, but ProjectSites has not been updated (handle other issues over there as well)

src/features/user/ManageProjects/components/ProjectSpending.tsx
mohitb35, line 1: While fetching expenses for the project, you're using the `Project` type. I think this should be handled similar to `ProjectCertificates` - where you've defined a `CertificateScopeProjects` interface.

src/features/user/ManageProjects/index.tsx
```diff
@@ -358,7 +334,11 @@ export default function ManageProjects({
 
   return (
     <DashboardView
-      title={projectGUID ? project?.name : t('manageProjects:addNewProject')}
+      title={
+        project && projectGUID
+          ? project?.name
```
mohitb35, line 339: The ? check in `project?.name` is redundant

mohitb35 (changes requested):
@prachigarg19 Some minor changes suggested, but there are 2 components which seem to have been left alone: `MapComponent` and `ProjectSites`. Could you address the TS warnings there too?

prachigarg19:
> @prachigarg19 Some minor changes suggested, but there are 2 components which seem to have been left alone: `MapComponent` and `ProjectSites`. Could you address the TS warnings there too, if they are not being addressed in #1856 ?

@sunil-sabat will be working on `MapComponent`
Error: `Unexpected any. Specify a different type.` was not resolved since currently I'm facing `500` error when I try to save a site.

mohitb35 (approved):
LGTM

PR #1874, opened 2023-09-28
Component unification (Treemapper>Plant locations)

prachigarg19:
This pr adds styles to make styles for `/profile/treemapper` consistent with remaining webapp

### DIFF
--- a/src/features/user/TreeMapper/TreeMapper.module.scss
+++ b/src/features/user/TreeMapper/TreeMapper.module.scss
@@ -434,11 +434,12 @@
   display: flex;
   justify-content: space-between;
   align-items: center;
-  div {
+  .treeMapperTitle {
     font-size: 2.25rem;
     font-weight: 700;
     width: 100%;
     margin-bottom: 10px;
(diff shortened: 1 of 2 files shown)


src/features/user/TreeMapper/TreeMapper.module.scss
```diff
@@ -50,6 +50,7 @@
   height: 85%;
   height: 100%;
   background-color: $backgroundColor;
+  margin-top: 20px;
```
mohitb35, line 53: This class does not seem to be used anywhere?

src/features/user/TreeMapper/TreeMapper.module.scss
```diff
@@ -501,6 +509,7 @@
     border-top-left-radius: 0px;
     border-top-right-radius: 0px;
     box-shadow: none;
+    padding-top: 65px;
     > :first-child {
```
mohitb35, line 513: The padding to the sides of the title is 50px on this screen, but it's 40 px elsewhere. Please correct this (check for different screen sizes)

PR #1876, opened 2023-10-02
Resolves null country while signing up

mohitb35:
This fixes the issue of the user getting stuck while signing up, due to the country being set as null


Shreyaschorge (approved):
LGTM!

Issue #1802, opened 2023-06-15
User Impersonation - multiple requests possible


mohitb35:
This comment is not related to the changes in PR, but while testing impersonate-user out, I saw the button has no spinner to denote the request is being processed. Along with this, it allowed sending multiple api requests.
<img width="378" alt="image" src="https://github.com/Plant-for-the-Planet-org/planet-webapp/assets/72646230/33a05854-1d1c-48d4-bb77-a033a4fa5449">

_Originally posted by @prachigarg19 in https://github.com/Plant-for-the-Planet-org/planet-webapp/issues/1801#issuecomment-1591716847_

PR #1878, opened 2023-10-03
User Impersonation - multiple requests possible fix

prachigarg19:
https://github.com/Plant-for-the-Planet-org/planet-webapp/issues/1802


src/features/user/Settings/ImpersonateUser/ImpersonateUser.module.scss
```diff
@@ -32,3 +32,31 @@
...
+  border: 2px solid #f3f3f3;
+  border-radius: 50%;
+  border-top: 2px solid transparent;
+  width: 16px !important;
+  height: 16px !important;
+  -webkit-animation: spin 1s linear infinite; /* Safari */
+  animation: spin 1s linear infinite;
+}
+
+@-webkit-keyframes spin {
```
sunilsabatp, line 46: since `-webkit-` for WebKit-based browsers (Safari and older versions of Chrome). will it work in Mozilla and other browsers ?

src/features/user/Settings/ImpersonateUser/ImpersonateUserForm.tsx
```diff
@@ -16,6 +17,7 @@ const ImpersonateUserForm = (): ReactElement => {
   const router = useRouter();
   const { t } = useTranslation('me');
   const [isInvalidEmail, setIsInvalidEmail] = useState(false);
+  const [isProcessing, setIsProcessing] = useState<boolean>(false);
```
sunilsabatp, line 20: there is no need to define type for Boolean. typescript can implicitly detect this

sunilsabatp (changes requested):

src/features/user/Settings/ImpersonateUser/ImpersonateUser.module.scss
prachigarg19: Yes. Also, this is the same code as written everywhere else for spinner class. I saw that the code for spinner has been repeated at multiple places and probably should be kept common in the future.

src/features/user/Settings/ImpersonateUser/ImpersonateUser.module.scss
```diff
@@ -32,3 +32,31 @@
   max-width: 208px;
   text-align: center;
 }
+// Loader for Buttons
+.spinner {
+  border: 2px solid #f3f3f3;
+  border-radius: 50%;
+  border-top: 2px solid transparent;
+  width: 16px !important;
+  height: 16px !important;
```
mohitb35, line 41: Remove `!important` for the width and height - it isn't needed here. Avoid using `!important` except in rare cases where nothing else works,the solution is very complicated (e.g. if you need to combine a large number of classes together to get enough specificity for a style to work), or a temporary urgent fix.

mohitb35 (changes requested):
Functionally this is fine. I've left some minor feedback, LGTM once that's addressed

mohitb35 (approved):
LGTM

PR #1881, opened 2023-10-04
Update email validation

prachigarg19:
Implement zod email validation in place of current validation to obtain a common approach throughout the webapp

### DIFF
--- a/src/utils/isEmailValid.ts
+++ b/src/utils/isEmailValid.ts
@@ -1,22 +1,11 @@
-const INVALID_DOMAIN_LIST = ['dummy.de', 'example.com', 'company.de'];
+import { z } from 'zod';
 
-const generateInvalidEmailRegex = (): string => {
-  const startingPattern =
-    '^[a-zA-Z0-9_.+-]+@(?:(?:[a-zA-Z0-9-]+.)?[a-zA-Z]+.)?(';
-  const domainList = INVALID_DOMAIN_LIST.join('|');
-  const endingPattern = ')$';
-  return `${startingPattern}${domainList}${endingPattern}`;
-};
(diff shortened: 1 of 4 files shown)


mariahosfeld:
Scope needs to be expanded before release.

src/utils/isEmailValid.ts
mohitb35, line 1: 1. Has 'zod' been imported? I see a warning - Cannot find module 'zod' or its corresponding type declarations.
2. The INVALID_DOMAIN_LIST is not being considered any more.

mohitb35 (changes requested):
@prachigarg19 I'm not sure if this PR was ready for review, but it currently does not work.

mohitb35 (approved):

src/utils/isEmailValid.ts
```diff
@@ -1,22 +1,20 @@
+import { z } from 'zod';
```
norbertschuler, line 1: `zod` is imported only as a dependency from the `next` package in our `package-lock.json`. But I would suggest adding all packages used in our code directly in our `package.json` with `npm add zod` and not rely on it being loaded by some dependency!

norbertschuler (changes requested):

src/utils/isEmailValid.ts
norbertschuler: Btw. you did add the package `zod` in https://github.com/Plant-for-the-Planet-org/planet-donations/pull/440 ;-)

norbertschuler: Did that myself now.

mohitb35 (approved):

PR #1893, opened 2023-10-13
Resolve codefactor errors

prachigarg19:
Files covered:

1. src\features\user\Profile\components\MyTrees\MyTrees.tsx 
2. src/tenants/common/LeaderBoard/index.tsx
3.  src\features\projects\components\maps\Explore.tsx 
4. src\features\user\Profile\ProjectsContainer.tsx 
5. src\features\user\Profile\components\AddTargetModal.tsx 
6. src\features\user\Profile\components\UserShareAndSupport.tsx 
7.  src\features\user\Profile\components\SocialShareContainer.tsx

### DIFF
--- a/src/features/common/types/user.d.ts
+++ b/src/features/common/types/user.d.ts
@@ -1,23 +1,5 @@
-import { UserType, Score, Image } from '@planet-sdk/common';
+import { Image } from '@planet-sdk/common';
 import { Geometry } from '@turf/turf';
-
-// TODOO - add PublicUser type to common sdk
-export interface PublicUser {
-  id: string;
-  slug: string;
-  type: UserType;
-  image?: string;
-  url?: string;
-  urlText?: string;
-  displayName: string;
-  score: Score;
-  supportedProfile?: string;
-  /** @deprecated do not use */
-  isReviewer: boolean;
-  bio?: string;
-  /** @deprecated do not use */
-  hasLogoLicense?: boolean;
-}
 
 export interface ContributionType {
   type: string;
--- a/src/features/common/types/user.d.ts
+++ b/src/features/common/types/user.d.ts
@@ -1,4 +1,5 @@
-import { UserType, Score } from '@planet-sdk/common';
+import { UserType, Score, Image } from '@planet-sdk/common';
+import { Geometry } from '@turf/turf';
 
 // TODOO - add PublicUser type to common sdk
 export interface PublicUser {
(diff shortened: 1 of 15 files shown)


src/features/common/types/user.d.ts
```diff
@@ -17,3 +18,33 @@ export interface PublicUser {
...
+  type: string;
+  plantDate?: string;
+  country: string;
+  images: Image[];
+  treeCount: number;
+  unitType: string;
+  project: Project;
+  recipient?: Owner | null;
+  giver?: Owner | null;
+}
```
mohitb35, line 38: I'd suggest a small renaming here. `ContributionItem` in place of `ContributionType`, and `ContributionProperties` instead of `Properties` as it is too generic.

src/features/common/types/user.d.ts
```diff
@@ -17,3 +18,33 @@ export interface PublicUser {
...
+  id: string;
+  type: string;
+  plantDate?: string;
+  country: string;
+  images: Image[];
+  treeCount: number;
+  unitType: string;
+  project: Project;
+  recipient?: Owner | null;
+  giver?: Owner | null;
```
mohitb35, line 37: The type may be identical, but `Owner` does not seem to be an appropriate name for the type used both by recipient and giver. 

Perhaps `Owner` could be renamed to `PersonOrOrganization` or `AvatarType` or something similar and updated in all places?

src/features/projects/components/PopupProject.tsx
```diff
@@ -19,8 +19,8 @@ interface Props {
   open: boolean;
   handleOpen: Function;
   handleClose: Function;
-  buttonRef: any;
-  popupRef: any;
+  buttonRef: MutableRefObject<null>;
+  popupRef: MutableRefObject<null>;
```
mohitb35, line 23: 1. These seem to be the only properties of `Props` being actually used. Perhaps the rest could be removed?
2. The types for `buttonRef`/`popupRef` seems to be incomplete. Is the type of the ref that can be held always `null` or is that only when initialized in `Markers.tsx`?

mohitb35 (changes requested):
@prachigarg19 Please go through my comments.

src/features/user/Profile/components/UserShareAndSupport.tsx
```diff
@@ -59,7 +59,7 @@ export default function UserShareAndSupport({ userprofile }: any) {
 
   return ready ? (
     <div style={{ position: 'relative' }}>
-      {showSocialBtn && screenWidth > 600 && (
+      {showSocialBtn && screenWidth && screenWidth > 600 && (
```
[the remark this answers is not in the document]
prachigarg19: Okay, I'll keep this in mind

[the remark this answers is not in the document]
mohitb35: This was left as is, so I updated the code

mohitb35 (changes requested):
@prachigarg19 Could you address the single point here? Otherwise all good.

I've made some minor changes, please pull the latest code.

mohitb35 (approved):
LGTM

PR #1895, opened 2023-10-20
Updates ad in payment history

mohitb35:
### Changes
- replaces current iframe with `Advertisement` (new reusable component)
- minor style updates for `History`
- updates `MuiButton` to correctly forward props to resolve TS warnings

### Pending
- [ ] Update the link from the current dummy link
- [ ] German translations (after merge to develop)

### Screenshots
**New Ad - Mobile**
mobile ad

**New Ad - Large Screen**
large screen ad


mariahosfeld (commented):
Looks good functionality wise.

src/features/user/Account/components/MembershipCta.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const { token } = useUserProps();
+
+  return (
+    <a
+      className={`${styles.membershipCta} ${
+        placement === 'top'
+          ? styles.membershipCtaTop
+          : styles.membershipCtaRight
+      }`}
+      href={getDonationUrl('proj_LOxkf5GYI054Fi0HcEUF3dKu', token)}
```
norbertschuler, line 23: Is it ok for hardcode the project id `proj_LOxkf5GYI054Fi0HcEUF3dKu` here - will it be the same for the production and staging system and not getting changed that often?

mohitb35: This is tentative, on the assumption that the project does not change (as the CTA drives contributions to the membership project)

@sagararyal your input here?

sagararyal: it can be hardcoded as long as it is id from prod.

PR #1909, opened 2023-11-07
Potential fix: login issue on TTC

mohitb35:
On logging in, the user is redirected to `/profile` directly rather than through an intermediate step of `/t/[slug]`. 

This applies if there is no redirect link available.


mariahosfeld (commented):
Login works in test deployment, treecounter page also works.

PR #1911, opened 2023-11-08
Downgrades next js to 13.4.19

mohitb35:
This is a temporary fix due to Netlify issues with Next JS 13.5+.


mariahosfeld (commented):
Did general functionality checks, everything works as it should.

norbertschuler (approved):
LGFM

PR #1922, opened 2023-11-16
Change markers for conservation projects

prachigarg19:
Change conservation markers style to: 
image


src/features/projects/components/maps/Markers.tsx
```diff
@@ -45,16 +45,13 @@ export default function Markers({
...
-        ? 'conservationMarker'
-        : 'topUnapproved';
     return project.isTopProject && project.isApproved
       ? 'topApproved'
       : project.allowDonations
-      ? topUnapprovedType
+      ? 'topUnapproved'
       : 'notDonatable';
   };
+  console.log(searchedProject.properties);
```
norbertschuler, line 54: You should remove this again if it works as expected.

src/features/projects/styles/ProjectsMap.module.scss
```diff
@@ -738,10 +738,6 @@
   background-color: #828282;
 }
 
-.conservationMarker {
```
norbertschuler, line 741: This style is still used at https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/develop/src/features/projects/components/maps/Location.tsx#L31

norbertschuler (changes requested):
Please check the code where the deleted style is still used.

src/features/projects/styles/ProjectsMap.module.scss
prachigarg19: @mariahosfeld do I need to change markers used here as well?

norbertschuler (approved):
LGFM

PR #1923, opened 2023-11-17
Improve markers rendering

prachigarg19:
Rendering markers such that all grey markers are in the background, the next layer is all green markers and on top of it all we have the top-projects (gold markers)

### DIFF
--- a/src/features/projects/components/maps/Home.tsx
+++ b/src/features/projects/components/maps/Home.tsx
@@ -42,43 +42,37 @@ export default function Home({
     setViewPort(newViewport);
   }, []);
 
-  const topProjects = searchedProject.filter(
+  const topApprovedProjects = searchedProject.filter(
     (project) =>
       project.properties.isTopProject && project.properties.isApproved
   );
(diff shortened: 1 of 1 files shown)


src/features/projects/components/maps/Home.tsx
```diff
@@ -42,15 +42,37 @@ export default function Home({
...
+
+  const topUnapprovedProjects = searchedProject.filter(
+    (project) =>
+      project.properties.allowDonations &&
+      !(project.properties.isTopProject && project.properties.isApproved)
+  );
+
+  const notDonatableProjects = searchedProject.filter(
+    (project) =>
+      !project.allowDonations &&
```
norbertschuler, line 58: Why do you use `project.allowDonations` here while using `project.properties.allowDonations` above?

prachigarg19: Oh I forgot to make the correction here

norbertschuler (approved):
LGFM

Issue #1872, opened 2023-09-27
Typescript warning: Type 'unknown[]' is not assignable to type 'SampleTree[]' (SampleTrees.tsx)


mohitb35:
The warning below has been left unresolved as resolution of this may require code changes to add logic that validates the contents of the csv. This should be addressed later (not high priority)

```
[{
	"resource": "/Users/mohit3/plant-for-the-planet/planet-webapp/src/features/user/TreeMapper/Import/components/SampleTrees.tsx",
	"owner": "typescript",
	"code": "2322",
	"severity": 8,
	"message": "Type 'unknown[]' is not assignable to type 'SampleTree[]'.\n  Type 'unknown' is not assignable to type 'SampleTree'.",
	"source": "ts",
	"startLineNumber": 71,
	"startColumn": 19,
	"endLineNumber": 71,
	"endColumn": 30
}]
```

mariahosfeld:
Resolved via #1930

PR #1930, opened 2023-11-28
Validate sample tree import data

prachigarg19:
https://github.com/Plant-for-the-Planet-org/planet-webapp/issues/1872

### DIFF
--- a/src/features/user/TreeMapper/Import/components/SampleTrees.tsx
+++ b/src/features/user/TreeMapper/Import/components/SampleTrees.tsx
@@ -96,12 +96,12 @@ export default function SampleTrees({
       reader.onload = (event) => {
         const csv = event.target?.result;
         if (typeof csv !== 'string') return;
-        Papa.parse(csv, {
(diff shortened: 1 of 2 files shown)


src/features/user/TreeMapper/Import/components/SampleTrees.tsx
```diff
@@ -67,15 +99,30 @@ export default function SampleTrees({
...
-            const resultsData: SampleTree[] = results.data;
-            const sampleTrees: SampleTree[] = resultsData.map((sampleTree) => {
-              return {
-                ...sampleTree,
-                otherSpecies: sampleTree.otherSpecies,
-              };
-            });
-            addSampleTrees(sampleTrees);
+            const parsedHeaders = results.meta.fields || [];
+            const resultsData = results.data as SampleTree[];
```
mohitb35, line 103: Instead of casting to specify the type of results data, you can use

`Papa.parse<SampleTree>()` instead of `Papa.parse()` as this specifies the type of result data.

PR #1936, opened 2023-12-04
Reorder project list

prachigarg19:


mariahosfeld:
Functionality is not working correctly, projects that I expect in the bottom part of the list are shown to me as the first ones.

norbertschuler (approved):
So for tenants with `topProjectsOnly=True` we do not sort the projects list and for `topProjectsOnly=False` we always sort the project list?

If that's the required logic, then it is ok for me.

PR #1937, opened 2023-12-06
Fix certificate toggle visibility condition

prachigarg19:
The condition for showing certificate toggle in `DetailedAnalysis` while editing/creating new project was changed from `uploadedFiles && !uploadedFiles.length > 0` to `uploadedFiles && uploadedFiles.length > 0` due to which the toggle was no longer visible for projects with no certificates. This PR deals with fixing that condition.


mohitb35 (approved):
LGTM

PR #1947, opened 2023-12-14
Resolves treemapper import bug (multiple species import)

mohitb35:
This removes the additional fields that were being sent while creating a plant location with multiple species. It also updates types for form data and API response where needed.

### DIFF
--- a/src/features/user/TreeMapper/Import/components/PlantingLocation.tsx
+++ b/src/features/user/TreeMapper/Import/components/PlantingLocation.tsx
@@ -37,7 +37,9 @@ import {
   GeometryObject,
 } from 'geojson';
 import { Species } from '../../../../common/types/plantLocation';
-import { PlantLocation } from '../../Treemapper';
+import { PlantLocation, PlantingLocationFormData } from '../../Treemapper';
+import { DevTool } from '@hookform/devtools';
+import { SetState } from '../../../../common/types/common';
 
 const dialogSx: SxProps = {
   '& .MuiButtonBase-root.MuiPickersDay-root.Mui-selected': {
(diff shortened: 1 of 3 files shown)


mariahosfeld (commented):
Functionality works for me.

src/features/user/TreeMapper/Import/components/PlantingLocation.tsx
```diff
@@ -154,8 +156,8 @@ function PlantedSpecies({
 interface Props {
   handleNext: () => void;
   userLang: string;
-  plantLocation: PlantLocation | null;
-  setPlantLocation: Function;
+  // plantLocation: PlantLocation | null;
```
prachigarg19, line 159: Why is it commented out instead of being removed?

src/features/user/TreeMapper/Import/components/PlantingLocation.tsx
prachigarg19, line 1: 1. Could you update all the places where `<PlantingLocation/>` has been called to remove the `plantLocation` prop from there as well?
2. There are some errors in the file that calls this component in the ` setPlantLocation` value specifically, could you look into it?

Apart from this, everything looks good to me

prachigarg19 (changes requested):

src/features/user/TreeMapper/Import/components/PlantingLocation.tsx
mohitb35: Removed

src/features/user/TreeMapper/Import/components/PlantingLocation.tsx
mohitb35: I've addressed these

prachigarg19 (approved):

PR #1966, opened 2024-01-30
Redesign vegetation change data box

prachigarg19:
The expected component design is :
image
This PR uses storybook to create three stories:
1. Dropdown with states `open` and `close`
2. Slider with a single state that takes the position of thumb as prop
3. The combination of all the above components, Databox that has a single state `Preview` which takes the `startYear` as prop

- [x] Design approved
Left:
- [ ] Requires text to be shown when info icon is hovered upon for all dropdown items
- [ ] Api Integration


prachigarg19:
Since implementing a triangular thumb was a bit complicated, the design has been change to rounded square for now.

mariahosfeld:
@mohitb35 if this is no longer needed for reference, can you close this?

mohitb35:
Closing this as not needed

PR #1969, opened 2024-02-02
Design Forest Fire Popup

prachigarg19:
Component:
image


gangakhushlani (approved):

mohitb35:
The background for most of the popup is currently transparent.

Screenshot 2024-04-17 at 12 18 32 PM

src/temp/FirePopup/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+              <span>High</span> alert confidence
+            </p>
+            <div className={styles.setUpAlertsContainer}>
+              <p className={styles.setUpAlerts}>
+                Set up alerts with <span>FireAlert</span>
+              </p>
+              <RightArrowIcon width={5} height={8} color={'#4F4F4F'} />
+            </div>
+          </div>
+        </div>
```
mohitb35, line 71: How could you structure the html for `popupContainer` for better semantics?

For example, would a `aside` tag be better than using a `div` for the outermost element? How can you identify the heading/label of this element?

src/temp/FirePopup/index.tsx
mohitb35, line 1: Add a comment to indicate that the contents are hardcoded. This component will need to be refactored based on the api/data that is needed.

mohitb35 (changes requested):
@prachigarg19 Please go through the comments above.

mohitb35 (approved):

PR #1970, opened 2024-02-05
Redesign project snippet components

prachigarg19:
Redesign project badges:
1. Top Project (`isApproved = true and isTopProject = true`)
image

2. Field Reviewed  (`isApproved = true and isTopProject = false`)
image

3. Off site reviewed (`isApproved = false and isTopProject = false`)
image

4. Non donatable (`allowDonations = false`)
image


sunilsabatp (approved):
LGTM

src/temp/ProjectBadge/ProjectBadge.tsx
```diff
@@ -0,0 +1,56 @@
...
+        title: t('projectDetails:fieldReviewed'),
+      };
+    else if (!isTopProject && !isApproved)
+      return {
+        icon: <OffSiteReviewedIcon width={10} color={'var(--light)'} />,
+        title: t('projectDetails:offSiteReviewed'),
+      };
+  };
+  return (
+    <div className={styles.topProjectBadge}>
```
mohitb35, line 45: Correct the className as the purpose is no longer to show only top projects in the badge.

src/temp/ProjectBadge/ProjectBadge.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <div className={styles.badgeIcon}>
+        {getTitleAndIcon(isApproved, isTopProject, allowDonations)?.icon}
+      </div>
+      <div className={styles.badgeTitle}>
+        {getTitleAndIcon(isApproved, isTopProject, allowDonations)?.title}
+      </div>
+    </div>
```
mohitb35, line 52: There isn't a need to call the function `getTitleAndIcon` twice.

src/temp/ProjectBadge/ProjectBadge.tsx
```diff
@@ -0,0 +1,56 @@
...
+      return {
+        icon: <TopProjectIcon color={'var(--light)'} width={11} />,
+        title: t('common:topProject'),
+      };
+    else if (isTopProject && !isApproved)
+      return {
+        icon: <FieldReviewedIcon width={10} color={'var(--light)'} />,
+        title: t('projectDetails:fieldReviewed'),
+      };
+    else if (!isTopProject && !isApproved)
```
mohitb35, line 38: If a project is not approved, can it appear on the webapp?

What is the definition of offSiteReviewed?

src/temp/ProjectBadge/ProjectBadge.tsx
```diff
@@ -0,0 +1,56 @@
...
+      return {
+        icon: <OffSiteReviewedIcon width={10} color={'var(--light)'} />,
+        title: t('projectDetails:offSiteReviewed'),
+      };
+  };
```
mohitb35, line 43: It's a good practice to specify the return type of a function as well, in order to catch unexpected return types.

In this case, the function can return `undefined`. However, we're still rendering empty divs for project badge, badge icon, and badge title in that case. This does not make a difference while rendering the content, but it's still not advisable.

mohitb35: This will get called twice for every project in the list, at every render. I'd suggest using useMemo for the calculation, as this logic will not change for a prohect after it is rendered.

mohitb35 (changes requested):
@prachigarg19 please go through the comments above.

src/temp/ProjectBadge/ProjectBadge.tsx
mohitb35: @prachigarg19 Please update the logic for `FieldReviewed` projects, and merge this in.

mohitb35 (approved):

PR #1974, opened 2024-02-08
Feature/design project info container

prachigarg19:
Current : 
imageimage image image

Before:
image


sunilsabatp:
image

When I hover over the InfoIcon, the marked portion moves slightly upward.

src/temp/ProjectInfo/MoreInfoContainer.tsx
```diff
@@ -0,0 +1,142 @@
+import { Fragment } from 'react';
```
sunilsabatp, line 1: remove the unused import

src/temp/ProjectInfo/MoreInfoContainer.tsx
```diff
@@ -0,0 +1,142 @@
...
+      <div className={styles.singleInfo}>
+        <div className={styles.halfInfo}>
+          <div className={styles.siteOwnershipTitle}>
+            <div> {t('manageProjects:siteOwnership')}</div>{' '}
+            <div className={styles.siteOwnershipLabelContainer}>
+              {siteOwnershipType.map((type) => (
+                <span key={type}>
+                  {translatedSiteOwnership(type)} · {t('manageProjects:since')}{' '}
+                  {acquiredSince}
+                </span>
```
sunilsabatp, line 93: Instead of having an inline arrow function in JSX, it's better to extract it as a separate function. This improves readability and makes the code easier to understand.
```
const renderSiteOwnershipType = (type) => (
  <span key={type}>
    {translatedSiteOwnership(type)} · {t('manageProjects:since')} {acquiredSince}
  </span>
);

// Inside JSX
{siteOwnershipType.map(renderSiteOwnershipType)}
```

sunilsabatp:
image
Could you add `cursor:pointer` to the infoIcon ?

src/temp/ProjectInfo/ContactDetails.tsx
```diff
@@ -0,0 +1,90 @@
...
+          width={10.5}
+          color={`${'var(--primary-font-color)'}`}
+        />
+      ),
+      title: websiteURL
+        .replace('https://www.', '')
+        .replace('http://', '')
+        .replace('https://', '')
+        .replace('https://www.', '')
+        .split(/[/?#]/)[0],
```
sunilsabatp, line 42: URL manipulation (removing 'https://www.', 'http://', etc.) is being done directly in the component logic. It might be better to handle this manipulation outside of the component, possibly in a utility function, to improve readability and maintainability.

src/temp/ProjectInfo/ContactDetails.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  return (
+    <div className={styles.contactDetailsContainer}>
+      <SingleProjectInfoItem
+        title={t('donate:contactDetails')}
+        itemContent={
+          <div className={styles.contactText}>
+            {contactDetails.map((contact, index) => (
+              <a
+                href={contact.link}
```
sunilsabatp, line 67: Using the index of the array as the key when mapping through contactDetails might lead to unexpected behavior if the array's items change position or if items are added or removed. It's generally safer to use a unique identifier if available.

src/temp/ProjectInfo/ExternalCertification.tsx
```diff
@@ -0,0 +1,91 @@
...
+      ),
+    },
+    {
+      title: `${t('manageProjects:projectSpending')}`,
+      content: (
+        <div className={styles.spendingsContainer}>
+          {spendings.map((expense) => (
+            <div className={styles.spendingDetail} key={expense.id}>
+              <time className={styles.certificationLabel}>{expense.year}</time>
+              <div>
```
sunilsabatp, line 43: When using map() to render arrays of JSX elements, it's important to provide a unique key prop for each element to help React efficiently update the DOM.

src/temp/ProjectInfo/ExternalCertification.tsx
```diff
@@ -0,0 +1,91 @@
...
+    },
+    {
+      title: `${t('projectDetails:progressReports')}`,
+      content: (
+        <div className={styles.reportsContainer}>
+          {progressReports.map((report, index) => (
+            <div key={index}>
+              <div className={styles.certificationLabel}>{report}</div>
+              <div className={styles.downloadIcon}>
+                <DownloadIcon
```
sunilsabatp, line 66: When using map() to render arrays of JSX elements, it's important to provide a unique key prop for each element to help React efficiently update the DOM.

src/temp/ProjectInfo/InfoContainer.tsx
```diff
@@ -0,0 +1,143 @@
...
+            {t('projectDetails:plantingSeasons')}
+          </div>
+          <div className={styles.infoDetail}>
+            {plantingSeasons.map((season, index) => (
+              <Fragment key={seasons[season - 1].title}>
+                {seasons[season - 1].title}
+                {index === plantingSeasons.length - 2 ? (
+                  <> {t('manageProjects:and')} </>
+                ) : index === plantingSeasons.length - 1 ? (
+                  '.'
```
sunilsabatp, line 130: {plantingSeasons.map(...)}. It may be beneficial to extract this logic into separate functions or components to improve readability.

src/temp/ProjectInfo/InfoContainer.tsx
```diff
@@ -0,0 +1,143 @@
...
+          <div className={styles.infoTitle}>
+            {t('projectDetails:firstTreePlanted')}
+          </div>
+          <time className={styles.infoDetail}>
+            {formatDate(
+              firstTree.split('-')[1].length === 1 ||
+                firstTree.split('-')[2].length === 1
+                ? addZeroToDate(firstTree)
+                : firstTree
+            )}
```
sunilsabatp, line 92: Directly accessing array elements like firstTree.split('-')[1] and firstTree.split('-')[2] without proper validation can lead to errors if the input is not as expected. It would be safer to validate the array length before accessing its elements.

src/temp/ProjectInfo/InfoContainer.tsx
```diff
@@ -0,0 +1,143 @@
...
+    { id: 8, title: ready ? t('common:september') : '' },
+    { id: 9, title: ready ? t('common:october') : '' },
+    { id: 10, title: ready ? t('common:november') : '' },
+    { id: 11, title: ready ? t('common:december') : '' },
+  ];
+
+  return (
+    <div className={styles.projectInfoContainer}>
+      <div className={styles.singleInfo}>
+        <div className={styles.halfInfo}>
```
sunilsabatp, line 64: There are patterns repeated in the code, such as the structure of <div className={styles.halfInfo}>...</div>. You could consider creating a reusable component for these patterns to reduce duplication.

src/temp/ProjectInfo/ExternalCertification.tsx
prachigarg19: I'm already using a key

src/temp/ProjectInfo/InfoContainer.tsx
prachigarg19: I'm already using a reusable component for the same, but this particular structure couldn't be accomm0dated withing it hence i haven't used it here

src/temp/ProjectInfo/ContactDetails.tsx
prachigarg19: This would not be the case as this is a static array

sunilsabatp (approved):

src/temp/ProjectInfo/SingleProjectInfoItem.tsx
```diff
@@ -0,0 +1,20 @@
...
+import styles from './ProjectInfo.module.scss';
+
+interface Props {
+  title: ReactElement | string;
+  itemContent: ReactElement;
+}
+
+const SingleProjectInfoItem = ({ title, itemContent }: Props) => {
+  return (
+    <div className={styles.singleInfo}>
```
mohitb35, line 11: I'd suggest using a `section` tag instead of `div` for the outermost element

src/temp/ProjectInfo/ContactDetails.tsx
```diff
@@ -0,0 +1,94 @@
...
+              >
+                <div className={styles.icon}>{contact.icon}</div>
+                <div className={styles.title}>{contact.title}</div>
+                <div className={styles.rightArrow}>
+                  <RightArrowIcon
+                    width={5}
+                    color={`${'var(--primary-font-color)'}`}
+                  />
+                </div>
+              </a>
```
mohitb35, line 85: Create a component out of this structure.

src/temp/ProjectInfo/ContactDetails.tsx
```diff
@@ -0,0 +1,94 @@
...
+    {
+      icon: (
+        <ViewProfileIcon
+          width={10.5}
+          color={`${'var(--primary-font-color)'}`}
+        />
+      ),
+      title: t('donate:viewProfile'),
+      link: '',
+    },
```
mohitb35, line 38: How do we link to the public profile page? You need to add a prop for that.

src/temp/ProjectInfo/ContactDetails.tsx
```diff
@@ -0,0 +1,94 @@
...
+      title: email,
+      link: `mailto:${email}`,
+    },
+  ];
+
+  return (
+    <div className={styles.contactDetailsContainer}>
+      <SingleProjectInfoItem
+        title={t('donate:contactDetails')}
+        itemContent={
```
mohitb35, line 67: Perhaps not as a part of this PR, but we should keep the translations for this page within one file.

src/temp/ProjectInfo/ProjectInfo.module.scss
```diff
@@ -0,0 +1,200 @@
...
+        border-radius: 50%;
+        display: flex;
+        align-items: center;
+        justify-content: center;
+        background-color: $backgroundColor;
+        flex-shrink: 0;
+    }
+    .title {
+        font-size: $fontXXSmall;
+        font-weight: 600;
```
mohitb35, line 160: For some reason, font-weight: 600 and font-weight: 700 look almost the same, perhaps this needs to become font-weight: 500

src/temp/ProjectInfo/ExternalCertification.tsx
```diff
@@ -0,0 +1,116 @@
...
+                  </a>
+                </div>
+              )}
+              {renderDownloadIcon()}
+            </div>
+          ))}
+        </div>
+      ),
+    },
+  ];
```
mohitb35, line 101: I suggest not using this approach of creating a `certificationContent` array, as there are enough differences between the content below each of the 3 headings. Try to create subcomponents for the content under each heading, and for the purple pills (`certificationLabel`) as well.

src/temp/ProjectInfo/SingleProjectInfoItem.tsx
```diff
@@ -0,0 +1,20 @@
...
+      <div className={styles.halfInfo}>
```
mohitb35, line 12: Why is this called `halfInfo`?

src/temp/ProjectInfo/InfoContainer.tsx
```diff
@@ -0,0 +1,143 @@
...
+              <InfoIconPopup
+                height={10}
+                width={10}
+                color={`${'rgba(var(--secondary-divider-color-new))'}`}
+              >
+                <div className={styles.infoIconPopupContainer}>
+                  {t('manageProjects:yearAbandonedInfo')}
+                </div>
+              </InfoIconPopup>
+            </>
```
mohitb35, line 93: Enclose the text/label within an `h1-h6` tag

src/temp/ProjectInfo/InfoContainer.tsx
mohitb35, line 1: The names `InfoContainer` and `MoreInfoContainer` do not indicate the purpose of the sections too well. Can these names be more informative?

src/temp/ProjectInfo/InfoContainer.tsx
```diff
@@ -0,0 +1,143 @@
...
+              {plantingDensity}
+              {maxPlantingDensity !== null
+                ? `-${maxPlantingDensity} ${t('manageProjects:treePerHa')}`
+                : ` ${t('manageProjects:treePerHa')}`}
+            </>
+          }
+        />
+        <SingleProjectInfoItem
+          title={t('manageProjects:employees')}
+          itemContent={<>{employees}</>}
```
mohitb35, line 131: Try to avoid the fragment structure (`<></>`) for a single element. It would be preferable to use a `<p>`, `span`, `div` tag to enclose a single item. Otherwise, you end up with html markup like below, which isn't the easiest to read. 

```
<div class="ProjectInfo_halfInfo___ZcGj">
	<div class="ProjectInfo_infoTitle__9SS4j">Employees</div>
	500
</div>
```

If I were inspecting this code, I may not understand what `500` represents without digging into the code. Enclosing this in `<div class="infoContent">` would make it easier to understand.

src/temp/ProjectInfo/InfoContainer.tsx
```diff
@@ -0,0 +1,143 @@
...
+                    firstTree.split('-')[2].length === 1
+                    ? addZeroToDate(firstTree)
+                    : firstTree
+                )}
+            </time>
+          }
+        />
+      </div>
+      <div className={styles.seperator}></div>
+      <div className={styles.singleInfo}>
```
mohitb35, line 117: `singleInfo` isn't the best name for this class. It seems like `row` or `infoRow` or something similar would be more apt.

mohitb35 (changes requested):
@prachigarg19 Please review the comments above

src/temp/ProjectInfo/SingleProjectInfoItem.tsx
prachigarg19: I followed the same structure as before. I'll name it better.

src/temp/ProjectInfo/InfoContainer.tsx
prachigarg19: I was using same class as https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1974#discussion_r1568767514 to use existing styles here, I've renamed the class to `singleRowInfoContainer`

src/temp/ProjectInfo/ContactDetails.tsx
prachigarg19: I have added new files `allprojects.json` and `projectDetail.json`, I could all the translations in the second file. Do I need to do that for this component specifically or all the components under projectinfo folder

mohitb35: I think this can be done at one go off the base PR.

src/temp/ProjectInfo/ExternalCertification.tsx
```diff
@@ -0,0 +1,62 @@
...
+  );
+};
+
+interface ExternalCertificationProps {
+  certification: string;
+  spendings: ProjectExpense[];
+  progressReports: number[];
+}
+
+const ExternalCertification = ({
```
mohitb35, line 27: Since this does not include only External Certifications, let's call this the `ProjectDownloads` component.

src/temp/ProjectInfo/ExternalCertification.tsx
```diff
@@ -0,0 +1,62 @@
...
+export const renderDownloadIcon = () => {
+  return (
+    <div className={styles.downloadIcon}>
+      <DownloadIcon
+        width={10}
+        color={`${'rgba(var(--certification-background-color-new))'}`}
+      />
+    </div>
+  );
+};
```
mohitb35, line 19: Let's split this out into a separate component in a separate file, with a `<button>` at the outermost level instead of a `<div>` - call it `DownloadButton`. Update imports accordingly

src/temp/ProjectInfo/KeyInfo.tsx
```diff
@@ -0,0 +1,134 @@
...
+              <InfoIconPopup
+                height={10}
+                width={10}
+                color={`${'rgba(var(--secondary-divider-color-new))'}`}
+              >
+                <div className={styles.infoIconPopupContainer}>
+                  {t('manageProjects:yearAbandonedInfo')}
+                </div>
+              </InfoIconPopup>
+            </h1>
```
mohitb35, line 93: Since an `h1` tag typically indicates the top level, and we only keep a single `h1` on a page (which will be the project name), this should probably be a `h2` or lower. Look at the page structure as a whole while deciding this. You want to have a heirarchy of 

h1 (title)
h2 (subtitles - multiple)
h3 (sub-subtitles - multiple)

... and so on.

src/temp/ProjectInfo/SingleProjectInfoItem.tsx
```diff
@@ -0,0 +1,20 @@
...
+interface Props {
+  title: ReactElement | string;
+  children: ReactElement;
+}
+
+const SingleProjectInfoItem = ({ title, children }: Props) => {
+  return (
+    <div className={styles.singleRowInfoContainer}>
+      <div className={styles.singleRowInfoContent}>
+        <div className={styles.singleRowInfoTitle}>{title}</div>
```
mohitb35, line 13: So that we provide a semantically meaningful header tag to the title content instead of a div, how about adding a default tag of `h2` (see if `h2` is the appropriate level, or something lower is needed) if the `title` is a string? 

Providing a ReactElement will override the default.

src/temp/ProjectInfo/KeyInfo.tsx
```diff
@@ -0,0 +1,134 @@
...
+        {index === plantingSeasons.length - 2 ? (
+          <span> {t('manageProjects:and')} </span>
+        ) : index === plantingSeasons.length - 1 ? (
+          '.'
+        ) : (
+          ', '
+        )}
+      </Fragment>
+    ));
+  };
```
mohitb35, line 75: I think this could be separated out to a `PlantingSeasons` component.

src/temp/ProjectInfo/ContactDetails.tsx
mohitb35: I've added a task for this in JIRA.

mohitb35 (approved):
@prachigarg19 I've left some minor feedback, but otherwise this looks good to merge in once that feedback is complete.

mohitb35:
@prachigarg19 One more thing - check for responsiveness in the larger mobile phone screens.

src/temp/ProjectInfo/SingleProjectInfoItem.tsx
prachigarg19: If I do this there will be an `h2` inside a `h2` element. This is the reason I went with fragments in the beginning as well. Below is the structure after the changes for `KeyInfo` component

`<div class="ProjectInfo_singleRowInfoContent__KdZ_q"><h2 class="ProjectInfo_singleRowInfoTitle__dtZ21"><h2 class="ProjectInfo_abandonmentTitle__obmbL">Abandonment<div aria-haspopup="true" style="cursor: pointer;"><svg </svg></div></h2></h2><p>Approx. 2005</p></div>`

mohitb35: It's conditional, so you would not end up with `h2` inside `h2`.

The logic would be
`if title is a string, use a h2 tag, otherwise, use a div tag`

PR #1984, opened 2024-02-21
Design graph component (CO2 Captured and Canopy Cover)

prachigarg19:
Component:
image

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -10,6 +10,7 @@ on:
       - feature/storybook-update
       - develop
       - feature/redesign-explore-btn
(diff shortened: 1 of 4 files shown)


sunilsabatp:
image

can you make the marked portion lighter instead of bold ?

mariahosfeld (changes requested):
- [ ] Per Figma design we do not have yearly steps, but rather monthly. Will depend on final integration, but should be considered for implementation
image
- [ ] Check design of vertical line that marks project onset: Per figma design it is not solid
image

mohitb35:
Keeping this open as reference, will close once this feature is ready.

PR #1985, opened 2024-02-22
dive into the project button

sunilsabatp:
image


prachigarg19 (changes requested):

src/temp/stories/DiveIntoMap.stories.tsx
```diff
@@ -1,5 +1,5 @@
 import type { Meta, StoryObj } from '@storybook/react';
-import ProjectView from '../diveButton/SingleProjectMap';
+import ProjectView from '../diveButton/DiveIntoMap';
```
prachigarg19, line 2: Could you change the function name to `DiveIntoMap` as well?

gangakhushlani (approved):

prachigarg19 (approved):

src/temp/diveButton/DiveIntoMap.tsx
```diff
@@ -0,0 +1,25 @@
+import DiveIcon from '../icons/DiveIcon';
+import style from './DiveIntoMap.module.scss';
+import { useTranslation } from 'next-i18next';
+
+interface DiveIntoMapProps {
+  active: boolean;
+}
+
+const DiveIntoMap = ({ active }: DiveIntoMapProps) => {
```
mohitb35, line 9: `active` isn't needed, so remove this and update props/stories.

src/temp/diveButton/DiveIntoMap.module.scss
```diff
@@ -0,0 +1,31 @@
+@import '../../theme/theme';
+.diveIntoMapMainContainer {
+  width: 306px;
+  height: 341px;
```
mohitb35, line 4: Why is the width hardcoded? Shouldn't this take the full screen width (accounting for padding/margin outside the container)?

Height can be left as is, but we probably need to adjust it while combining this element.

src/temp/diveButton/DiveIntoMap.tsx
mohitb35, line 1: 1. Separate the container and the button component, as the container used may actually be very different. i.e. isolate the button code to a separate component e.g. `ShowMapButton`
2. Reconsider the naming of this component, classes, interfaces used etc, as `DiveIntoMap` may not be relevant if the text changes. Something like `MapPreview` may be more relevant for the container component, and `ShowMapButton` for the button component. Accordingly, update file names as well.

src/temp/diveButton/DiveIntoMap.module.scss
```diff
@@ -0,0 +1,31 @@
...
+  position: relative;
+}
+
+.diveIntoProjectButton {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  gap: 6px;
+  position: absolute;
+  width: 274px;
```
mohitb35, line 16: Do not hardcode width, but let width be taken based on the container.

src/temp/diveButton/DiveIntoMap.module.scss
```diff
@@ -0,0 +1,31 @@
...
+  height: 40px;
+  background: $primaryColorNew;
+  border-radius: 8px;
+  bottom: 16px;
+  left: 16px;
```
mohitb35, line 21: Absolute positioning - is that really needed? Can't we use `display: flex` and `align-items: flex-end` on the container, and margin on the button?

Absolute positioning should be a last resort if nothing else works. While it may not seem like a big difference, it can create headaches later on when adjustments need to be made.

mohitb35 (changes requested):

mohitb35 (approved):

PR #2007, opened 2024-03-14
Updates caching key generation logic to include environment

mohitb35:
Includes VERCEL_ENV in caching key (getTenantConfig)

## Summary by CodeRabbit

- **Refactor**
	- Enhanced caching key with environment context in multi-tenancy support to improve data retrieval efficiency.


norbertschuler (approved):
LGFM - so I would use `preview` as default for local testing with the staging backend I assume?

mohitb35:
> LGFM - so I would use `preview` as default for local testing with the staging backend I assume?

Yes, that's correct

PR #1999, opened 2024-03-07
replaced core map library

sunilsabatp:
replaced core map package(`mapbox-gl -> maplibre-gl`)

PR #2015, opened 2024-03-21
replaced core map library

mohitb35:
replaced core map package(mapbox-gl -> maplibre-gl). 

Duplicate of #1999 which was deleted.


src/features/projects/components/maps/TimeTravel.tsx
```diff
@@ -68,14 +67,14 @@ export default function TimeTravel(): ReactElement {
       // A selector or reference to HTML element
       const container = '#comparison-container';
 
-      const _compare = new MapboxCompare(before, after, container, {
-        mousemove: false, // Optional. Set to true to enable swiping during cursor movement.
+      const compare = new MapboxCompare(before, after, container, {
```
prachigarg19, line 70: `compare` variable is not being used anywhere. Could you just declare the object without initializing the value to a variable.

src/features/projects/components/maps/TimeTravel.tsx
prachigarg19, line 1: In this file `//@ts-ignore` has been used. This practice is not recommended. Is there a reason why this has been done? There aren't many errors in this file if we remove this.

prachigarg19 (changes requested):

prachigarg19 (approved):

norbertschuler:
Will the replacement of mapbox-gl with maplibre-gl be continued somewhere? Would be nice as it could reduce the overhead of code libraries to download a little bit for the website.

mohitb35:
We are replacing mapbox-gl with maplibre-gl gradually. The new profile already uses this, and the upcoming facelift for the home page / project details page also uses maplibre.

mohitb35:
Closing as not needed

PR #2033, opened 2024-04-05
Replaces membership cta link

mohitb35:
Updates membership CTA link on /history, with a separate link for users seeing the page in German.


mohitb35:
> If it is ok not to use some shortlinks with pp.eco here, I am fine with it.

The pp.eco links (pp.eco/donor-circle, pp.eco/foerdermitgliedschaft) direct to donate.planet, so we need new links or we need to reconsider those redirects.

mariahosfeld (commented):

PR #2055, opened 2024-04-26
Resolve multiple api calls in data explorer

prachigarg19:
**Description**

There are multiple api calls being made at below places:
1. When adding a HID or selecting a date in the HID/Date filter, the application is making excessive API calls. Specifically, every time the user presses the backspace key in the input field, a new API call is triggered, leading to unnecessary server requests and potentially impacting performance.
2. In the Data Explorer section, selecting dates in the From or To fields triggers an API call each time a date is chosen from the calendar.

 

**Expected behavior**

For first point, debouncing should be implemented to reduce the number of calls per second

In the second point, API calls should be made only after the user has finalized their date selection, rather than for each individual date chosen from the calendar.

## Summary by CodeRabbit

- **New Features**
	- Improved the project location fetching mechanism in the Map component with a delayed execution to enhance performance.
	- Enhanced date selection controls in the Project Filter component for better user input management.

- **Refactor**
	- Implemented local state management for date handling in the Project Filter component, allowing for more precise control.

### DIFF
--- a/src/features/user/TreeMapper/Analytics/components/Map/index.tsx
+++ b/src/features/user/TreeMapper/Analytics/components/Map/index.tsx
@@ -49,6 +49,7 @@ import {
 import { ErrorHandlingContext } from '../../../../../common/Layout/ErrorHandlingContext';
 import PlantLocationDetails from './components/PlantLocationDetails';
 import MyForestMapCredit from '../../../../Profile/components/MyForestMap/microComponents/MyForestMapCredit';
+import { useDebouncedEffect } from '../../../../../../utils/useDebouncedEffect';
 
 const EMPTY_STYLE = {
   version: 8,
@@ -299,10 +300,18 @@ export const MapContainer = () => {
           return;
         }
       }
-      const getResponse = setTimeout(fetchProjectLocations, 500);
-      return () => clearTimeout(getResponse);
     }
   }, [project, species, queryType, fromDate, toDate]);
+
+  useDebouncedEffect(
+    () => {
+      if (project && species) {
+        fetchProjectLocations();
+      }
+    },
+    500,
+    [project, species, queryType, fromDate, toDate]
+  );
 
   // Set the map style to the default style
   // Currently this only shows Intervention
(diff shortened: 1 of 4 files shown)


norbertschuler:
We once upon a time invented https://github.com/Plant-for-the-Planet-org/planet-webapp/blob/develop/src/utils/useDebouncedEffect.ts to throttle the projects search to not trigger it for every single input change of the users in the search field. Maybe this concept would also be useful here?

norbertschuler (approved):
Works, but a little bit ugly code now having two hooks and also the type check for the paramters for useDebouncedEffect is failing now of course.
(I fixed some wrong label on the page saying "from" instead of "to".)

src/features/user/TreeMapper/Analytics/components/Map/index.tsx
```diff
@@ -299,10 +300,19 @@ export const MapContainer = () => {
...
   }, [project, species, queryType, fromDate, toDate]);
 
+  useDebouncedEffect(
+    () => {
+      if (project && species) {
+        fetchProjectLocations();
+      }
+    },
+    500,
+    [project, species, queryType, fromDate, toDate]
```
norbertschuler, line 313: You maybe should change this paramter to `deps: any[]` in  `useDebouncedEffect`.

prachigarg19: It then won't trigger a call to `fetchProjectLocations` function anytime we change the filter value

prachigarg19:
> Works, but a little bit ugly code now having two hooks and also the type check for the paramters for useDebouncedEffect is failing now of course. (I fixed some wrong label on the page saying "from" instead of "to".)

I could reset the changes in this case.

src/features/user/TreeMapper/Analytics/components/Map/index.tsx
norbertschuler: I did so now and do not see any problem. Can you explain this once more?

prachigarg19: Yes you're right, it won't cause any effect.

PR #2056, opened 2024-04-28
load total number of donated trees dynamically

norbertschuler:
Fixes https://planetfoundation.slack.com/archives/C015AEAU75Z/p1714117221663939?thread_ts=1713974028.645419&cid=C015AEAU75Z

Changes in this pull request:
- load the number of total donated trees dynamically

To-Do:
- [x] do we want to configure the URL for `https://automate.plant-for-the-planet.org/webhook/` in the environment like we already did for the Salesforce deplyoment defined as `WEBHOOK_URL`?
- do we really want to display this number dynamically?

### DIFF
--- a/app.json
+++ b/app.json
@@ -59,6 +59,11 @@
     "SITE_IMAGERY_API_URL": {
       "required": false,
       "description": "For showing vegetation change"
+    },
+    "WEBHOOK_URL": {
+      "required": true,
+      "description": "URL for webhooks of automation service",
+      "value": "https://automate.plant-for-the-planet.org/webhook"
     }
   },
   "formation": {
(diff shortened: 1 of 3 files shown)


mohitb35 (approved):
Codewise, LGTM.

I think the webhook url should be kept in the environment, we plan to do this while migrating the salesforce tenant to the multitenant project anyway.

Not sure if we want to hold off on release due to reasons of data consistency.

norbertschuler:
Ok, I'll just keep this as a draft until releases will work normally again.

sagararyal:
@mohitb35 What's the data inconsistency issue?

The numbers shared generally was from Q1. vs the query returns number for 1 week ago.

mohitb35:
There isn't any issue as such, but we do choose to hardcode certain figures at times, and I wondered if this was one of those cases.

mohitb35:
@norbertschuler We can release this now, since www1.plant-for-the-planet.org is now redirected to web.plant-for-the-planet.org, so changes will be visible.

norbertschuler:
> @norbertschuler We can release this now, since www1.plant-for-the-planet.org is now redirected to web.plant-for-the-planet.org, so changes will be visible.

Ok, I added using an env named WEBHOOK_URL which I also added to all deployments at Heroku and Vercel.

PR #2066, opened 2024-05-13
Feature/myforest map popup

sunilsabatp:
- Feature-> popup for registered tree
- Feature-> popup for contribution tree

- [x] integrate api
- [x] translation
- [x] define type


mariahosfeld:
Currently this PR is breaking when trying to check it out. After loading the new version of the profile, after a few seconds I get either a client side error or a completely white screen.
As there are no markers on the map, I cannot check the functionality of the pop-up.

src/features/user/MFV2/MyContributions/ProjectItemCard.tsx
```diff
@@ -93,8 +93,8 @@ const ProjectItemCard = ({
                 projectPurpose={project.purpose}
                 buttonText={
                   pageType === 'public'
-                    ? tProfile('donate')
-                    : tProfile('donateAgain')
+                    ? tProfile('donateAgain')
+                    : tProfile('donate')
```
mohitb35, line 97: Reverse this - the Donate Again button can only be seen when the profile type is private

src/features/user/MFV2/ContributionsMap/index.tsx
```diff
@@ -3,11 +3,16 @@ import Map, { MapStyle } from 'react-map-gl-v7/maplibre';
...
 import { NavigationControl } from 'react-map-gl-v7';
-import SinglePointMarkers from './Markers/SinglePointMarkers';
 import MapCredits from './Common/MapCredits';
+import Markers from './Markers';
+import { useRef, MutableRefObject } from 'react';
 import style from './Common/common.module.scss';
 import ContributionStats from './Common/ContributionStats';
 
+interface ContributionsMapProps {
+  pageType: 'public' | 'private';
```
mohitb35, line 13: Let's rename this to `profilePageType` in line with other places.

mariahosfeld (commented):
- [ ] - On desktop: For each declustered marker, I can see the pop-up, but I am unable to test the functionality of the donation button within the pop-up, as it is hidden as soon as the cursor leaves the corresponding marker.
- [ ] - Opposed to the design, the little triangles are present also for the project pop-ups, not just for the registered trees.
- [ ] - In German: On mobile the "Donate again" button is to wide and disrupts the layout by weired breakpoints (noted directly on mobile, not when simulating in browser)

mariahosfeld (changes requested):
Make sure that the pop-up always open to the top, for reference compare to main map. This will make sure, that it will not hide immediately.

mariahosfeld:
@prachigarg19 , please make sure that the unit is not hard coded but is rather taken from the response, I have test donations to m2 projects, but the pop-up shows trees as the unit

mariahosfeld (commented):
LGTM!

prachigarg19 (approved):

mariahosfeld (approved):

mohitb35:
Should we make adjustments to the Registered Trees popups as well?

Some observations:
1. Fonts for "Registered" and the date are a little too small, and the colors don't look quite right.
2. The shadow outside the Registered tree popup seems to be different, it does not match the shadow for the Donation popup.
3. I know the Figma design has an arrow, but the arrow here doesn't look great and might affect functionality. It's separated from the main popup body with a line. Also, the popup disappears as soon as I move off the marker, unlike for the Donation popup where I can hover above the popup itself to keep it in view. Should the arrow be removed? If not, it needs to look continuous with the popup body, and be closer to the marker itself so that it can remain in view when the user tries to hover above it.

Screenshot 2024-06-27 at 11 16 35 AM

Screenshot 2024-06-27 at 11 16 46 AM

src/features/user/MFV2/ContributionsMap/MyForestV2.module.scss
mohitb35, line 1: 1. File name should be updated. This is for the ContributionsMap, and should be named accordingly.
2. Several styles here seem to be unused / moved to a different file (e.g. mapCreditsContainer is now in the common.module.scss file). Clean this up.

src/features/user/MFV2/ContributionsMap/MyForestV2.module.scss
```diff
@@ -0,0 +1,223 @@
...
+  > div {
+    padding: 0px;
+  }
+}
+
+.customDonateButton {
+  padding: 5px 16px;
+  font-size: 12px;
+  white-space: wrap;
+}
```
mohitb35, line 36: Not sure this needed anymore, as the donate button in the popup is the standard tertiary WebappButton.

src/features/user/MFV2/ContributionsMap/Popup/ProjectInfoSection.tsx
```diff
@@ -0,0 +1,76 @@
...
+          projectPurpose={purpose}
+          buttonText={
+            profilePageType === 'private'
+              ? tProfile('myContributions.donateAgain')
+              : tProfile('myContributions.donate')
+          }
+          projectSlug={slug}
+          contributionUnitType={unitType}
+          customClass={style.customDonateButton}
+        />
```
mohitb35, line 70: `customClass` may not be needed anymore, as we're using a standard tertiary WebappButton

src/features/user/MFV2/MyContributions/DonateButton.tsx
```diff
@@ -18,6 +18,7 @@ type DonateButtonProps = (SupportedDonationButton | RegularDonationButton) & {
   projectSlug: string;
   contributionUnitType: 'tree' | 'm2';
   buttonText: string;
+  customClass?: string;
 };
```
mohitb35, line 22: While `customClass` isn't needed any more, it can be left in `DonateButtonProps` the make it easy to customize the button.

src/features/user/MFV2/MyContributions/index.tsx
```diff
@@ -52,9 +52,9 @@ const MyContributions = ({ profilePageType, userProfile }: ProfileV2Props) => {
       <ListHeader
         profilePageType={profilePageType}
         displayName={
-          userProfile.type === 'individual'
-            ? userProfile.displayName.split(' ').slice(0, 1)[0]
-            : userProfile.displayName
+          userProfile?.type === 'individual'
+            ? userProfile?.displayName.split(' ').slice(0, 1)[0]
+            : userProfile?.displayName
```
mohitb35, line 57: Is the null check on `userProfile` needed? Can it ever be `null` or `undefined`?

src/theme/global.scss
```diff
@@ -748,3 +748,7 @@ input[type='number'] {
     font-size: 30px;
   }
 }
+
+// .maplibregl-popup-anchor-bottom .maplibregl-popup-tip {
+//   display: none !important;
+// }
```
mohitb35, line 754: Remove if not needed

src/features/common/WebappButton/index.tsx
```diff
@@ -38,7 +38,11 @@ function WebappButton({
           {otherProps.icon !== undefined && (
             <div className={styles.webappButtonIcon}>{otherProps.icon}</div>
           )}
-          <div className={styles.webappButtonLabel}>{otherProps.text}</div>
+          <div
+            className={`${styles.webappButtonLabel} ${otherProps.buttonClasses}`}
+          >
+            {otherProps.text}
+          </div>
```
mohitb35, line 45: Is buttonClasses needed for the label? It was intended to be applied at the outer `button` element.

src/features/user/MFV2/ContributionsMap/Popup/ProjectInfoSection.tsx
```diff
@@ -0,0 +1,76 @@
...
+      ? tProfile('myForestMap.plantedTree', {
+          count: Number.isInteger(totalContributionUnits)
+            ? totalContributionUnits
+            : totalContributionUnits.toFixed(2),
+        })
+      : tProfile('myForestMap.savedArea', {
+          area: Number.isInteger(totalContributionUnits)
+            ? totalContributionUnits
+            : totalContributionUnits.toFixed(2),
+        });
```
mohitb35, line 41: We need to consider both restored and conserved area.

e.g. 
 - 150 m² conserved (if project purpose is conservation)
 - 150m² restored (if project purpose is trees and unitType is m2)

prachigarg19: @mariahosfeld said that we don't need to show we just need to show `m2`

src/utils/superclusterConfig.ts
```diff
@@ -1,12 +1,15 @@
...
+  DonationProperties,
+  MyContributionsSingleRegistration,
+} from '../features/common/types/myForestv2';
+import {} from '../features/common/Layout/MyForestContextV2';
 
-const clusterConfigV2 = {
+const _clusterConfigV2 = {
   radius: 40,
   maxZoom: 3,
 };
```
mohitb35, line 15: When we use `_`, warnings for unused variables are suppressed as per our current eslint rules.

public/static/locales/en/profile.json
```diff
@@ -18,6 +18,7 @@
     },
     "myForestMap": {
       "plantedTree": "{count, plural, =1 {# Tree} other {# Trees}}",
+      "savedArea": "{area} m²",
```
mohitb35, line 21: May not be needed as you can use "areaContributed"

public/static/locales/en/profile.json
mohitb35, line 1: We may consider removing some duplication in translations in this file, in a separate PR.

For example, there are several terms for "trees" (with pluralization).

mohitb35 (changes requested):
@prachigarg19 I've left some feedback.

src/features/user/MFV2/MyContributions/index.tsx
prachigarg19: I saw the error stating `cannot read properties of undefined` while I was working in the PR though it does not seem to occur now but I'm unsure about the steps to reproduce the same. Though, it does not seem to occur now so I've removed the check for now

public/static/locales/en/profile.json
prachigarg19: We don't need to show `conserved` and `restored` keyword hence I didn't use `areaContributed`

src/features/common/WebappButton/index.tsx
prachigarg19: I've removed the class, seems to be added by mistake

src/features/user/MFV2/MyContributions/index.tsx
mohitb35: Should be safe as we're checking for profile before rendering `MyContributions` in any case.

src/features/user/MFV2/ContributionsMap/Markers/PointMarkers.tsx
```diff
@@ -22,29 +28,72 @@ function isRegistration(
...
+      >
+        {isRegistration(superclusterResponse.properties) ? (
+          <>
+            {(isCursorOnMarker || isCursorOnPopup) && (
+              <RegisteredTreesPopup
+                superclusterResponse={
+                  superclusterResponse as PointFeature<MyContributionsSingleRegistration>
+                }
+              />
+            )}
```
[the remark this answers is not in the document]
mohitb35: Let's leave the `isCursorOnPopup` condition in, update the RegisteredTreesPopup to accept `setIsCursorOnPopup` and continue to display the popup while the cursor is above it.

public/static/locales/en/profile.json
mohitb35: Created a separate issue

mohitb35 (approved):
LGTM

PR #2068, opened 2024-05-15
logic for cluster marker

sunilsabatp:
Feature-> logic for cluster marker 

- [x] fix type issue 
- [x] integrate api
- [x] clustering of registered tree


mariahosfeld (commented):
For testing, I registered some trees near each other, and on the highest zoom level, everything looked fine, but when zooming in, the declustering seems a bit confusing, as it just goes from the cluster marker to the single marker. This is probably due to the close locations, but should be covered.

public/assets/images/icons/myForestV2Icons/ClusterMarker/RegisteredTreeClusterMarkerIcon.tsx
mohitb35, line 1: Move this to `/myForestMapIcons` since we're using that for the others.

src/features/user/MFV2/ContributionsMap/Markers/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+          <RegisteredTreeClusterMarker superclusterResponse={geoJson} />
+        ) : (
+          <PointMarkers superclusterResponse={geoJson} />
+        );
+      })}
+    </>
+  ) : (
+    <></>
+  );
+};
```
mohitb35, line 70: Add a unique key for each marker. Currently, I see a warning in the console:

```
Warning: Each child in a list should have a unique "key" prop. Check the render method of `Markers`. 
```

src/utils/superclusterConfig.ts
mohitb35, line 1: Clean up unused code in this file.

src/features/user/MFV2/ContributionsMap/Markers/ClusterMarker.tsx
```diff
@@ -0,0 +1,102 @@
...
+  viewport: ViewportProps;
+  mapRef: MutableRefObject<null>;
+}
+
+export type ExtractedData = {
+  unitType: UnitTypes;
+  classification: TreeProjectClassification;
+  purpose: ProjectPurposeTypes;
+  contributionCount: number;
+};
```
mohitb35, line 29: The name `ExtractedData` does not give a clear idea about the purpose. How can this be named more relevantly?

src/utils/myForestV2Utils.ts
```diff
@@ -0,0 +1,122 @@
...
+  // store all project whose unitType and purpose are different than maxContributingProject
+  extractedData.map((item) => {
+    if (
+      item.unitType !==
+        (maxContributingObject !== null && maxContributingObject.unitType) ||
+      item.purpose !==
+        (maxContributingObject !== null && maxContributingObject.purpose)
+    )
+      remainingProjects.push(item);
+  });
```
mohitb35, line 105: Shouldn't this be a `forEach`, as we're just iterating through `extractedData` here.

src/utils/myForestV2Utils.ts
```diff
@@ -0,0 +1,122 @@
...
+    tertiaryProjectColor: '',
+    secondaryProjectColor: '',
+    mainProjectColor: '',
+  };
+};
+
+export const extractAndClassifyProjectData = (
+  clusterChildren: PointFeature<AnyProps>[] | undefined
+) => {
+  const extractedData: ExtractedData[] = [];
```
mohitb35, line 64: `extractedData` - what is the purpose? Name accordingly. Also, it shouldn't be given the same name as a single piece of `ExtractedData` - there isn't a way to know that this is a list from the name.

src/utils/myForestV2Utils.ts
```diff
@@ -0,0 +1,122 @@
...
+      unitType: item.properties.projectInfo.unitType,
+      classification: item.properties.projectInfo.classification,
+      purpose: item.properties.projectInfo.purpose,
+      contributionCount: item.properties.contributionInfo.contributionCount,
+    };
+    extractedData.push(extractedItem);
+  });
+
+  // Loop through the array to find the object with the maximum contributionCount
+  const { maxContributingObject } = extractedData.reduce(
```
mohitb35, line 81: Should this be `maxContributingProject` instead of `maxContributingObject`?

src/utils/myForestV2Utils.ts
```diff
@@ -0,0 +1,122 @@
...
+    (acc: Accumulator, item: ExtractedData | null) => {
```
mohitb35, line 82: It'd be easier to read this if we used `maxContributionDetails` instead of `acc`, and a similar name in place of the `Accumulator` type.

src/features/user/MFV2/ContributionsMap/Markers/ClusterMarker.tsx
mohitb35: e.g. `ExtractedProjectData`

src/utils/myForestV2Utils.ts
```diff
@@ -0,0 +1,122 @@
...
+    const key = unitType + '-' + purpose;
+
+    if (!uniqueCombinations.has(key)) {
+      uniqueCombinations.set(key, obj);
+    }
+  });
+
+  const uniqueObjects: ExtractedData[] = Array.from(
+    uniqueCombinations.values()
+  );
```
mohitb35, line 119: I think there's scope for better naming here.

`uniqueCombinations`, `uniqueObjects` etc does not convey much about the purpose of the variable.

src/utils/myForestV2Utils.ts
```diff
@@ -0,0 +1,122 @@
...
+  const uniqueCombinations = new Map();
+  // Loop through the array to find  the object with the unique purpose and unit type
+  remainingProjects.forEach((obj) => {
+    const { unitType, purpose } = obj;
+    const key = unitType + '-' + purpose;
+
+    if (!uniqueCombinations.has(key)) {
+      uniqueCombinations.set(key, obj);
+    }
+  });
```
mohitb35, line 115: This could be more efficient. We're looping through the data 4 times, perhaps we can loop through just once while keeping things readable?

src/utils/myForestV2Utils.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  const uniqueObjects: ExtractedData[] = Array.from(
+    uniqueCombinations.values()
+  );
+
+  return { uniqueObjects, maxContributingObject };
```
mohitb35, line 121: Should it be possible for `maxContributingObject` to be null?

mohitb35 (dismissed):
@sunilsabatp functionality works great, but readability could improve. Please go through my feedback.

mariahosfeld:
> @sunilsabatp functionality works great, but readability could improve. Please go through my feedback.

I agree, it works great. Also clustering/de-clustering with different project types works very well.

src/utils/myForestV2Utils.ts
sunilsabatp: This logic has been removed

src/utils/myForestV2Utils.ts
sunilsabatp: Yes, it is possible for `maxContributingProject `to be null based on the logic provided in the `extractAndClassifyProjectData` function. The function initializes `maxContributingProject` to null and updates it conditionally based on the highest` contributionCount` found among` clusterChildren`

prachigarg19 (approved):

PR #2073, opened 2024-05-22
Adds Leaderboard Api for myForest v2

mohitb35:
1. Implements the myForestv2.leaderboard API which accepts inputs > `slug` and `profileGuid`
2. Response contains
    - `mostRecent` - basic details for last 10 gifts received 
    - `mostTrees` - basic details for the top 10 gifters (aggregated by name)

### DIFF
--- a/src/server/procedures/myForestV2/leaderboard.ts
+++ b/src/server/procedures/myForestV2/leaderboard.ts
@@ -1,0 +1,105 @@
+import { z } from 'zod';
+import { procedure } from '../../trpc';
+import { Leaderboard } from '../../../features/common/types/myForestv2';
+import prisma from '../../../../prisma/client';
+import { TRPCError } from '@trpc/server';
+import { Prisma } from '@prisma/client';
+import { fetchProfile } from '../../utils/fetchProfile';
+import { fetchGroupTreecounterData } from '../../utils/fetchGroupTreecounterData';
+
+async function fetchMostRecentGifts(profileIds: number[]) {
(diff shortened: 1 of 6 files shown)


src/server/procedures/myForestV2/leaderboard.ts
```diff
@@ -0,0 +1,109 @@
...
+			metadata->>'$.giver.name' as giverName
+		FROM gift
+		WHERE 
+			recipient_id IN (${Prisma.join(profileIds)}) 
+			AND deleted_at IS NULL 
+			AND value <> 0 
+			AND metadata->>'$.giver.name' IS NOT NULL
+			AND metadata->>'$.giver.name' <> ''
+		GROUP BY giverName
+		ORDER BY totalQuantity DESC
```
sunilsabatp, line 48: Extract the gift-related SQL queries into a separate service or utility module to keep the procedure clean

sunilsabatp (changes requested):

src/server/procedures/myForestV2/leaderboard.ts
mohitb35: I prefer to leave it as is because:

- They aren't used anywhere else, and are already within isolated functions. 
- Moving them elsewhere will make the code less readable, as I'd need to move between files to see the functionality. 

I've already kept some functions that could be reused in a utils folder.

sunilsabatp (approved):

PR #2076, opened 2024-05-27
Adds `purpose` property to leaderboard API

mohitb35:
Adds `purpose` property to `mostRecent` and `topGifter` results.


src/server/procedures/myForestV2/leaderboard.ts
```diff
@@ -92,6 +96,7 @@ export const leaderboardProcedure = procedure
       name: gift.giverName,
       units: gift.quantity,
       unitType: 'tree',
+      purpose: gift.purpose,
     }));
 
     // Fetch the top 10 gifters for the profile(s)
```
prachigarg19, line 102: Won't we need to add `purpose` for `most trees` array elements as well?

mohitb35: yes, missed that.

prachigarg19 (approved):
lgtm

PR #2077, opened 2024-05-28
Fixes queries for contributions API to avoid null gift plantdate 

mohitb35:
Update query logic while fetching gift plant date to consider `redemption_date` if `payment_date = null`


mariahosfeld:
For me the query makes sense and I would go ahead with merging it in.

PR #2078, opened 2024-05-29
Feature/progress bar

sunilsabatp:
UI for progress bar


mariahosfeld (changes requested):
Please change the following in the private profile:

- [x] Layout: Order of contribution types in the target setting module, number 2 and 3 from the top need to be changed to reflect the order in the progress bar
- [x] Functionality: Users should be able to toggle of the visibility of the target, even though they have set one (currently you can only toggle it of when you set the target to 0)

mariahosfeld (changes requested):
- [x] In the pop-up Area conserved target and Area restored target need to be switched to reflect the order in the progress-bar itself
- [x] Even though I am setting targets for all possible values, and see them after saving, when either viewing the public profile or reloading the private profile, I can again only see the progress for trees
- [x] Edit functionality cannot be available in the public profile version
- [ ] Seems like the community trees are not shown correctly in the progress bar
- [ ] Please check which value is shown for the community trees below the bar, seems to be the count of contributions, not the amount of trees
- [x] Small correction: please substitute "our community" with "your community"

mariahosfeld (changes requested):
I am now facing the issue that if I set a target without having made any contributions, I sometimes see instead of 0% NaN% (refer to screenshot)
image
When checking the public profile directly on my iPhone, I cannot see any progress bar (visible when simulated on computer). In the private profile I see more all progress bars, even though I have disable one while testing on the computer. When editing on mobile, the changes are not visible on computer.

sunilsabatp:
> Most things are done, but one think remains (was not able to detect it before): When I disable a target on my phone, and - after refreshing the page - re-enable it on my computer, then the field on the computer is empty. Same thing happens when I disable on my computer, then the field on my phone is empty. After a refresh the fields are then also empty on the device where it was disabled.

To address this issue, we need to update the /app/profile endpoint. Specifically, we need to include a key in the profile response that keeps a record of the toggle state. Since it is not possible to manage toggle states on the client side, the state will be reset if the page is reloaded

mariahosfeld:
> > Most things are done, but one think remains (was not able to detect it before): When I disable a target on my phone, and - after refreshing the page - re-enable it on my computer, then the field on the computer is empty. Same thing happens when I disable on my computer, then the field on my phone is empty. After a refresh the fields are then also empty on the device where it was disabled.
> 
> To address this issue, we need to update the /app/profile endpoint. Specifically, we need to include a key in the profile response that keeps a record of the toggle state. Since it is not possible to manage toggle states on the client side, the state will be reset if the page is reloaded

Lets move this point to later improvements.

mariahosfeld (approved):

src/utils/myForestV2Utils.js
mohitb35, line 1: Why is this a JS file?

mohitb35:
The progress bars of adjacent cells looks unaligned if the one cell has community contributions, and the other does not. I think it should always be in the same line. If community contributions (gifts) are present, then they can be shown below the bar.

Screenshot 2024-06-11 at 5 47 52 PM

mohitb35:
When no targets are set, and no scores are available, the TargetInitialize component is shown. Align the text and the button to the vertical center of the container.

Screenshot 2024-06-11 at 6 07 04 PM

mohitb35:
1. Refactor progress bars to use a common component e.g ForestProgressItem containing ProgressData and EditTarget

sunilsabatp:
> 1. The progress bars of adjacent cells looks unaligned if the one cell has community contributions, and the other does not. I think it should always be in the same line. If community contributions (gifts) are present, then they can be shown below the bar.
> 2. Check calculations for the percentage shown at the end of the bar, and the ratio of different portions of the bar. Test scenarios
> 
> * Target = 100, gift=10, personal=25 >> percentage shown should be 35%, and gift portion of bar should be 10% width, personal portion should be 25% width, rest should be 65% width
> * Target = 0, gift=10, personal=25 >> no percentage will be shown. Gift portion of bar would be (10/35), personal portion of bar would be (25/35)
> * Target = 0, gift=10, personal=0>>no percentage will be shown. Gift portion of bar would be 100%
> * Target=0, gift =0, personal=0>> don't show item itself
> 
> Screenshot 2024-06-11 at 5 47 52 PM

Done

sunilsabatp:
> When no targets are set, and no scores are available, the TargetInitialize component is shown.
> 
> 1. Align the text and the button to the vertical center of the container.
> 2. The button looks too large. I've set up common buttons in [Feature/my contributions #2083](https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/2083), so the button can be replaced after merging. Take a look and see if any adaptations need to be made.
> 
> Screenshot 2024-06-11 at 6 07 04 PM

I have referred the Figma design  -> https://www.figma.com/design/lT7AVzzAUM0NyThCyvacob/MyForest-Dashboard?node-id=2635-6591&t=dqO5GL8lJN9up6o3-4

src/features/user/MFV2/ForestProgress/microComponents/EmptyProgress.tsx
```diff
@@ -0,0 +1,22 @@
...
+const EmptyProgress = ({ handleOpen }: EmptyProgressProps) => {
+  const tProfile = useTranslations('Profile');
+  return (
+    <div className={targetBarStyle.EmptyProgressContainer}>
+      <div className={targetBarStyle.setTargetLabel}>
+        {tProfile('progressBar.targetInfo')}
+      </div>
+      <button className={targetBarStyle.setTargetButton} onClick={handleOpen}>
+        {tProfile('progressBar.setTargets')}
+      </button>
```
mohitb35, line 17: Replace this with WebappButton (so that we have consistent button styles)

e.g.
```
<WebappButton
      elementType="link"
      target="_blank"
      variant="tertiary"
      text={buttonText}
      href={donateLink}
      buttonClasses={buttonClasses}
    />

File: src/features/user/MFV2/ForestProgress/TreeTargetBar.module.scss [hunk 2]
```diff
@@ -0,0 +1,247 @@
...
+}
+
+.giftPercentageBar {
+  border-top-right-radius: 5px;
+  border-bottom-right-radius: 5px;
+}
+
+//target initialize  copntainer style
+
+.EmptyProgressContainer {
```
mohitb35, line 217: Please name classes consistently i.e. .emptyProgressContainer as done for the other classes.

src/features/common/Layout/MyForestContextV2.tsx
```diff
@@ -52,15 +52,35 @@ interface MyForestContextV2Interface {
...
+  restoreTarget: number;
+  setRestoreTarget: SetState<number>;
+  conservTarget: number;
+  setConservTarget: SetState<number>;
+  treeChecked: boolean;
+  setTreeChecked: SetState<boolean>;
+  restoreChecked: boolean;
+  setRestoreChecked: SetState<boolean>;
+  conservChecked: boolean;
+  setConservChecked: SetState<boolean>;
```
mohitb35, line 70: Reconsider using context for everything, and duplicating data within the context. It makes your code more complicated and prone to bugs.

For example, `treesPlanted`, `areaRestored` etc. are simple calculations and can be easily done where needed instead of duplicating the data contained in `stats`. Why not use `stats` directly where needed?

`treeChecked`, `restoreChecked` --> why do these need to be in the context?

src/features/common/Layout/MyForestContextV2.tsx
```diff
@@ -52,15 +52,35 @@ interface MyForestContextV2Interface {
...
   userInfo: UserInfo | null;
   setUserInfo: SetState<UserInfo | null>;
+  isLoading: boolean;
+  setIsLoading: SetState<boolean>;
```
mohitb35, line 74: We have 3 API calls here. What does `isLoading` signify?

src/features/user/MFV2/ForestProgress/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+    target: number,
+    checked: boolean,
+    treesDonated: number
+  ): boolean => {
+    if (treesDonated > 0 || (target > 0 && checked)) {
+      return true;
+    } else {
+      return false;
+    }
+  };
```
mohitb35, line 42: 1. `shouldShowBar` may be a more appropriate name here.
2. Why is the last argument called `treesDonated`, when that isn't applicable for `areaRestored` or `areaConserved`? `score` or `contributedUnits` may be more relevant.
3. You could truncate the complete `if` statement to `return (treesDonated > 0 || (target > 0)  && checked)`?

src/features/user/MFV2/ForestProgress/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+      areaConserved === 0,
+    [
+      treesPlanted,
+      treeTarget,
+      areaRestored,
+      restoreTarget,
+      conservTarget,
+      areaConserved,
+    ]
+  );
```
mohitb35, line 110: What condition am I checking? Why am I checking it?

src/features/user/MFV2/ForestProgress/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+          checked={conservChecked}
+        />
+      )}
+    </>
+  );
+};
+
+const ForestProgress = () => {
+  const [open, setOpen] = useState(false);
+  const [enableTarget, setEnableTarget] = useState(false);
```
mohitb35, line 82: In keeping with the names used here, `enableTarget` could be `shouldShowProgress` instead.

src/features/user/MFV2/ForestProgress/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+            <EmptyProgress handleOpen={handleOpen} />
+          ) : (
+            <ProgressBars
+              handleOpen={handleOpen}
+              treesDonated={treesPlanted}
+              areaRestored={areaRestored}
+              areaConserved={areaConserved}
+            />
+          )}
+
```
mohitb35, line 132: If the target is enabled, why would I show `EmptyProgress`? It appears that the logic for setting `enableTarget` is inverted.

src/features/user/MFV2/ForestProgress/microComponents/ForestProgressItem.tsx
```diff
@@ -0,0 +1,312 @@
...
+      case 'treesPlanted':
+        return primaryDarkColor;
+      case 'areaRestored':
+        return electricPurpleColor;
+      case 'areaConserved':
+        return mediumBlueColor;
+      default:
+        return '';
+    }
+  }, [dataType]);
```
mohitb35, line 53: I feel like using css for these styles would make things clearer. You could simply provide `dataType` as a class name.

mohitb35:
1. Trees planted progress bar -  a transparent white background is currently seen instead of the green background in the design
2. When I update a target and save, the targets and the bars do not update. They only update if I refresh the screen.
3. Translation strings need to be better named so that their purpose is clear for translators

src/features/common/Layout/MyForestContextV2.tsx
sunilsabatp: I have cleaned up the context by removing all the states that can be managed at the component level.

src/features/common/Layout/MyForestContextV2.tsx
sunilsabatp: I was using this state variable to manage loader which is actually not needed, I will remove it

src/features/common/Layout/UserPropsContext.tsx
```diff
@@ -116,6 +116,7 @@ export const UserPropsProvider: FC = ({ children }) => {
   React.useEffect(() => {
     if (token) {
       loadUser();
+      console.log('profileFetched');
     }
```
mohitb35, line 120: Remember to remove

src/theme/global.scss
```diff
@@ -747,3 +747,48 @@ input[type='number'] {
...
+  .personalPercentageBar {
+    background: $mediumBlueColor;
+  }
+  .giftPercentageBar {
+    background: $skyBlueColor;
+  }
+  .editTargetLabel {
+    color: $mediumBlueColor;
+  }
+}
```
mohitb35, line 794: Can these styles be moved to ForestProgress.module.scss?

src/features/common/Layout/MyForestContextV2.tsx
```diff
@@ -88,6 +102,14 @@ export const MyForestProviderV2: FC = ({ children }) => {
     slug: `${userInfo?.slug}`,
   });
 
+  useEffect(() => {
+    if (userInfo) {
+      setTreeTarget(userInfo.targets.treesDonated);
+      setRestoreTarget(userInfo.targets.areaRestored);
+      setConservTarget(userInfo.targets.areaConserved);
+    }
+  }, [userInfo]);
```
mohitb35, line 111: I'm not clear why we need to maintain the targets as separate variables instead of simply using `userInfo`?

We've already stored part of the profile data in the `userInfo` and now we're segregating the data further.

src/features/user/MFV2/ForestProgress/microComponents/EditButton.tsx
```diff
@@ -0,0 +1,26 @@
...
+>;
+
+const EditButton = ({ handleOpen, target, dataType }: EditButtonProp) => {
+  const tProfile = useTranslations('Profile.progressBar');
+
+  return (
+    <div className={styles.editTargetButtonContainer}>
+      <button className={styles.editTargetContainer} onClick={handleOpen}>
+        <EditTargetIcon width={9} color={targetColor(dataType)} />
+        <p className={`${styles.editTargetLabel} editTargetLabel`}>
```
mohitb35, line 19: Use modular styles instead of global.

Avoid mixing style sources.

src/features/user/MFV2/ForestProgress/ForestProgressItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const { asPath } = useRouter();
+
+  const _getAchievedTarget = useMemo(
+    () => getAchievedTarget(target, gift, personal),
+    [target, gift, personal]
+  );
+
+  return (
+    <div className={`${styles.progressMainContainer} ${dataType}`}>
+      {asPath === '/en/profile/mfv2' && (
```
mohitb35, line 34: I'd suggest not using path comparisons for logic. Paths can be changed, and then we need to remember to update the logic everywhere.

Use the profile page type to determine whether to show the Edit buttons, similar to how it is passed for `CommunityContributions`, `MyContributions` etc. in `ProfileLayout` and `PublicProfileLayout`

src/features/user/MFV2/ForestProgress/microComponents/PersonalPercentage.tsx
```diff
@@ -0,0 +1,28 @@
...
+    <div
+      style={{
+        width: `${personalPercentage}%`,
+        borderTopRightRadius: `${
+          giftPercentage !== 0 || target > totalAchievment ? 0 : 5
+        }px`,
+        borderBottomRightRadius: `${
+          giftPercentage !== 0 || target > totalAchievment ? 0 : 5
+        }px`,
+      }}
```
mohitb35, line 22: You don't really need to programatically provide a border radius for `PersonalPercentage` and `GiftPercentage`. Having border-radius on the parent `barContainer` div with`overflow: hidden` will take care of the border radius.

src/features/user/MFV2/ForestProgress/microComponents/GiftPercentageBar.tsx
```diff
@@ -0,0 +1,30 @@
...
+        width: `${giftPercentage}%`,
+        borderTopRightRadius: `${
+          target > gift && totalAchievment < target ? 0 : 5
+        }px`,
+        borderBottomRightRadius: `${
+          target > gift && totalAchievment < target ? 0 : 5
+        }px`,
+        borderTopLeftRadius: `${personalPercentage === 0 ? 5 : 0}px`,
+        borderBottomLeftRadius: `${personalPercentage === 0 ? 5 : 0}px`,
+      }}
```
mohitb35, line 24: You don't really need to programatically provide a border radius for `PersonalPercentage` and `GiftPercentage`. Having border-radius on the parent `barContainer` div with`overflow: hidden` will take care of the border radius.

src/features/user/MFV2/ForestProgress/microComponents/GiftPercentageBar.tsx
```diff
@@ -0,0 +1,30 @@
...
+      className={`${styles.giftPercentageBar} giftPercentageBar`}
+    ></div>
```
mohitb35, line 26: Avoid mixing global and modular styles.

src/features/user/MFV2/ForestProgress/microComponents/PersonalPercentage.tsx
```diff
@@ -0,0 +1,28 @@
...
+      className={`${styles.personalPercentageBar} personalPercentageBar`}
```
mohitb35, line 23: Avoid mixing global and modular styles.

src/features/user/MFV2/ForestProgress/microComponents/ProgressData.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  const giftReceivedProps = {
+    label,
+    gift,
+  };
+
+  return (
+    <div className={styles.progressContainer}>
+      <div className={styles.statisticsMainContainer}>
+        <div className={`${styles.iconContainer} iconContainer`}>
```
mohitb35, line 122: Avoid mixing global styles with modular styles

src/features/user/MFV2/ForestProgress/microComponents/ProgressData.tsx
```diff
@@ -0,0 +1,135 @@
...
+}) => {
+  return (
+    <>
+      {' '}
+      {gift !== undefined && gift > 0 && (
+        <div className={styles.communityReceived}>{label}</div>
+      )}
+    </>
+  );
+};
```
mohitb35, line 47: I feel like this abstraction was unnecessary. You could simply include this logic within `ProgressData` itself.

src/features/user/MFV2/ForestProgress/ForestProgressItem.tsx
```diff
@@ -0,0 +1,56 @@
+import styles from './ForestProgress.module.scss';
+import { getAchievedTarget } from '../../../../utils/myForestV2Utils';
+import { useMemo } from 'react';
+import { useRouter } from 'next/router';
+import EditButton from './microComponents/EditButton';
+import ProgressData from './microComponents/ProgressData';
+
+export type DataType = 'treesPlanted' | 'areaRestored' | 'areaConserved';
+
```
mohitb35, line 9: For clarity, name the type `ProgressDataType`.

src/features/user/MFV2/ForestProgress/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+      (contributionsResult?.stats.treesDonated.personal ?? 0) +
+      (contributionsResult?.stats.treesDonated.received ?? 0) +
+      (contributionsResult?.stats.treesRegistered ?? 0);
+    const areaRestored =
+      (contributionsResult?.stats.areaRestoredInM2.personal ?? 0) +
+      (contributionsResult?.stats.areaRestoredInM2.received ?? 0);
+    const areaConserved =
+      (contributionsResult?.stats.areaConservedInM2.personal ?? 0) +
+      (contributionsResult?.stats.areaConservedInM2.received ?? 0);
+    return { treesDonated, areaRestored, areaConserved };
```
mohitb35, line 79: I suggest separating the `stats` object in MyForestContextV2 and just using that here. It will make the code easier to read. 

It also makes sense to separate the main parts of the `ContributionsResponse`

src/features/user/MFV2/ForestProgress/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+  };
+
+  const contributions = useMemo(aggregateContributions, [contributionsResult]);
+  const { treesDonated, areaRestored, areaConserved } = contributions;
+
```
mohitb35, line 84: This destructuring operation could be moved into the useEffect below, which reduces the number of dependencies for the useEffect.

src/features/user/MFV2/ForestProgress/microComponents/TargetModalIconLabel.tsx
```diff
@@ -0,0 +1,50 @@
...
+      return <TreesPlantedIcon width={16} />;
+    case 'areaRestored':
+      return <AreaRestoredIcon width={14} />;
+    case 'areaConserved':
+      return <ConservedAreaIcon width={12} />;
+  }
+};
+
+const TargetModalIconLabel = ({ dataType }: TargetModalIconLabel) => {
+  const tProfile = useTranslations('Profile.progressBar');
```
mohitb35, line 26: `TargetFormInputLabel` would be a clearer name.

src/features/user/MFV2/ForestProgress/TargetsModal.tsx
```diff
@@ -0,0 +1,150 @@
...
+            setCheck={setIsRestoredAreaTargetActive}
+          />
+          <TargetFormInput
+            dataType={'areaConserved'}
+            target={conservTarget}
+            latestTarget={areaConservedTarget}
+            setLatestTarget={setAreaConservedTarget}
+            check={isConservedAreaTargetActive}
+            setCheck={setIsConservedAreaTargetActive}
+          />
```
mohitb35, line 136: There is some issue with the logic to turn enable/disable to switch.

src/features/user/MFV2/ForestProgress/TargetFormInput.tsx
```diff
@@ -0,0 +1,79 @@
...
+import TargetModalIconLabel from './microComponents/TargetModalIconLabel';
+import { DataType } from './ForestProgressItem';
+
+type TargetFormInputProps = {
+  dataType: DataType;
+  target: number;
+  latestTarget: number;
+  setLatestTarget: SetState<number>;
+  check: boolean;
+  setCheck: SetState<boolean>;
```
mohitb35, line 17: `enabled` or `checked` would be more appropriate - `check` is an action

src/features/user/MFV2/ForestProgress/TargetsModal.tsx
mohitb35: Still appears to be an issue

mohitb35:
@sunilsabatp Logic to enable switches for the target modal form has some slight issues, everything else is good.

mohitb35 (approved):
LGTM

PR #2079, opened 2024-05-29
Minor Changes for Bulk Codes and Payment History

mohitb35:
Changes:
1. Removes projects with `unitType = m2` from permitted projects to generate bulk codes
2. Enables donor certificate download for projects with `purpose = conservation` and projects with `purpose = tree, unitType = m2`


mariahosfeld (commented):
Functionality looks good.

PR #2084, opened 2024-06-03
Bug Fixes

shyambhongle:
Bug Fixes on Map

This PR addresses commonly reproducible bugs, including the following:

Zoom-in issues after selecting a project.
Delays or unexpected behavior when zooming/viewing the map after selecting a polygon.
Data not changing on side panel when hovering over different areas.
Data related to sample trees not updating  on side panel when clicking on specific sample trees.
Issue with the ImageSlider where images were not updating on data change.

Re: When a user clicks on a project after searching for it in the side panel, it takes time to zoom in to the project.
Cause: This delay occurs because the pages are rendered on the server, causing a delay only when the cache is expired. Once the cache is present for the page, the CTA behaves as expected.

### DIFF
--- a/src/features/projects/components/maps/Project.tsx
+++ b/src/features/projects/components/maps/Project.tsx
@@ -94,17 +94,6 @@ export default function Project({
   React.useEffect(() => {
     if (project.sites && siteExists && !router.query.ploc && !selectedPl) {
       loadRasterData();
-      zoomToProjectSite(
-        {
-          type: 'FeatureCollection',
-          features: project.sites,
-        },
-        selectedSite,
-        viewport,
-        setViewPort,
-        setSiteViewPort,
-        3000
-      );
     } else if (selectedPl) {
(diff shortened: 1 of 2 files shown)


mariahosfeld (changes requested):
- When loading for the first time, markers on the map are not aligned correctly (see screenshot), reload solves the issue:
image
- When switching between sites, zooming sometimes get stuck (sides are near one another, e.g., for the Yucatan project)
- In incognito window zooming still was stuck when loading a project (one reload resolved the issue)
- When loading another project, I saw shortly a white screen before it zoomed in correctly, but project details were not visible (using the browser back button made them appear), no pattern recognizable
- Wobbling of the base layer also still occurs sometimes (no pattern recognizable)
How to reproduce: Select different projects from the top project section in no particular order

mariahosfeld:
While testing directly on my iPhone, zooming does not correctly work when opening a project and after some tries I am experiencing an error, where I just see the empty map, no project details.

PR #2085, opened 2024-06-10
Fix: Polygon Zoom Issues

shyambhongle:
This PR addresses all zoom and hover-related issues with polygons. The following issues have been fixed:

1. When the user hovered over any polygons, the data did not update.
2. When a user select any polygon, the zoom was not working as expected. It either got stuck or was delayed.
3. When a user selected any site, it did not respond.

4. The PR also has a fix for ImageSlider crash
steps to reproduce:
    1. select a plantlocation
    2. select sample Trees
    3. Hover over any other polygon that don't have sample trees(app crash)
    or
    3. Select other polygon that don't have sample trees(app crash)

5. Fixes crash while visiting a project with less sites after visiting a project with more sites and selecting one of the last sites - this was due to the selected site index not being reset, resulting in an index out of bound error when visiting the new project. Also takes care of projects with no sites while fixing this.

### DIFF
--- a/src/features/projects/components/PlantLocation/ImageSlider.tsx
+++ b/src/features/projects/components/PlantLocation/ImageSlider.tsx
@@ -29,12 +29,6 @@ export default function ImageSlider({
     return ImageSource;
   };
 
-  React.useEffect(() => {
-    if (images.length > 0) {
-      setupSlider()
-    }
-  }, [images]);
-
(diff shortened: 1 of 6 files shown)


mohitb35 (approved):
LGTM

mohitb35 (approved):
LGTM

PR #2086, opened 2024-06-10
Map Crash fixes - Reference (DO NOT MERGE)

shyambhongle:
This PR addresses map crash issues. Below are the issues and steps to reproduce them on the production app:

1.When a user selects a project with fewer sites than the previously selected site, the map crashes because the current state index exceeds the number of sites in the new project.

Steps I followed to reproduce the issue:
1.Open Browser incognito(Mac OS. Chrome)
2.Clicked on "Yucatan" project.
3.Selected site with higher index. eg- "Inifap site B"
4.Go back from Project details UI.
5. Selected a project with less number of site.(Eg. Saving The Amazon(Columbia)"
6. Although the map zoom in everything on the details parts is gone.(You see a White Screen before zoom-in) 

2.Image slider issue: If a user selects a polygon with images and then hovers over a polygon without images, the app crashes. According to the fix for the issue I found on their Github, content should be set to null if there are no images, not an empty array. The code has been updated accordingly.
Issue Link- https://github.com/mohitk05/react-insta-stories/issues/213#issuecomment-1359356779
Plant location which don't have sampleTrees: en/yucatan?ploc=04DR23 (dev env)

3.When someone selects sites from the dropdown, the app sometimes doesn't respond. There was a redundant query parameter condition in zoomToSite() that is already handled in "Sites.tsx", so I removed the unnecessary condition.
This issue is not very common.  I removed the query condition as there was already a effect that handles siteChange condition.

Note:
Fixes 1 and 2 are now present in #2085, remaining changes can be left here as a reference but should not be merged in. The map logic is complicated - it may be more effective to redo the logic, and update the libraries.

### DIFF
--- a/src/features/projects/components/ProjectSnippet.tsx
+++ b/src/features/projects/components/ProjectSnippet.tsx
@@ -55,8 +55,8 @@ export default function ProjectSnippet({
     ? getImageUrl('project', 'medium', project.image)
     : '';
 
-  const { selectedPl, hoveredPl } = useProjectProps();
-  const { tenantConfig } = useTenant();
+    const { selectedPl, hoveredPl, setSelectedSite } = useProjectProps();
+    const { tenantConfig } = useTenant();
 
   let progressPercentage = 0;
 
@@ -110,6 +110,7 @@ export default function ProjectSnippet({
       ) : null}
       <div
         onClick={() => {
+          setSelectedSite(0)
           router.push(
             `/${locale}/${project.slug}/${
               embed === 'true'
--- a/src/features/projects/components/PlantLocation/ImageSlider.tsx
+++ b/src/features/projects/components/PlantLocation/ImageSlider.tsx
@@ -66,7 +66,7 @@ export default function ImageSlider({
 
   return <Stories
     stories={slider}
-    defaultInterval={300}
+    defaultInterval={7000}
     width="100%"
     height={height}
     loop={true}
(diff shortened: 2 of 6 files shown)


src/features/projects/components/ProjectSnippet.tsx
```diff
@@ -110,6 +110,7 @@ export default function ProjectSnippet({
       ) : null}
       <div
         onClick={() => {
+          setSelectedSite(0)
```
mohitb35, line 113: There are some projects for which there are zero sites. The app crashes when such a project is selected. 

Relevant part of the error: 

```
Uncaught TypeError: Cannot read properties of undefined (reading 'type')
    at S (turf.min.js:1:5604)
    at Object.j [as bbox] (turf.min.js:1:9286)
    at zoomToProjectSite (zoomToProjectSite.ts:31:18)
```

e.g. eden-reforestation-projects on staging

mohitb35: When the project snippet is visible, we do not always have access to the sites data as that is part of a different API call.

mohitb35:
I'm not sure of the exact issue here, but after visiting Yucatan, selecting a site lower in the list, coming back to the home page, selecting Saving the Amazon, the base layer of the map went blank and did not lead any more until I refreshed the page.

This was seen while testing on localhost, in incognito mode.

I did see some console warnings, possible relevant ones are below: 

after going back to the project list page.
`client.js:1 Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
`

`WARNING: Too many active WebGL contexts. Oldest context will be lost.`
 
 I was able to reproduce the error by visiting several projects and coming back to the home page in between.

PR #2091, opened 2024-06-17
Disable donations when not allowed for the project

mohitb35:
Does not display the donation button if the project does not allow donations.


mariahosfeld (approved):
LGTM

PR #2097, opened 2024-06-21
Feature/mfv2 layout tuning

mohitb35:
Changes:
1. Move CommunityContributions above MyContributions in small screen layout
2. Allows more space for the map to avoid variability based on profile card height
3. Adjusts fonts/spacing for consistency
4. Restricts project name to 2 lines in mobile
5. Restricts latest contributions to 2 items per project
6. Refactor to use common `profilePageType` prop instead of mixing `profileType`, `pageType` and `profilePageType` for the same purpose
7. Handle empty state for my contributions
8. Adapt public profile layout for impersonation
9. Standardizes loading UX - removes borders while showing the loaders.


prachigarg19:
In private profile, for mobile design, the header text needs to be centered
image

src/features/user/MFV2/ProfileLayout/ProfileLayout.module.scss
```diff
@@ -77,8 +77,11 @@
       border-radius: 24px;
     }
 
+    .mapContainer {
```
prachigarg19, line 98: Is there a reason why we've fixed the height for map container? Also, due to this the height of profile card will automatically be fixed to `450px`, would that not cause an issue for profile card with larger descriptions or have we set a character limit for the same?

src/features/user/MFV2/CommunityContributions/communityContributions.module.scss
```diff
@@ -141,15 +136,17 @@
   display: flex;
   flex-direction: column;
   gap: 12px;
-  font-size: 0.75em;
+  font-size: $fontSmall;
```
prachigarg19, line 139: Is the size deliberately increased here since according to `0.75em` it should be `$fontXSmall`

src/features/user/MFV2/ProfileLayout/ProfileLayout.module.scss
mariahosfeld: The height is fixed as the map will otherwise look squished if there is no description in the profile. The description itself is limited to 300 characters, so that is not creating an issue.

src/features/user/MFV2/CommunityContributions/communityContributions.module.scss
mohitb35: Yes, tried to make font sizes consistent across the screen, so I used the predefined sizes.

mohitb35:
> @mohitb35 In public profile, for mobile design, the header text needs to be centered (both vertically and horizontally) image
> 
> 2. we could further add a flex wrap in the button container for public profile to wrap `my wesbite` and `share` button
> 
> image

Point 1 > Was already done, but I'll align the text to the center of the containing div as well in this case
Point 2 > Noted, unlikely to be an issue as this is only seen on mobile screens < 330px, which is not very prevalent any more. But will add this as a point to take care of along with other feedback that emerges.

src/features/user/MFV2/CommunityContributions/communityContributions.module.scss
prachigarg19: `fontSmall` is `14px` though

prachigarg19 (approved):

PR #2104, opened 2024-06-26
Configures Donate buttons on public profile for direct gifts (supporting the profile)

mohitb35:
Donate buttons on the public profile page are now configured to add a `s = :slug` in the donation link, in order for donations to support the relevant public profile.


mariahosfeld (approved):
LGTM!

PR #2105, opened 2024-06-26
Handle no leaderboard layout on public profile

mohitb35:
This handles the public profile layout when no leaderboard should be shown.

Changes:
1. When no leaderboard is shown, the My Contributions section takes the full width of the screen
2. When no leaderboard is shown, and space permits, the MyContributionsList can contain 2 columns (subject to a minimum card width of 450 px).

For testing, the leaderboard will always be hidden on the public profile.

### DIFF
--- a/src/features/user/MFV2/MyContributions/MyContributions.module.scss
+++ b/src/features/user/MFV2/MyContributions/MyContributions.module.scss
@@ -7,8 +7,8 @@
 
 .myContributionsList {
   margin: 16px 12px 16px;
-  display: flex;
-  flex-direction: column;
+  display: grid;
+  grid-template-columns: 1fr;
   gap: 16px;
 
   @include smTabletView {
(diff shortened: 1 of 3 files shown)


mariahosfeld (commented):
LGTM, tested on desktop and on mobile, works as expected.

src/features/user/MFV2/MyContributions/MyContributions.module.scss
```diff
@@ -308,10 +312,14 @@
 .registrationItemCard {
   display: flex;
   flex-direction: column;
+  justify-content: center;
   background-color: #ffffff;
   border: 3px solid rgba(#007a49, 0.1);
   border-radius: 12px;
   overflow: hidden;
+  width: 100%;
+  min-width: 0; // Allow shrinking below 450px
```
prachigarg19, line 321: Why was this added? The registration cards seem to shrink in a similar way for width<450px without specifying `min-width:0`

prachigarg19 (approved):

src/features/user/MFV2/MyContributions/MyContributions.module.scss
mohitb35: These styles apply to the project card and registration cards. When this wasn't done, the cards stopped resizing as the container got narrower, and the content at the right got hidden.

PR #2106, opened 2024-06-26
add viewbox parameters for some icons to render them smaller in mfv2

norbertschuler:
Fixes oversized icons in new contributions map of my forest v2.

Changes in this pull request:
- added viewbox parameter where not present for all icons used in component `ProjectTypeIcon`

This fixed the icons in the above mentioned app. Problem is I do not know if these icons are also used anywhere else where the smaller size is not wanted, but they should be displayed in a larger form?

@


mariahosfeld:
@norbertschuler , thanks for taking care of this. Everything should work, as this is the file with the markers for this map, therefore it should be all.

mariahosfeld (approved):
LGTM!

PR #2110, opened 2024-06-27
Fix cluster outline color issue

prachigarg19:
Issue: Clusters don’t work correctly anymore. In the Cluster provided in the screenshot, a restoration project with unitType m2 is included. When originally testing, the correct cluster marker was shown (Green for main marker, right outline was purple), now all elements of the marker are green
image

Expected behavior:  To show different colored outlines if cluster contains markers of multiple colors
image


mariahosfeld (commented):

mohitb35 (approved):
LGTM.

@prachigarg19 I've added a return type to the `extractAndClassifyProjectData` function so that it's purpose is clearer.

PR #2112, opened 2024-06-27
Fix marker size

prachigarg19:
Fix the sizing of maker for `other-planting` and `agroforestry` category. 

Currently it looks like this where the marker is smaller as compared to other makers: 
image


mariahosfeld (commented):
Sizing looks good to me now

mariahosfeld:
I checked and I think that in general the sizing is correct, however sometimes it seems as some of the markers are slightly smaller than the others (might be related to the size of the contained icon, so an optical illusion). When potentially refactoring the file, please check for that.

mohitb35:
While adding the height allowed the icon to take the correct size, the root issue was that of a problem with the SVG code 
itself which resulted in an offset from the marker position.  

Exporting the SVGs again and updating the icons resolved this issue.

Screenshot 2024-06-28 at 10 47 38 AM
 
Screenshot 2024-06-28 at 10 47 43 AM

Screenshot 2024-06-28 at 10 47 30 AM

mohitb35 (approved):
LGTM

PR #2115, opened 2024-06-28
Resolves missing translation for `name` field label in edit profile

mohitb35:
Changes:

Resolves missing translation for `name` field label in edit profile


mariahosfeld (approved):
LGTM!

PR #2119, opened 2024-06-28
Fix logic for enabling/disabling Issue Codes button

mohitb35:
This pull request fixes the logic for enabling/disabling the Issue Codes button in the IssueCodesForm component. 

The logic was previously inverted i.e. the button was disabled when it should have been enabled, and vice versa.

In addition, a check on totalAmount has been added to ensure the button isn't enabled when the calculated total is 0.


mariahosfeld (approved):
LGTM!

PR #2124, opened 2024-07-02
fix: align popup properly by removing offset and anchor properties

sunilsabatp:
Remove offset and anchor properties to properly align popup

- Eliminated unnecessary offset and anchor properties
- Ensured popup aligns correctly without these adjustments
- Tested and verified alignment across different scenarios

### DIFF
--- a/src/features/user/MFV2/ContributionsMap/Markers/PointMarkers.tsx
+++ b/src/features/user/MFV2/ContributionsMap/Markers/PointMarkers.tsx
@@ -45,9 +45,13 @@ const PointMarkers = ({
     setIsCursorOnPopup,
   };
 
+  const handleMouseLeaveFromMarker = () => {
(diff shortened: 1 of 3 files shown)


mohitb35 (changes requested):
When I hover above a registered trees marker, the popup often flickers in and out of view. This is usually when my cursor is towards the bottom of the marker when the popup first appears. Could this be addressed?

Although this does not appear for the donation popup, you may want to use the same solution over there (for safety)

mohitb35 (approved):

mohitb35 (approved):
@sunilsabatp Reduced the timeout duration, LGTM now.

mariahosfeld:
Works for me, sometimes we still face the overlap, but that is possibly a trade off that we cannot avoid.

PR #2126, opened 2024-07-05
fix: switch locale through param in the embed mode

sunilsabatp:
- This pull request introduces functionality for conditional URL redirection based on parameters found in the URL (locale and embed).

### DIFF
--- a/middleware.ts
+++ b/middleware.ts
@@ -88,7 +88,18 @@ export default async function middleware(req: NextRequest) {
     (locale) => !pathname.startsWith(`/${locale}/`) && pathname !== `/${locale}`
   );
 
-  console.log('isLocaleMissing', isLocaleMissing);
+  const paramLocale = url.searchParams.get('locale');
+  const hasEmbedParam = url.searchParams.has('embed');
+  const pathLocale = url.pathname.split('/')[1];
+  const queryString = url.searchParams.toString();
+  //If the 'embed' parameter is present and the 'locale' parameter is set
+  //and different from the locale in the path, redirect to a new URL with the paramLocale
+  if (hasEmbedParam && paramLocale && paramLocale !== pathLocale) {
+    const newUrl = new URL(`/${paramLocale}?${queryString}`, req.url);
+    if (newUrl.href !== req.url) {
+      return NextResponse.redirect(newUrl);
+    }
+  }
 
   if (isLocaleMissing) {
     const locale = getLocale(req, commonSupportedLocales);


middleware.ts
```diff
@@ -88,7 +88,18 @@ export default async function middleware(req: NextRequest) {
     (locale) => !pathname.startsWith(`/${locale}/`) && pathname !== `/${locale}`
   );
 
-  console.log('isLocaleMissing', isLocaleMissing);
+  const paramLocale = url.searchParams.get('locale');
+  const hasEmbedParam = url.searchParams.has('embed');
+  const pathLocale = url.pathname.split('/')[1];
```
mohitb35, line 93: Locale may not be present in the path on the initial request (e.g. if the user visits web.plant-for-the-planet/profile)

middleware.ts
```diff
@@ -88,7 +88,18 @@ export default async function middleware(req: NextRequest) {
     (locale) => !pathname.startsWith(`/${locale}/`) && pathname !== `/${locale}`
   );
 
-  console.log('isLocaleMissing', isLocaleMissing);
+  const paramLocale = url.searchParams.get('locale');
+  const hasEmbedParam = url.searchParams.has('embed');
```
mohitb35, line 92: Why do we need to the check `embed` param?

middleware.ts
```diff
@@ -88,7 +88,18 @@ export default async function middleware(req: NextRequest) {
     (locale) => !pathname.startsWith(`/${locale}/`) && pathname !== `/${locale}`
   );
 
-  console.log('isLocaleMissing', isLocaleMissing);
```
mohitb35, line 91: Please add the console log back. This is in order to detect problems in the middleware.

middleware.ts
```diff
@@ -88,7 +88,18 @@ export default async function middleware(req: NextRequest) {
...
+  const pathLocale = url.pathname.split('/')[1];
+  const queryString = url.searchParams.toString();
+  //If the 'embed' parameter is present and the 'locale' parameter is set
+  //and different from the locale in the path, redirect to a new URL with the paramLocale
+  if (hasEmbedParam && paramLocale && paramLocale !== pathLocale) {
+    const newUrl = new URL(`/${paramLocale}?${queryString}`, req.url);
+    if (newUrl.href !== req.url) {
+      return NextResponse.redirect(newUrl);
+    }
+  }
```
mohitb35, line 102: Logic does not work. We lose the path that the user is trying to visit in this redirect.

mohitb35:
Closed as not needed any more

PR #2129, opened 2024-07-09
Fixes loss of embed state when project details page loads

mohitb35:
Preserves query parameters when project details page loads and the initial site is selected.

When a page e.g.`en/yucatan?embed=true` was visited directly, the embed parameters in the URL were lost before storing the state in QueryParamsContext. This fixes that issue.

Note that window.location.search is used instead of the router object, as the router object no longer contained the original query parameters after NextResponse.rewrite in the middleware.


mariahosfeld (commented):
Functionality looks good to me, checked with en and de as language parameters.

sunilsabatp (approved):

PR #2130, opened 2024-07-09
also sort top project listing not donable projects last

norbertschuler:
Fixes https://planetfoundation.slack.com/archives/C02393H3GDC/p1720455166668549

Changes in this pull request:
- also sort top project listing not donable projects last

PR #2051, opened 2024-04-24
BASE: My Forest V2

mohitb35:
This is the base PR for the feature `My forest - v2`

New pages (temporary links to be replaced after feature is ready and old pages are retired):
1. Private profile - /profile/mfv2
2. Public profile - /mfv2-public/[profile]

Updated pages
1. Edit Profile - /profile/edit 

Completed:
1. Basic structure #2053  
2. APIs #2064 
3. Project Locations #2070 
4. Profile card #2059 
5. Leaderboard API #2073
6. Map Layout #2058 
7. Public Profile Map #2074 
8. Amended Leaderboard API #2076 
9. Updated contributions API #2077
10. Updated contributions API #2080 
11. Redesign Edit Profile #2072 
12. Zoom and Marker Functionality #2060
13. My Contributions #2083 
14. Community Contributions Refactor #2087 
15. Community Contributions #2075  
16. Donation Stats #2067 
17. #2091 
18. #2078
19. #2068 
20. #2101 
21. #2102 
22. #2097 
23. #2104
24. #2106  
25. #2114 
26. #2066 
27. #2112 
28. #2110
29. #2105 
30. #2124 
31. #2121 
32. #2122 
33. #2123 
34. #2117 
35. #2125 
36. #2131 
37. #2132
38. #2128  

Todo

- [x] Empty state management (My Contributions and Overall)
- [x] Loading management (My Contributions and Overall)
- [ ] Minor background bleed issue for rounded outside borders while displaying gift label (My Contributions)
- [x] Responsiveness adjustments to be done in conjunction with other features on the page. e.g. https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/2083#issuecomment-2162067308

PR #2134, opened 2024-07-10
Release feature MyForest 2.0

mariahosfeld:
**Includes:**

- #2051 
- #2130 
- #2133 
- #2137 
- #2136
- #2138 
- #2139 
- #2140


public/static/locales/de/profile.json
```diff
@@ -33,6 +38,8 @@
       "editTarget": "Ziel bearbeiten"
     },
     "myContributions": {
+      "titlePrivate": "Meine Beiträge",
+      "titlePublic": "{name}'s Beiträge",
```
norbertschuler, line 42: Some people in Germany do not like using `'s` on names as in English as correctly mostly you just append an `s` to the name if possible (unfortunately it's more complicated for different kind of names, so it cannot simply be solved here except using some different wording as `Beiträge von {name}`), compare https://de.pons.com/p/wissensecke/grammatik-to-go/deppenapostroph.

mohitb35 (approved):
This can be released

PR #2144, opened 2024-07-16
Fix signup

norbertschuler:
Fixes https://planetfoundation.slack.com/archives/C02393H3GDC/p1721111290718719

Changes in this pull request:
- [x] fixed label names
- [x] fixed isPublic/isPrivate logic
- [ ] fix getNews parameter (@jmiridis `getNews` not stored correctly by the API endpoint at signup)


mohitb35 (approved):
@norbertschuler LGTM. I don't think the last todo will need any front end changes, so we can get it out as is.

PR #2146, opened 2024-07-16
hotfix: hide website button if url is missing

sunilsabatp:
Purpose: Hide the website button in the public profile if the user's URL is missing.

Changes:

Added a conditional check for userProfile?.url before rendering the `WebappButton.`
If userProfile?.url is not present, the `my website`  button will not be displayed.

### DIFF
--- a/src/features/user/MFV2/ProfileCard/ProfileActions/index.tsx
+++ b/src/features/user/MFV2/ProfileCard/ProfileActions/index.tsx
@@ -15,7 +15,7 @@ import WebappButton from '../../../../common/WebappButton';
 const ProfileActions = ({ profilePageType, userProfile }: ProfileV2Props) => {
   const [isRedeemModalOpen, setIsRedeemModalOpen] = useState(false);
   const t = useTranslations('Profile');
-
+  const userWebsiteUrl = userProfile?.url;
   const handleRedeemModalOpen = () => {
     setIsRedeemModalOpen(true);
   };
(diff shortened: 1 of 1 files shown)


src/features/user/MFV2/ProfileCard/ProfileActions/index.tsx
```diff
@@ -69,13 +69,15 @@ const ProfileActions = ({ profilePageType, userProfile }: ProfileV2Props) => {
...
       />
       <div className={styles.websiteShareActions}>
-        <WebappButton
-          icon={<WebsiteLinkIcon />}
-          text={t('feature.website')}
-          elementType={'link'}
-          href={handleWebsiteShareUrl()}
-          target={'_blank'}
-        />
+        {userWebsiteUrl && (
```
mohitb35, line 72: 1. Suggest directly checking for userProfile.url over here, rather than storing the value in a separate variable and then performing the check. The check condition is not complex, nor is it reused.
2. While conditionally rendering using &&, especially with user generated information, it is safer to make the check explicitly `true` or `false` i.e. `userProfile?.url !== undefined` is probably a safer option.

mohitb35 (changes requested):
@sunilsabatp

mariahosfeld (commented):
Functionality wise it looks good to me, please check the implementaion changes suggested by Mohit.

mohitb35 (approved):

PR #2150, opened 2024-07-17
Basic Structure for Redesigned Pages

mohitb35:
**Changes**

1. Set up temporary new pages
2. Set up basic structure for desktop
3. Set up basic structure for mobile with mode switching between list and map
4. ~~Update base map styles - left as is~~
5. Use maplibre instead of mapbox
6. Persistent map layout while switching between project details and list pages (preserve map state, enabling smooth zoom transitions between list and details pages)


sunilsabatp (approved):
LGTM

PR #2159, opened 2024-07-24
Fix profile contributions restriction and theme color

mohitb35:
1. Fixes a bug where profile contributions were not being restricted to purpose = trees/conservation. 
2. Also, replaces the renamed theme color `primaryDarkColorX` with `primaryColorNew` in multiple files.


mariahosfeld (approved):
Looks good to me.

PR #2162, opened 2024-07-29
Updates paymentOptions API call for Bulk Codes

mohitb35:
1. This pull request adds a new profile parameter to the paymentOptions API call to fetch a non cached response (for every different value of profile)
2. Authenticates the paymentOptions API calls as response returned is based on the authenticated user.


mariahosfeld (commented):
Functionality looks good.

sunilsabatp (approved):

PR #2188, opened 2024-08-20
Update planet-sdk and refactor code to get targets from user.scores

mohitb35:
Changes
1. Updates the planet-sdk to version 0.1.40, with updated user types.
2. Refactors the code to retrieve targets from the `user.scores` object instead of the deprecated `user.targets` object.


mariahosfeld (commented):

sunilsabatp (approved):
LGTM

PR #2197, opened 2024-08-23
changed appDomain for default tenant

norbertschuler:
Fixes #

Changes in this pull request:
- changed `appDomain` from `www1.` for default tenant to `web.`
- updated `defaultTenant` from API response of https://app.plant-for-the-planet.org/app/tenants/ten_NxJq55pm?_scope=deployment, but already set `appDomain` right and removed `auth0ClientId`
- adapted two returned values for `languages` and `font` as to apply to the type of `TenantConfig`

Open issue:
- the type of `header.items` of `TenantConfig` does not apply to the current values really used

@Shreyaschorge I am not sure of the effect  this has and wanted to ask why this value wasn't yet changed?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,7 @@
         "@mui/x-date-pickers": "^5.0.15",
         "@netlify/plugin-nextjs": "^4.33.0",
         "@next/bundle-analyzer": "^10.2.3",
-        "@planet-sdk/common": "^0.1.40",
+        "@planet-sdk/common": "^0.1.41",
         "@prisma/client": "^4.13.0",
         "@sentry/browser": "^6.15.0",
         "@sentry/integrations": "^6.19.2",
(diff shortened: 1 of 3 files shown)


mohitb35:
@norbertschuler The `defaultTenant` config would end up being used in `getStaticProps` for each page if we are not able to get tenant information via the tenants API (GET /app/tenants?_scope=deployment) for any reason.

The incorrect appDomain itself would not have affected anything, as it is not referenced anywhere. (The `appDomain` used in src/utils/multiTenancy/helpers.ts is from the API response).

Having said that, `defaultTenant` should be updated to match the response to `https://app.plant-for-the-planet.org/app/tenants/ten_NxJq55pm?_scope=deployment` - certain config elements like header will affect the app, and not updating this was an oversight.

mohitb35:
A couple of notes:

1. The API response currently has `"appDomain": "https://www1.plant-for-the-planet.org"` --> as `customDomain` is provided and takes precedence over `appDomain`, this does not affect anything. However, in the defaultTenant config, both `appDomain` and `customDomain` can be set as "https://web.plant-for-the-planet.org",
2. `auth0ClientId` does not need to be added to the `defaultTenant` config - in the absence of this, we fall back to the default `auth0ClientId` set in the env vars.

norbertschuler:
@mohitb35 I updated the data from the API response and applied the two changes you mentioned. So that would result into what you see now in this PR.

mohitb35 (approved):
@norbertschuler I updated the Tenant type in planet-sdk, and updated the default tenant config accordingly.

LGTM now.

PR #2199, opened 2024-08-26
Refactor useEffect logic to simplify impersonation mode check

sunilsabatp:
- Changed useEffect to directly set impersonation mode based on localStorage check.
- Removed redundant conditional check.


src/features/common/Layout/UserPropsContext.tsx
```diff
@@ -120,9 +120,8 @@ export const UserPropsProvider: FC = ({ children }) => {
   }, [token, refetchUserData]);
 
   React.useEffect(() => {
```
mohitb35, line 122: As `impersonationModeOn` is a dependency, and we are calling `setIsImpersonationModeOn` within the useEffect, unnecessary calls will be made to `setIsImpersonationModeOn` (at least 2 times while loading impersonation mode initially from local storage), and in the worst case we could end up in a loop.

Could we change this to:

React.useEffect(() => {
  const impersonationData = localStorage.getItem('impersonationData');
  if (impersonationData !== null && !isImpersonationModeOn) {
    setIsImpersonationModeOn(true);
  } else if (impersonationData === null && isImpersonationModeOn) {
    setIsImpersonationModeOn(false);
  }
}, []);

Would removing the dependency cause an issue?

mohitb35 (approved):
LGTM

src/features/common/Layout/UserPropsContext.tsx
mohitb35: outdated - closing this

PR #2223, opened 2024-09-11
Fixes unsupported currencies in country selector

mohitb35:
This pull request fixes an issue where countries with unsupported currencies were still being displayed in the country selector.


mariahosfeld (approved):
Looks good to me.

PR #2247, opened 2024-10-01
plant location image slider

sunilsabatp:
This PR addresses the modal view for the plant location image slider


mohitb35 (commented):
Looks good mostly. 

Just have one question, why do we pass imageSize = "large" for ImageSlider in PlantLocationInfo?

mohitb35 (commented):
Looks good mostly. 

Just have one question, why do we pass imageSize = "large" for ImageSlider in PlantLocationInfo?

>> This is because the "medium" size results in a 403 response while fetching images from the cdn

mohitb35 (approved):
LGTM

PR #2251, opened 2024-10-04
Adapt dataExplorer export for intervention_start_date

mohitb35:
This pull request includes changes to adapt the dataExplorer export for the new field `intervention_start_date`. The `plantDate` field has been deprecated in the `PlantLocationBase` type and replaced with `interventionStartDate`. The export data interface has also been updated to reflect this change. Additionally, various components and functions have been modified to handle the new field.


mohitb35:
This can be closed. Can revisit if we plan to update data explorer in the webapp in the future.

PR #2255, opened 2024-10-09
Zoom into sample plant location

sunilsabatp:
This PR handles:

1. Zooming into the sample plant location.
2. Displaying the whole plot functionality (present in the sample tree info sidebar).

### DIFF
--- a/src/features/projectsV2/ProjectDetails/components/ProjectInfo.tsx
+++ b/src/features/projectsV2/ProjectDetails/components/ProjectInfo.tsx
@@ -16,7 +16,7 @@ import ImageSlider from './microComponents/ImageSlider';
 import { ExtendedProject } from '../../../common/types/projectv2';
 
 interface ProjectInfoProps {
-  project: ExtendedProject;
+  project: ExtendedProject | null;
   isMobile: boolean;
   setSelectedMode: SetState<ViewMode> | undefined;
 }
@@ -26,6 +26,7 @@ const ProjectInfo = ({
   isMobile,
   setSelectedMode,
 }: ProjectInfoProps) => {
+  if (!project) return <></>;
   const tCountry = useTranslations('Country');
   const {
     metadata,
--- a/src/features/projectsV2/ProjectDetails/components/ProjectInfo.tsx
+++ b/src/features/projectsV2/ProjectDetails/components/ProjectInfo.tsx
@@ -16,7 +16,7 @@ import ImageSlider from './microComponents/ImageSlider';
 import { ExtendedProject } from '../../../common/types/projectv2';
 
 interface ProjectInfoProps {
-  project: ExtendedProject | null;
+  project: ExtendedProject;
   isMobile: boolean;
   setSelectedMode: SetState<ViewMode> | undefined;
 }
@@ -26,7 +26,6 @@ const ProjectInfo = ({
   isMobile,
   setSelectedMode,
 }: ProjectInfoProps) => {
-  if (!project) return <></>;
   const tCountry = useTranslations('Country');
   const {
     metadata,
--- a/src/features/projectsV2/ProjectDetails/components/microComponents/SampleTrees.tsx
+++ b/src/features/projectsV2/ProjectDetails/components/microComponents/SampleTrees.tsx
@@ -1,23 +1,38 @@
+import { useTranslations } from 'next-intl';
+import { SetState } from '../../../../common/types/common';
 import { SamplePlantLocation } from '../../../../common/types/plantLocation';
 import styles from '../../styles/PlantLocationInfo.module.scss';
-import { useTranslations } from 'next-intl';
 
 interface Props {
   sampleInterventions: SamplePlantLocation[];
+  setSelectedSamplePlantLocation: SetState<SamplePlantLocation | null>;
 }
-const SampleTreeList = ({ sampleInterventions }: Props) => {
+const SampleTreeList = ({
+  sampleInterventions,
+  setSelectedSamplePlantLocation,
+}: Props) => {
   const tProjectDetails = useTranslations('ProjectDetails');
+  const selectSampleSpecies = (speciesName: string) => {
+    if (!speciesName) return;
+    const result = sampleInterventions.find(
+      (sampleTree) => speciesName === sampleTree.scientificName
+    );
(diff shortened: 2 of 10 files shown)


mariahosfeld (changes requested):
When a sample location is selected, it should always show the location dot. This is currently only happening, when the parent location was clicked upon, not just highlighted via hover and the sample location selection happens via the infos on the left.

mohitb35:
Do we need to actually zoom into sample plant locations?

src/features/projectsV2/ProjectDetails/components/microComponents/SampleTrees.tsx
```diff
@@ -1,23 +1,38 @@
...
+    if (result) setSelectedSamplePlantLocation(result);
+  };
   return (
     <div className={styles.sampleSpeciesListContainer}>
       {sampleInterventions.map((sampleSpecies, index) => {
         return (
           <div key={sampleSpecies.id} className={styles.sampleSpeciesContainer}>
-            <div className={styles.scientificNameContainer}>
+            <button
+              className={styles.scientificNameContainer}
```
mohitb35, line 28: font color has changed due to button styles getting applied. Correct this.

src/features/projectsV2/ProjectDetails/components/microComponents/SampleTrees.tsx
```diff
@@ -1,23 +1,38 @@
...
+  setSelectedSamplePlantLocation,
+}: Props) => {
   const tProjectDetails = useTranslations('ProjectDetails');
+  const selectSampleSpecies = (speciesName: string) => {
+    if (!speciesName) return;
+    const result = sampleInterventions.find(
+      (sampleTree) => speciesName === sampleTree.scientificName
+    );
+    if (result) setSelectedSamplePlantLocation(result);
+  };
```
mohitb35, line 21: I don't get why we're trying to select a sample species - are we not selecting a samplePlantLocation?

src/features/projectsV2/ProjectDetails/components/ProjectInfo.tsx
```diff
@@ -16,7 +16,7 @@ import ImageSlider from './microComponents/ImageSlider';
 import { ExtendedProject } from '../../../common/types/projectv2';
 
 interface ProjectInfoProps {
-  project: ExtendedProject;
+  project: ExtendedProject | null;
   isMobile: boolean;
```
mohitb35, line 20: I don't think `null` should be accepted as a possible value for project. It allows us to skip the check while using `ProjectInfo` which is not advisable.

src/features/projectsV2/ProjectDetails/index.tsx
```diff
@@ -143,6 +143,24 @@ const ProjectDetails = ({
...
+      : selectedPlantLocation,
+    setSelectedSamplePlantLocation,
+    isMobile,
+  };
+
+  const projectInfoProps = {
+    project: singleProject,
+    isMobile,
+    setSelectedMode,
+  };
```
mohitb35, line 163: This wasn't needed IMO, the prop list isn't too lengthy for any of the components, so this actually makes the code less readable.

src/features/projectsV2/ProjectsMap/SingleProjectView.tsx
```diff
@@ -36,12 +41,15 @@ const SingleProjectView = ({ mapRef, setIsOnSampleMarker }: Props) => {
...
-  // Zoom to plant location polygon
+  // Zoom to plant location
   useEffect(() => {
-    const isPlantLocationReadyToZoom =
-      selectedPlantLocation && router.isReady && requestedPlantLocation;
+    const canZoomToPolygonPlantLocation =
+      selectedPlantLocation &&
+      router.isReady &&
+      requestedPlantLocation &&
+      !selectedSamplePlantLocation;
```
mohitb35, line 50: Make it a point to ensure that a variable with a boolean name contains a boolean value, instead of depending on implicit truthyness/falsyness.

src/features/projectsV2/ProjectsMap/SingleProjectView.tsx
```diff
@@ -36,12 +41,15 @@ const SingleProjectView = ({ mapRef, setIsOnSampleMarker }: Props) => {
...
     if (
-      isPlantLocationReadyToZoom &&
+      canZoomToPolygonPlantLocation &&
       selectedPlantLocation.geometry.type === 'Polygon'
     ) {
```
mohitb35, line 54: `selectedPlantLocation.geometry.type === 'Polygon'`

could be part of the `canZoomToPolygonPlantLocation` logic itself.

src/features/projectsV2/ProjectsMap/SingleProjectView.tsx
```diff
@@ -51,13 +59,33 @@ const SingleProjectView = ({ mapRef, setIsOnSampleMarker }: Props) => {
...
+          zoomToLocation(setViewState, lon, lat, 20, 3500, mapRef);
+        }
+      }
     }
-  }, [selectedPlantLocation, requestedPlantLocation, router.isReady]);
+  }, [
+    selectedPlantLocation,
+    selectedSamplePlantLocation,
+    requestedPlantLocation,
+    router.isReady,
```
mohitb35, line 82: What are the conditions under which we'd zoom to neither a polygon nor a point location? Handling those earlier will make it easier for typescript to exclude non-possible types.

mohitb35 (changes requested):
@sunilsabatp have left feedback

src/features/projectsV2/ProjectDetails/components/microComponents/SampleTrees.tsx
sunilsabatp: We should call it 'samplePlant' instead of 'samplePlantLocation' because we are displaying plant information (i.e., measurements, scientific name, etc.) rather than location information

sunilsabatp:
> Do we need to actually zoom into sample plant locations?

I have confirmed with @mariahosfeld that we don't need to zoom further to the sample plant location.

mariahosfeld (commented):

mohitb35 (approved):

PR #2260, opened 2024-10-14
fix: redirect user to the previous page

sunilsabatp:
This PR handles the back button functionality of project snippet
possible ways to navigate to the project details page 
- [x] - List Screen within the WebApp
- [x] - List of Projects within an RO profile
- [x] - When the RO views the project via the project management or via payment history
- [ ] - The status page on the Website (we need to check)
- [x] - Whereever the link is used, also from outside the WebApp/Website


mariahosfeld (changes requested):
- [ ] When just using a link, without being on the WebApp before, nothing happens when I click on the back button within the project details. If we do not have a callback URL we either need to send them to the main WebApp otherwise disable the back button

mariahosfeld (commented):
We need to check if we can improve the loading times when going to the home screen, as this is taking some time and sometimes you don't see the loader right away, but just stay on the page you have been on before, the button seemingly not reacting.

mariahosfeld (commented):
Functionality looks good to me, either to the previous page or to the list screen. Worked as expected in every test.

mariahosfeld (approved):

src/features/projectsV2/ProjectSnippet/microComponents/ImageSection.tsx
```diff
@@ -34,22 +36,26 @@ const ImageSection = (props: ImageSectionProps) => {
...
+    const previousPageRoute = localStorage.getItem('redirectLink');
+    const defaultRoute = `/${locale}/prd`;
+    const queryParams = {
+      ...(isEmbed ? { embed: 'true' } : {}),
+      ...(isEmbed && callbackUrl !== undefined
+        ? { callback: callbackUrl }
+        : {}),
+    };
+    if (previousPageRoute) {
+      router.push(previousPageRoute);
```
mohitb35, line 51: Why are query params not passed on in this condition?

src/features/projectsV2/ProjectsContext.tsx
```diff
@@ -406,6 +410,8 @@ export const ProjectsProvider: FC<ProjectsProviderProps> = ({
       setSelectedSamplePlantLocation,
       selectedSite,
       setSelectedSite,
+      preventShallowPush,
```
mohitb35, line 413: Since `preventShallowPush` is only accessed in the context, don't pass it on as a part of `value`

pages/sites/[slug]/[locale]/prd/index.tsx
```diff
@@ -26,6 +26,7 @@ const ProjectListPage: NextPageWithLayout = ({ pageProps, isMobile }) => {
   const { setTenantConfig } = useTenant();
 
   useEffect(() => {
+    localStorage.removeItem('redirectLink');
```
Comment (mohitb35) [inline pages/sites/[slug]/[locale]/prd/index.tsx:29]: Since `redirectLink` is used for the purpose of login, use another name here.

e.g. `backNavigationUrl`

You could also use sessionStorage instead of localStorage (as there is no redirect path).

src/features/user/ManageProjects/ProjectsContainer.tsx
```diff
@@ -127,6 +128,7 @@ export default function ProjectsContainer() {
   // This effect is used to get and update UserInfo if the isAuthenticated changes
   React.useEffect(() => {
     if (contextLoaded && token) {
+      localStorage.setItem('redirectLink', router.asPath);
       loadProjects();
```
mohitb35, line 132: I don't particularly like that we are setting a local storage value to detect where the user came to the project details page from.

There could be issues with this approach - perhaps the local storage value gets set when this page is initially visited, but through internal navigation in the app the user accesses another page which links to a project. We'll need to ensure the value is up to date whenever such a page is visited, otherwise the back navigation may take the user to an unexpected place.

Would it be more reliable to pass the `backNavigationUrl` as a query param to the project details page? That could simplify the logic, we'd only need to take the user to the `backNavigationUrl` if available, or to the the project list page (default).

sunilsabatp: The query param logic is the good idea

mohitb35: Sounds good, keep this query param specific to the project details page only.

mariahosfeld (changes requested):
When coming from a direct link and then using the back button, the zoom on the map component to show the whole map on the list screen is not working correctly. It stops after just zooming out a little bit.

mariahosfeld (commented):
Zooming functionality has been restored.

src/features/projectsV2/ProjectSnippet/index.tsx
```diff
@@ -146,20 +158,41 @@ export default function ProjectSnippet({
...
     }
+
+    const paramsString = params.toString();
+    if (paramsString) {
+      path += `?${paramsString}`;
+    }
+
     return path;
-  }, [locale, project.slug, embed, callbackUrl]);
+  }, [locale, project.slug, embed, callbackUrl, page]);
```
mohitb35, line 181: There's something wrong in the params logic.

If I visit `en/prd?embed=true`, the projectPath gets calculated as `/en/prd/:projectSlug?backNavigationUrl=/en/prd?embed=true?embed=true`.

You probably need to add url encoding to the backNavigationUrl, and also check the logic for adding params so that we have the pattern of `?param1=value1&param2=value2` not `?param1=value1?param2=value2`

src/features/user/Account/History.tsx
```diff
@@ -64,6 +63,7 @@ export default function History({
   };
 
   useEffect(() => {
+    localStorage.setItem('backNavigationUrl', router.asPath);
```
mohitb35, line 66: Was this left in by accident?

src/features/user/ManageProjects/ProjectsContainer.tsx
```diff
@@ -128,6 +132,7 @@ export default function ProjectsContainer() {
   React.useEffect(() => {
     if (contextLoaded && token) {
       loadProjects();
+      localStorage.setItem('backNavigationUrl', router.asPath);
     }
```
mohitb35, line 136: Was this left in by accident?

src/features/user/Account/components/AccountRecord.tsx
```diff
@@ -167,7 +168,10 @@ export function DetailsComponent({ record }: DetailProps): ReactElement {
         <div className={styles.singleDetail}>
           <p className={styles.title}>{tMe('project')}</p>
           {record.projectGuid ? (
-            <a title={record.details.project} href={`/${record.projectGuid}`}>
+            <a
+              title={record.details.project}
+              href={`/prd/${record.projectGuid}?backNavigationUrl=${router.asPath}`}
```
mohitb35, line 173: I feel we could create a util function to generate project path in a standard way for each place.

mohitb35 (changes requested):
@sunilsabatp Looks mostly good. There a few points for cleanup, and the params logic used to generate projectPath that needs to be addressed.

src/features/user/ManageProjects/ProjectsContainer.tsx
sunilsabatp: Yes

src/features/user/Account/History.tsx
sunilsabatp: Yes

mariahosfeld (commented):
Functionality looks good to me, embedding behavior needs to be reviewed in context of base PR.

mariahosfeld (approved):

src/features/projectsV2/ProjectSnippet/microComponents/ImageSection.tsx
```diff
@@ -34,22 +36,31 @@ const ImageSection = (props: ImageSectionProps) => {
...
+        pathname: routerPath,
+        query: queryParams,
+      })
+      .then(() => sessionStorage.removeItem('backNavigationUrl'))
+      .catch((error) => {
+        console.error('Navigation failed:', error);
       });
-    }
-  }, []);
+  };
```
mohitb35, line 63: A small issue with this logic. 

When an absolute URL is provided, the embed param / callbackUrl param gets added to the absolute URL after clicking back. So if I had added "https://www.google.com/" as the `backNavigationUrl`

mohitb35 (approved):
@sunilsabatp LGTM. Take a look at the comment, once that's addressed, go ahead and merge this in.

PR #2267, opened 2024-10-17
feat: expand the clickable area for project snippet

sunilsabatp:
- [ ]  Expand clickable area in project pop-up (on map) so that most of it is included and will lead to project

### DIFF
--- a/src/features/common/WebappButton/index.tsx
+++ b/src/features/common/WebappButton/index.tsx
@@ -35,6 +35,7 @@ function WebappButton({
         href={otherProps.href}
         target={otherProps.target || '_self'}
         className={styles.webappButtonLink}
+        onClick={(e) => e.stopPropagation()}
       >
         <button
           className={`${styles.webappButton} ${buttonVariantClasses} ${otherProps.buttonClasses}`}
--- a/src/features/common/WebappButton/index.tsx
+++ b/src/features/common/WebappButton/index.tsx
@@ -37,6 +37,7 @@ function WebappButton({
         className={styles.webappButtonLink}
       >
         <button
+          onClick={(e) => e.stopPropagation()}
           className={`${styles.webappButton} ${buttonVariantClasses} ${otherProps.buttonClasses}`}
         >
           {otherProps.icon !== undefined && (
--- a/src/features/common/WebappButton/index.tsx
+++ b/src/features/common/WebappButton/index.tsx
@@ -37,7 +37,6 @@ function WebappButton({
         className={styles.webappButtonLink}
       >
         <button
-          onClick={(e) => e.stopPropagation()}
           className={`${styles.webappButton} ${buttonVariantClasses} ${otherProps.buttonClasses}`}
         >
           {otherProps.icon !== undefined && (
--- a/src/features/projectsV2/ProjectSnippet/index.tsx
+++ b/src/features/projectsV2/ProjectSnippet/index.tsx
@@ -164,7 +164,11 @@ export default function ProjectSnippet({
         {page === 'project-details' ? (
           <ProjectSnippetContent {...ProjectSnippetContentProps} />
         ) : (
-          <Link href={getProjectPath()} style={{ cursor: 'pointer' }}>
+          <Link
+            href={getProjectPath()}
+            style={{ cursor: 'pointer' }}
+            onClick={(e) => e.stopPropagation()}
+          >
             <ProjectSnippetContent {...ProjectSnippetContentProps} />
           </Link>
         )}
--- a/src/features/projectsV2/ProjectSnippet/index.tsx
+++ b/src/features/projectsV2/ProjectSnippet/index.tsx
@@ -16,6 +16,8 @@ import ImageSection from './microComponents/ImageSection';
 import styles from './styles/ProjectSnippet.module.scss';
 import { getProjectCategory } from '../../../utils/projectV2';
 import TpoName from './microComponents/TpoName';
+import { useLocale } from 'next-intl';
+import { useRouter } from 'next/router';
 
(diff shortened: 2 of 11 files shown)


mariahosfeld (changes requested):
Generally the expansion of the clickable area looks fine, but once we are in the project details we should reduce the clickable area, as nothing will happen once we click there.

mohitb35:
Not related to this PR, and might have already been addressed:

When I opened /prd/trees-for-peace-an-ayudh-europe-campaign, I saw a client side error:

Screenshot 2024-10-18 at 10 14 50 AM

src/features/projectsV2/ProjectSnippet/index.tsx
```diff
@@ -119,17 +122,36 @@ export default function ProjectSnippet({
...
+        >
+          <ImageSection {...imageProps} />
+          <div className={styles.progressBar}>
+            <div
+              className={`${styles.progressBarHighlight} ${progressBarClass}`}
+              style={{ width: `${progressPercentage}%` }}
+            />
+          </div>
+          <ProjectInfoSection {...projectInfoProps} />
+        </button>
```
mohitb35, line 154: Can we make this a link instead of a button (with the same routing logic)?

src/features/projectsV2/ProjectSnippet/index.tsx
mohitb35, line 1: Because of the margin above the `TpoName` component, there is a small area of the project snippet which is not clickable. 


Screenshot 2024-10-18 at 10 23 05 AM

Can the layout be updated to make that area clickable as well? Perhaps by removing the margin above `TpoName` and adding bottom padding to the component above.

Test different places the ProjectSnippet is seen to ensure no inadvertent issues are caused by doing this.

mohitb35 (changes requested):
@sunilsabatp Minor feedback, please address.

mariahosfeld (changes requested):
General functionality looks good now.
But, compared to the base, the "empty" area between CtA and name of the RO is too big, please check.

mariahosfeld (changes requested):
Now on both desktop and mobile the space is correct. However, when clicking on the project snippet now, while viewing the project details, in both versions I get either a white screen (desktop) or a message about a clientside error (mobile).

mariahosfeld (commented):
Looks good to me, error is resolved.

mariahosfeld (approved):

src/features/common/WebappButton/index.tsx
```diff
@@ -37,6 +37,7 @@ function WebappButton({
         className={styles.webappButtonLink}
       >
         <button
+          onClick={(e) => e.stopPropagation()}
```
mohitb35, line 40: Move the `onClick` handler to the `Link` element, as we want to stop event propagation beyond the link. 

While the current code seems to work, stopping event propagation at the Link element makes the intention of the code clearer, and we ensure that no click events related to this component (whether they originate from the button or the link) will bubble up. This could prevent subtle bugs in more complex layouts where event bubbling might cause unexpected behavior

mohitb35:
> Not related to this PR, and might have already been addressed:
> 
> When I opened /prd/trees-for-peace-an-ayudh-europe-campaign, I saw a client side error:
> 
> Screenshot 2024-10-18 at 10 14 50 AM

@sunilsabatp 

> Not related to this PR, and might have already been addressed:
> 
> When I opened /prd/trees-for-peace-an-ayudh-europe-campaign, I saw a client side error:
> 
> Screenshot 2024-10-18 at 10 14 50 AM

While this is addressed, there is some doubt about how projects can have null site geometry, and how this is to be handled. @sunilsabatp Could you revert these specific changes, and keep them in a separate PR, which can be worked on/reviewed once the doubts are clarified?

src/features/projectsV2/ProjectSnippet/index.tsx
```diff
@@ -114,22 +112,63 @@ export default function ProjectSnippet({
...
-            style={{ width: `${progressPercentage}%` }}
-          />
-        </div>
-        <ProjectInfoSection {...projectInfoProps} />
+        {page === 'project-details' ? (
+          <ProjectSnippetContent {...ProjectSnippetContentProps} />
+        ) : (
+          <Link href={getProjectPath()} style={{ cursor: 'pointer' }}>
+            <ProjectSnippetContent {...ProjectSnippetContentProps} />
+          </Link>
```
mohitb35, line 169: Avoid calling functions from JSX, unless the function is an event handler. It is harder to debug, and could cause unexpected behavior. Here, you could store `projectPath` in a memoized variable and directly pass it to `href`

mohitb35 (changes requested):
@sunilsabatp Minor feedback, main thing is to revert the changes in 71bd0c253046de12430bd5d537b83e8595a0288f

mohitb35:
> e161501

Is this resolved?

sunilsabatp:
> > General functionality looks good now. But, compared to the base, the "empty" area between CtA and name of the RO is too big, please check. 
> 
> Is this resolved?

yes

src/features/common/WebappButton/index.tsx
mohitb35: You've removed the `onClick={(e) => e.stopPropagation()}`, but not added it to the `<Link>`. As a result, the Donate button does not work as intended any more, the user is taken to /prd as the click propagates to the parent container.

src/features/projectsV2/ProjectsMapContext.tsx
mohitb35, line 1: These changes were part of the `geometry` null check, and have not been reverted. Should they be?

PR #2273, opened 2024-10-23
Fix data explorer queries for single tree registrations

mohitb35:
- Fixes the data explorer queries to handle single tree registrations, and also makes tweaks to exclude deleted interventions, restrict data to single and multi tree registrations
- Updates the PlantLocationDetails component to handle single tree registrations, and also display a `hid` to identify the selected plant location


mariahosfeld (commented):
Functionality looks good to me, data makes sense and integration of HID looks good.

sunilsabatp (approved):
LGTM

PR #2274, opened 2024-10-24
Projects redesign/facelift feedback

sunilsabatp:
- [ ] handle error when map tries to show satellite layer when site layer is not there (no sites scenario) e.g. /prd/mothersforest(`browser devtool warning`)
- [x] Include localized separators for planting density


mariahosfeld (commented):
Unfortunately it seems to me, that the separators are not localized, but that could also be due to the language bug we need to fix. When trying to access de, it always switches back to en, so I cannot be sure.

mohitb35:
It seems like https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/2275/ and this PR address some common points.

mohitb35:
Localized separators work as expected. The translation is a separate issue.

mohitb35 (approved):
LGTM

PR #2275, opened 2024-10-25
fix: handle site geometry null case

sunilsabatp:
This PR addresses the case where a `project site geometry is null`, ensuring that the application handles this scenario gracefully without causing errors or unexpected behavior.

### DIFF
--- a/src/features/projectsV2/ProjectsContext.tsx
+++ b/src/features/projectsV2/ProjectsContext.tsx
@@ -269,13 +269,10 @@ export const ProjectsProvider: FC<ProjectsProviderProps> = ({
     projectSlug: string,
     siteIndex: number | null
   ) => {
-    if (!singleProject?.sites?.length) return;
-
+    if (!singleProject?.sites) return;
     setSelectedSite(siteIndex);
-
     const siteId =
       siteIndex !== null ? singleProject.sites[siteIndex]?.properties.id : null;
-
     updateUrlWithSiteId(locale, projectSlug, siteId);
   };
 
--- a/src/features/projectsV2/ProjectsContext.tsx
+++ b/src/features/projectsV2/ProjectsContext.tsx
@@ -237,7 +237,7 @@ export const ProjectsProvider: FC<ProjectsProviderProps> = ({
       setSelectedPlantLocation(null);
       setSingleProject(null);
       setHoveredPlantLocation(null);
-      setSelectedSite(0);
+      setSelectedSite(null);
       setPreventShallowPush(false);
     }
   }, [page]);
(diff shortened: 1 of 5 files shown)


mariahosfeld (changes requested):
For projects without sites I don't encounter any issues.
But when checking out a project with sites and treemapper data (/yucatan), I first see the vegetation layer and it takes a while (compared to the base) until the intended map with treemapper data shows up. Don't know if this is related, so please check and if needed amend.

mohitb35 (approved):
LGTM

mohitb35:
> For projects without sites I don't encounter any issues. But when checking out a project with sites and treemapper data (/yucatan), I first see the vegetation layer and it takes a while (compared to the base) until the intended map with treemapper data shows up. Don't know if this is related, so please check and if needed amend.

I see this issue too.

sunilsabatp:
> > For projects without sites I don't encounter any issues. But when checking out a project with sites and treemapper data (/yucatan), I first see the vegetation layer and it takes a while (compared to the base) until the intended map with treemapper data shows up. Don't know if this is related, so please check and if needed amend.
> 
> I see this issue too. May not show up on local, unless you disable `reactStrictMode`

I have resolve this issue in this PR [projects-redesign/facelift-feedback](https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/2274)

mariahosfeld (commented):
Now the sat layer is not shown when selecting /yucatan, but still a quite long latency for TreeMapper data to show. Maybe it is because of a first load of the project, but we need to keep an eye out and think about how we could optimize it.

src/features/projectsV2/ProjectsContext.tsx
```diff
@@ -271,7 +271,7 @@ export const ProjectsProvider: FC<ProjectsProviderProps> = ({
     projectSlug: string,
     siteIndex: number | null
   ) => {
-    if (!singleProject?.sites?.length) return;
+    if (!singleProject?.sites) return;
```
mohitb35, line 274: I'm not sure why this was changed, but the sites array could be empty. In that case, we should not be able to set a selectedSite or add the site query param in the url

mohitb35 (approved):
@sunilsabatp Minor feedback, can address that and merge

mohitb35 (approved):
LGTM

mohitb35:
@sunilsabatp 
1. While visiting `/prd/trees-for-peace-an-ayudh-europe-campaign`, the site with the null geometry remains in the url.
2. While visiting `/prd/mothersforest?site=asdfkjasfjsdnfjksdn`, the site param remains in the url even though the site is not found.

mariahosfeld (commented):

mohitb35 (approved):
LGTM

PR #2277, opened 2024-10-28
hotfix: avoid calling plantlocation api while switching between list …

sunilsabatp:
- [ ] I noticed that multiple calls are made to the plantLocations API in mobile devices (map mode) - This is getting called everytime map mode is exited


mohitb35 (changes requested):
@sunilsabatp This change breaks the existing functionality

mohitb35 (approved):
LGTM

PR #2278, opened 2024-10-29
Projects redesign/key info container

sunilsabatp:
- [ ] Evaluate styles/classes defined for KeyInfo / AdditionalInfo. For example, we have no class defined called infoDetail but this is used in AdditionalInfo, and KeyInfo has code which results in nested singleRowInfoContainer classes. Refactor, and rename where needed. (sep PR)


mohitb35 (approved):
LGTM

PR #2279, opened 2024-10-30
Projects redesign/minor UI fixes

mohitb35:
This pull request focuses on making minor UI fixes as part of the projects redesign.


mariahosfeld (changes requested):
- [x] change also color of "x" button to link color
image
- [ ] on mobile also add an opaque circle behind "x" to exit Dive into Project, when using Satellite view, it gets kind of lost. (comparable to image slider expansion on mobile)

mohitb35:
> * [x]  change also color of "x" button to link color
>   image
> * [x]  on mobile also add an opaque circle behind "x" to exit Dive into Project, when using Satellite view, it gets kind of lost. (comparable to image slider expansion on mobile)

@mariahosfeld Resolved, can be rechecked

mariahosfeld (approved):
Changes look good to me.

mariahosfeld (changes requested):
Just noticed: in the project list itself, I just see the green loader, not the new one, as it is already visible for the pop-up.

mohitb35:
> Just noticed: in the project list itself, I just see the green loader, not the new one, as it is already visible for the pop-up. At least for the top-projects this has to still be adapted.

There is no difference between the popup and the list. You may see a brief splash of the green background before the default image loads - leaving that background out resulted in a blank background getting replaced by the green default image, and that effect was more jarring than having a green background already in place.

mariahosfeld (approved):

sunilsabatp:
Could we shift the cross icon a little to the left to align it with the other icon (i.e., the satellite icon)?
image

mohitb35:
> Could we shift the cross icon a little to the left to align it with the other icon (i.e., the satellite icon)? image

Fixed

sunilsabatp (approved):

PR #2280, opened 2024-10-30
refactor: consistent type import

sunilsabatp:
- [x] import types explicitly (consistent-type-imports)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,7 +4,13 @@ module.exports = {
     browser: true,
     es2020: true,
   },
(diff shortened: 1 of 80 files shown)


mohitb35 (changes requested):
Overall things look good. 

1. For consistency, can we separate type imports from other imports, and group type imports together? Use the `import type { Type1, Type2 }` while importing only types, and not `import {type Type1, Component2}`. This helps to make it clearer when types are being imported, or runtime imports - making code review easier.

3. In places where we are not updating the type imports, please reverse the changes. For example, in `[src/features/projectsV2/ProjectDetails/components/AboutProject.tsx](https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/2280/files#diff-e2fd2fa620bea11ebc28cd0ca94ad7758147360ae243fd4b2435c6c8b0cf448c) the imports are just reordered.

mohitb35 (approved):
Made minor refactors. LGTM.

PR #2284, opened 2024-11-05
Cleaned up the temp folder

sunilsabatp:
https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/2149#discussion_r1820475143


mohitb35 (changes requested):
The `InfoIconPopup` can also be moved out of the temp folder, as it used in `KeyInfo`. Otherwise this is good.

mohitb35 (approved):

PR #2287, opened 2024-11-05
Project Details Navigation Improvements + Translation Issue Fix

mohitb35:
## Changes
- **Fixed locale persistence during shallow routing in project details page**
- Improved query parameter management for site and plant location navigation
- Cleaned up URL appearance by hiding internal routing params (locale, slug, p)
- Added proper TypeScript types and removed type assertions
- Updates back navigation logic to preserve the locale
- Refactors back navigation logic to ensure path was localized and avoid an additional middleware run

## Technical Details
1. Introduced proper parameter management in `buildProjectDetailsQuery` (renamed from `updateUrlWithParams`):
  - Safely converts Next.js router query params to string-only values
  - Preserves necessary routing parameters while cleaning up internal ones
  - Returns strongly typed `Record<string, string>` for URL query params

2. Enhanced `updateProjectDetailsPath` (renamed from `pushWithShallow`):
  - Maintains internal routing state while showing clean URLs
  - Uses proper URL handling for both browser display and Next.js router
  - Keeps locale consistent during navigation

3. Unified handling of both site and plant location navigation:
  - Both now use the same query parameter management flow
  - Consistent approach to shallow routing throughout the project details page
  - Proper TypeScript types for navigation parameters

4. Enhanced `generateProjectLink` logic (supplementary change)
  - Updates logic for `backNavigationUrl` to strip locale from path
  - Preserves locale on going back to an "in-app page" even if locale was changed on the project details page

5. Refactored `handleBackButton` logic (supplementary change)
  - Handle locale prefixing in the client-side navigation logic instead of relying on middleware
  - Properly handle relative paths without locale prefix (e.g. '/prd' -> '/en/prd')
  - Handle root path localization correctly (e.g. '/' -> '/en')
  - Move path localization logic to a separate utility function
  - These changes reduce middleware runs and avoid potential caching issues by handling locale prefixing directly in the navigation code
   
## Understanding the Main Fix
The core issue stemmed from the interaction between shallow routing and our middleware:

1. Normal Navigation:
  - Middleware rewrites `/de/prd/yucatan` to `/sites/planet/de/prd/yucatan`
  - Next.js extracts dynamic parameters (`locale`, `slug`, `p`) from the rewritten path
  - These parameters are available in `router.query` but not in the URL query string

3. Shallow Routing:
  - Bypasses middleware (no server-side execution)
  - No URL rewrite occurs
  - We need to manually maintain these parameters in query state
  - Parameters need to be hidden from the URL to maintain consistent URL structure

The fix ensures that during shallow routing:
- Internal routing parameters (`locale`, `slug`, `p`) are preserved in `router.query`
- URL remains clean and consistent with middleware-processed URLs
- Locale context is maintained for next-intl

### DIFF
--- a/src/features/projectsV2/ProjectsContext.tsx
+++ b/src/features/projectsV2/ProjectsContext.tsx
@@ -27,7 +27,7 @@ import getStoredCurrency from '../../utils/countryCurrency/getStoredCurrency';
 import { getRequest } from '../../utils/apiRequests/api';
 import { ErrorHandlingContext } from '../common/Layout/ErrorHandlingContext';
 import { useTenant } from '../common/Layout/TenantContext';
-import { updateUrlWithParams } from '../../utils/projectV2';
+import { buildProjectDetailsQuery } from '../../utils/projectV2';
 
 interface ProjectsState {
   projects: MapProject[] | null;
@@ -245,15 +245,37 @@ export const ProjectsProvider: FC<ProjectsProviderProps> = ({
       setPlantLocations(null);
   }, [page]);
 
-  const pushWithShallow = (
+  const updateProjectDetailsPath = (
     locale: string,
     projectSlug: string,
-    queryParams = {}
+    queryParams: Record<string, string> = {}
   ) => {
+    console.log('===== Pushing with shallow =====');
+    console.log('Locale:', locale);
+    console.log('Project Slug:', projectSlug);
+    console.log('Query Params:', queryParams);
+    console.log('Router Pathname:', router.pathname);
+    console.log('Router AsPath:', router.asPath);
+    console.log('Router Query:', router.query);
+    console.log('==============================');
     const pathname = `/${locale}/prd/${projectSlug}`;
-    router?.push({ pathname, query: queryParams }, undefined, {
-      shallow: true,
-    });
+
+    // Extract only the visible query params for the URL
+    const { locale: _, slug: __, p: ___, ...visibleParams } = queryParams;
+
+    router?.push(
+      {
+        pathname,
+        query: queryParams,
+      },
+      // Only show necessary params in the URL
+      `${pathname}${
+        Object.keys(visibleParams).length
+          ? '?' + new URLSearchParams(visibleParams).toString()
+          : ''
+      }`,
+      { shallow: true }
+    );
   };
 
   const updateUrlWithSiteId = (
@@ -261,12 +283,12 @@ export const ProjectsProvider: FC<ProjectsProviderProps> = ({
     projectSlug: string,
     siteId: string | null
   ) => {
-    const updatedQueryParams = updateUrlWithParams(
+    const updatedQueryParams = buildProjectDetailsQuery(
       router.asPath,
       router.query,
-      siteId
+      { siteId }
     );
-    pushWithShallow(locale, projectSlug, updatedQueryParams);
+    updateProjectDetailsPath(locale, projectSlug, updatedQueryParams);
   };
 
   const updateSiteAndUrl = (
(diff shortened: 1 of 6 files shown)


mariahosfeld (changes requested):
Seems to mostly work. Only functionality issue detected:
When I change the language while on the project details page, using the back arrow provided by the WebApp, it changes back to the language it had before.
How to reproduce:
English on Map Screen, select project, when in project details switch to any other language, use back arrow -> Language has switched back to English

mariahosfeld (commented):
Now the back navigation works, the language switched in the project details also persists after navigating back to the map view.

mariahosfeld (approved):

src/utils/projectV2.ts
```diff
@@ -44,22 +34,40 @@ const paramsToDelete = [
...
+  const currentQuery: Record<string, string> = Object.entries(query).reduce(
+    (stringQueryParams, [key, value]) => {
+      if (typeof value === 'string') {
+        stringQueryParams[key] = value;
+      }
+      return stringQueryParams;
+    },
+    {} as Record<string, string>
+  );
+
```
sunilsabatp, line 56: To improve readability, can we refactor the reduce method? We can use the logic below to achieve the same result.
```
const currentQuery: Record<string, string> = Object.fromEntries(
  Object.entries(query).filter(([_, value]) => typeof value === 'string')
);
```

sunilsabatp (changes requested):

src/features/projectsV2/ProjectsContext.tsx
```diff
@@ -245,28 +245,50 @@ export const ProjectsProvider: FC<ProjectsProviderProps> = ({
...
+    queryParams: Record<string, string> = {}
   ) => {
+    console.log('===== Pushing with shallow =====');
+    console.log('Locale:', locale);
+    console.log('Project Slug:', projectSlug);
+    console.log('Query Params:', queryParams);
+    console.log('Router Pathname:', router.pathname);
+    console.log('Router AsPath:', router.asPath);
+    console.log('Router Query:', router.query);
+    console.log('==============================');
```
sunilsabatp, line 260: Remove the console log's

src/utils/projectV2.ts
```diff
@@ -44,22 +34,40 @@ const paramsToDelete = [
...
-    if (currentUrlParams.has(param)) {
-      currentQuery[param] = currentUrlParams.get(param) ?? '';
-    }
-  });
+  routeParams: RouteParams
+): Record<string, string> => {
+  console.log('updateUrlWithParams:');
+  console.log('  Path:', asPath);
+  console.log('  Query:', query);
+  console.log('  Route Params:', routeParams);
```
sunilsabatp, line 45: Remove the console log's

src/utils/projectV2.ts
```diff
@@ -44,22 +34,40 @@ const paramsToDelete = [
...
+  if (routeParams.siteId) {
+    currentQuery.site = routeParams.siteId;
+  }
+
+  if (routeParams.plocId) {
+    currentQuery.ploc = routeParams.plocId;
+  }
+
+  console.log('updateUrlWithParams: Updated Query Parameters:');
+  console.log('  Current Query:', JSON.stringify(currentQuery, null, 2));
```
sunilsabatp, line 70: Remove the console logs

sunilsabatp (changes requested):

src/utils/projectV2.ts
mohitb35: Done, had to add an additional type check.

sunilsabatp (approved):
LGTM

PR #2289, opened 2024-11-06
Story of  address list

sunilsabatp:


mariahosfeld (changes requested):
The highlighting behind the 3 dots is not correctly aligned, dots should be the center (they are already placed correctly.
Also check alignment based on Figma design, it should be below and to the right side of the dots, not overlapping with the address field and the dots itself.
Top and bottom padding seems to be a bit much, should be the same as left and right.

mohitb35:
UI feedback
1. The primary/billing address tags should contain text in a single line
2. Address format needs to be worked on. Everything should not be fitted in a single line. I'd suggest keeping `address` on the first line, `address2` on the second line, and combining `zipcode`, `city`, `state` and `country` on the last. Also, `address2` should be shown if it has a value, this is currently not done.

DPR5A

src/features/user/Settings/EditProfile/AddressManagment/microComponents/AddressActionMenu.tsx
mohitb35, line 1: 1. The structure is different compared to the plan. While this is ok when needed, the plan should be updated. This applies for renaming as well, as it helps me compare the plan and the code easily.
2. One missing link - how would I identify which address an action is being performed on? There is no provision to pass in the address id.

src/features/user/Settings/EditProfile/AddressManagment/microComponents/AddressContent.tsx
mohitb35, line 1: 1. I feel we should simply pass in the `address`, as planned initially. It contains all of the information needed, and most fields within address appear in the content. 
2. You can probably perform the formatting within this component itself.

mohitb35 (dismissed):

mohitb35 (changes requested):
@sunilsabatp Please go through the feedback above

src/features/user/Settings/EditProfile/AddressManagment/AddressManagement.module.scss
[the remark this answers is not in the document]
sunilsabatp: I have removed hardcoded values in this PR
https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/2312

src/features/user/Settings/EditProfile/AddressManagment/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+export type AddressType = 'primary' | 'mailing' | 'other';
+
+export interface UpdatedAddress extends Address {
+  id: string;
+  type: AddressType;
+  name: string | null;
+  state: string | null;
+  isPrimary: boolean | null;
+  address2: string | null;
+}
```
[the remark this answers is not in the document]
mohitb35: I've updated planet-sdk to version 0.1.43, so please update in the webapp as well

src/features/user/Settings/EditProfile/AddressManagment/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+export const addressType = ['primary', 'mailing', 'other'];
+const AddressManagement = () => {
+  const { user } = useUserProps();
+  const tProfile = useTranslations('Profile.addressManagement');
+  const [userAddresses, setUserAddresses] = useState<UpdatedAddress[]>(
+    user?.addresses
+  ); // need to update planet-sdk to include addresses key
```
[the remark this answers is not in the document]
sunilsabatp: We are not using user.addresses because it does not provide an updated list of addresses if an address is deleted from the address list. Instead, I have used the /app/addresses endpoint to fetch the user's addresses rather than relying on the addresses from the profile response. I have addressed this issue in the Address Action PR https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/2308

src/features/user/Settings/EditProfile/AddressManagment/microComponents/AddressActionMenu.tsx
sunilsabatp: 1. I will update the Notion page document and rename the components as decided in the Notion page.

src/features/user/Settings/EditProfile/AddressManagment/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const [addressAction, setAddressAction] = useState<AddressAction | null>(
+    null
+  );
```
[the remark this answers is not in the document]
sunilsabatp: I have updated the notion docs 
1. ` `add` , `edit` , `delete` , `setPrimary` , `setBilling``  I have used `addressAction`  state variable
2.  `selectedAddressForAction`   I have covered in this PR -> https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/2304

src/features/user/Settings/EditProfile/index.tsx
```diff
@@ -4,17 +4,21 @@ import CenteredContainer from '../../../common/Layout/CenteredContainer';
...
+    <>
+      <DashboardView title={t('editProfile')} subtitle={null}>
+        <SingleColumnView>
+          <CenteredContainer>
+            <EditProfileForm />
+          </CenteredContainer>
+          <AddressManagement />
+        </SingleColumnView>
+      </DashboardView>
+    </>
```
mohitb35, line 22: You can remove the empty fragments, they serve no purpose now.

mohitb35:
General comment:
Resolve the TS warnings for type imports in all files changed by this PR.

src/features/user/Settings/EditProfile/AddressManagment/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+    return userAddresses.sort((a, b) => {
+      return (
+        addressTypeOrder.indexOf(a.type) - addressTypeOrder.indexOf(b.type)
+      );
+    });
+  }, [userAddresses]);
+
+  return (
+    <section className={styles.addressManagement}>
+      <h2 className={styles.addressManagementTitle}>
```
mohitb35, line 33: There is no `addressManagement` class defined in the scss file, so nothing appears in the browser inspector.

I'd suggest moving some styles from the `addressManagementTitle`

e.g.
```
.addressManagement {
	display: flex;
	flex-direction: column;
	margin-top: 40px;
	gap: 20px;
}

.addressManagementTitle {
  font-size: $fontLarge;
  font-weight: 600;
  margin: 40px 0 20px; //Remove this
}
```

mohitb35 (changes requested):
@sunilsabatp Minor feedback, please address and then this can be merged in.

mohitb35 (approved):
LGTM

PR #2292, opened 2024-11-07
Release Facelift WebApp

mariahosfeld:
**Includes:**

- #2149 
- #2294
- #2295 
- #2296


mariahosfeld:
Looking good to me.

mohitb35 (approved):
LGTM.

I think we need to monitor it for a few days after it goes live.

mariahosfeld:
> LGTM.
> 
> I think we need to monitor it for a few days after it goes live.

Agreed!

PR #2301, opened 2024-11-08
fix: adjust degradation year and year abandoned condition to exclude …

sunilsabatp:
- Updated the condition for `metadata.degradationYear`  and  `metadata.yearAbandoned` to ensure that projects with a degradation year and yearAbandoned year of `0` are excluded. 
- Minor position adjustment of the explore button


sunilsabatp:
Duplicate PR

PR #2306, opened 2024-11-12
Ensure planting density is always passed to PlantingDetails component

mohitb35:
Always display plantingDensity, regardless of sample interventions.


mariahosfeld (approved):
Looks good to me.

PR #2312, opened 2024-11-20
Edit profile form (code clean up)

sunilsabatp:
1. Removed the address-related field and its logic from the edit profile.
2. Removed hardcode stylings.
3. use `mixin` for consistent layout styling.


src/features/user/Settings/EditProfile/AddressManagment/AddressManagement.module.scss
```diff
@@ -1,8 +1,29 @@
 @import '../../../../../theme/theme';
 
-.addressManagement {
+$border-radius-sm: 8px;
+$border-radius-lg: 16px;
+$padding-standard: 24px;
```
mohitb35, line 5: Personally, I feel these variables didn't need to be abstracted, and the code is harder to read with them.

Do we want to change padding-standard in one place and affect both SingleAddress and the address shown in the type update modals?

mohitb35:
Address conflicts

PR #2316, opened 2024-11-27
feat: add address limit check and show message when limit is reached

sunilsabatp:
- Added a check to verify if the user has reached the maximum address limit (5).
- If the user has reached the limit, display a message informing them that they need to remove an existing address to add a new one.


mariahosfeld (dismissed):
Please remove the higlighted space, other than that it works for me
image

src/features/user/Settings/EditProfile/AddressManagment/index.tsx
```diff
@@ -132,6 +132,8 @@ const AddressManagement = () => {
     billingAddress,
     addressAction,
   ]);
+  const maxAddressLimit = 5;
```
mohitb35, line 135: Define this in src/utils/addressManagement.ts similar to the other constants

mohitb35 (approved):
@sunilsabatp Approved with minor feedback. You can merge after addressing that.

PR #2324, opened 2024-12-05
feat: add `locale` as a dependency for `getSearchProjects` useCallback

sunilsabatp:
Issue :
There was an issue where the search functionality only returned results when searching for `Germany` in German, but not for `Deutschland`. This issue occurred when the language was switched, and the search did not account for the localized term for Germany in German.

Solution:
Added `locale` as a  dependency  for `getSearchProjects` useCallback.

### DIFF
--- a/src/features/projectsV2/ProjectsContext.tsx
+++ b/src/features/projectsV2/ProjectsContext.tsx
@@ -171,7 +171,7 @@ export const ProjectsProvider: FC<ProjectsProviderProps> = ({
       });
       return filteredProjects;
     },
-    []
+    [locale]
   );
 
   const filteredProjects = useMemo(() => {


mariahosfeld (commented):
Funktionality looks good.

src/features/projectsV2/ProjectsContext.tsx
mohitb35, line 1: This fixes the issue of not being able to search by the translated country name once the locale is changed.

But I noticed a related issue - the projects are not fetched again when the locale is changed. As a result, when locale is changed, I cannot search by the translated project name.

For example, 
In french, the Yucatan project has the name `Restauration du Yucatán`
In English, the name is `Yucatán Restoration`

As the `projects` in the context are not updated when the `locale` changes, when I initially load the page in french, and switch to english, searching for 'yucatan restoration' does not produce a result.

mohitb35 (changes requested):

norbertschuler (approved):
LGFM

src/features/projectsV2/ProjectsContext.tsx
```diff
@@ -188,15 +189,12 @@ export const ProjectsProvider: FC<ProjectsProviderProps> = ({
 
   useEffect(() => {
     async function loadProjects() {
-      if (page !== 'project-list' || !currencyCode) {
-        return;
-      }
-      if (projects !== null) {
-        return;
-      }
+      setProjectLocale(locale);
```
mohitb35, line 192: Setting `projectLocale` here could lead to a race conditions where projectLocale is updated, but projects fail to load. Move this to after projects are successfully fetched.

src/features/projectsV2/ProjectsContext.tsx
```diff
@@ -100,6 +100,7 @@ export const ProjectsProvider: FC<ProjectsProviderProps> = ({
   >([]);
   const [debouncedSearchValue, setDebouncedSearchValue] = useState('');
   const [isSearching, setIsSearching] = useState(false);
+  const [projectLocale, setProjectLocale] = useState('');
```
mohitb35, line 103: Call this `projectsLocale` instead, as it is not for a single project.

mohitb35 (changes requested):
@sunilsabatp Please review and resolve

mohitb35 (approved):
LGTM

src/features/projectsV2/ProjectsContext.tsx
```diff
@@ -188,15 +189,11 @@ export const ProjectsProvider: FC<ProjectsProviderProps> = ({
...
     async function loadProjects() {
-      if (page !== 'project-list' || !currencyCode) {
-        return;
-      }
-      if (projects !== null) {
-        return;
-      }
+      if (page !== 'project-list' || !currencyCode) return;
+      if (projectsLocale === locale && projects !== null) return;
+
```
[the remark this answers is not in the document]
mohitb35: This is a false positive.

There is no difference between the suggested condition and the original condition, and we want to avoid calling the API if `projectsLocale === locale && projects !== null` , as we'd be fetching projects data again for the same language. We only fetch the data if the language has changed, or if projects were not fetched in the first place.

src/features/projectsV2/ProjectsContext.tsx
```diff
@@ -171,7 +171,7 @@ export const ProjectsProvider: FC<ProjectsProviderProps> = ({
       });
       return filteredProjects;
     },
-    []
+    [locale]
```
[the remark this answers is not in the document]
mohitb35: `locale` and `tCountry` change together, so just `locale` as a dependency is good enough.

PR #2337, opened 2024-12-16
Fix tab selection logic for topProjectsOnly condition

mohitb35:
Corrects the tab selection behavior to ensure the appropriate tab is selected when the tenant configuration indicates that only top projects are returned in the API, and no tabs are displayed.

Earlier, only top projects were being displayed by default, and there was no way to switch tabs.

Now, all projects are displayed by default when topProjectsOnly is true for a tenant.


mariahosfeld (commented):

norbertschuler (approved):
works for me (though I have the feeling we also should check the rest of the code if it still works as intended for the special Salesforce use case)

PR #2344, opened 2024-12-27
fix: show correct unit type for manage projects list screen

sunilsabatp:
#FIxes -
This PR ensures the correct unit type is used for restoration projects (unit type: m²; purpose: trees) in the "Manage Project" section.


src/features/user/ManageProjects/ProjectsContainer.tsx
```diff
@@ -65,7 +65,9 @@ function SingleProject({ project }: { project: Properties }) {
               Number(project.countPlanted),
               1
             )}{' '}
-            {tCommon('tree', { count: Number(project.countPlanted) })}
+            {project.unitType === 'm2'
+              ? tCommon(project.unitType)
+              : tCommon('tree', { count: Number(project.countPlanted) })}
           </p>
```
mohitb35, line 71: 1. I suggest simply using something like
`tCommon(project.unitType, {formattedCount, count})` 
instead of handling different unitTypes in different ways. Make sure to not affect places that use the existing translation keys - you probably need new keys.

2. API needs to be updated to return `countPlanted` / `areaConserved` etc. for `unitType = 'm2'`. This should consider conservation projects as well. Discuss this with @jmiridis

3. Show the total units value if unitType is 'trees' or 'm2'.

4. Update the type of project to reflect the API response. Be specific about the values `unitType` can take.

mohitb35 (changes requested):
@sunilsabatp Please go through my feedback

mohitb35 (changes requested):
- [x] Use the type from planet sdk once it is ready
- [x] After merging from develop, there is a client side error while impersonating the Plant-for-the-Planet tpo

```
TypeError: Cannot read properties of null (reading 'm2')

src/features/user/ManageProjects/ProjectsContainer.tsx (48:34)
  46 |     project.unitType === 'tree'
  47 |       ? (project as TreeProperties).unitsContributed?.tree
> 48 |       : project.unitsContributed.m2;
```

mohitb35 (approved):
LGTM

mariahosfeld (changes requested):
Purpose Trees and unitType m2 is not correctly rendered.

PR #2346, opened 2025-01-02
Update ESLint workflow to use Node.js setup action and specify versions

mohitb35:
Fixes the issue causing failure of the ESLint error report workflow.

Revise the ESLint workflow to utilize the Node.js setup action, ensuring consistent Node.js and npm versions.

The long term fix would be to update the node engine (as specified in package.json) to 20.x, but that would need to be tested.


sunilsabatp (approved):
LGTM

PR #2350, opened 2025-01-06
Bug: direct link to plant location does not work (first time)

sunilsabatp:
Fix :
Previously, visiting a `plant location via a direct link` in an incognito tab (e.g., https://dev.pp.eco/en/yucatan?ploc=JOCK9R) redirected to the default site instead of the specified plant location. This fix ensures the correct plant location is resolved and displayed when accessing such links, even in incognito mode.

### DIFF
--- a/src/features/projectsV2/ProjectSnippet/microComponents/ImageSection.tsx
+++ b/src/features/projectsV2/ProjectSnippet/microComponents/ImageSection.tsx
@@ -41,9 +41,7 @@ const ImageSection = (props: ImageSectionProps) => {
   const isEmbed = embed === 'true';
 
   const handleBackButton = () => {
-    if (setPreventShallowPush) {
-      setPreventShallowPush(true);
-    }
+    if (setPreventShallowPush) setPreventShallowPush(true);
 
     const previousPageRoute = sessionStorage.getItem('backNavigationUrl');
     const defaultRoute = `/${locale}`;
--- a/src/features/projectsV2/ProjectsContext.tsx
+++ b/src/features/projectsV2/ProjectsContext.tsx
@@ -314,7 +314,8 @@ export const ProjectsProvider: FC<ProjectsProviderProps> = ({
       page !== 'project-details' ||
       singleProject === null ||
       selectedSite !== null ||
-      (requestedPlantLocation && requestedSite)
+      (requestedPlantLocation && requestedSite) ||
+      plantLocations === null
     )
       return;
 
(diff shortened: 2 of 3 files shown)


mariahosfeld (dismissed):
Same issue as already remarked in PR #2352 : Loading-time is too long

src/utils/projectV2.ts
```diff
@@ -263,7 +263,7 @@ export const getLocalizedPath = (path: string, locale: string): string => {
   }
 
   // If path already starts with locale, return as is
-  if (cleanPath.startsWith(`/${locale}/`)) {
+  if (cleanPath.startsWith(`/${locale}`)) {
```
mohitb35, line 266: Couldn't this cause an issue?

What if the cleanPath starts with `/france-salesforce` which is a project slug?

mohitb35 (changes requested):
@sunilsabatp There are a few issues with the changes:

1. Instead of checking `plantLocations = null` and blocking site selection (url update) / map zoom when not needed, introduce a dedicated `isLoadingPlantLocations` state. This way the purpose is more explicit.
2. Check the `isLoadingPlantLocations` only when needed e.g. when `ploc` is passed as a url query param, or a selectedPlantLocation exists.
3. The current implementation prevents site loading when the project is not a `tree` purpose project. e.g. conservation-aa-andes

You might want to consider separating site selection logic from plant location handling (perhaps not now, but as part of a near future update)

sunilsabatp:
> @sunilsabatp There are a few issues with the changes that add a null check for `plantLocations` before loading data
> 
> 1. Instead of checking `plantLocations = null` and blocking site selection (url update) / map zoom when not needed, introduce a dedicated `isLoadingPlantLocations` state. This way the purpose is more explicit.
> 2. Check the `isLoadingPlantLocations` only when needed e.g. when `ploc` is passed as a url query param, or a selectedPlantLocation exists.
> 3. The current implementation prevents site loading when the project is not a `tree` purpose project. e.g. conservation-aa-andes
> 
> You might want to consider separating site selection logic from plant location handling (perhaps not now, but as part of a near future update)

Instead of introducing a new state, I have followed a different approach. Please review it and provide your feedback.

mohitb35:
@sunilsabatp Can we close this?

sunilsabatp:
> @sunilsabatp Can we close this?

Yes, we should close this. I’ve tested the dev branch multiple times, and direct visits to the intervention  using query parameters are working perfectly fine.

mohitb35:
Not needed as issue fixed by #2352

PR #2352, opened 2025-01-08
Hotfix/map zoom out

sunilsabatp:
Issue: When visiting a specific plant location via a direct link (e.g., [link](https://dev.pp.eco/en/yucatan?ploc=JOCK9R)), and then using the back button to return to the landing page, the map becomes stuck at an incorrect zoom level and latitude/longitude coordinates. This behavior contradicts the expected behavior set by the flyTo() function, which is responsible for zooming the map out.

Fix: This PR fixes the issue by ensuring that the map zoom level and coordinates reset correctly when navigating back from the plant location.

### DIFF
--- a/src/features/projectsV2/ProjectsMap/index.tsx
+++ b/src/features/projectsV2/ProjectsMap/index.tsx
@@ -4,7 +4,7 @@ import type { SetState } from '../../common/types/common';
 import type { PlantLocationSingle } from '../../common/types/plantLocation';
 import type { ExtendedMapLibreMap, MapRef } from '../../common/types/projectv2';
 
-import { useCallback, useMemo, useRef } from 'react';
+import { useCallback, useEffect, useMemo, useRef } from 'react';
 import 'maplibre-gl/dist/maplibre-gl.css';
 import Map, { NavigationControl } from 'react-map-gl-v7/maplibre';
 import { useProjectsMap } from '../ProjectsMapContext';
@@ -24,6 +24,7 @@ import MultiPlantLocationInfo from '../ProjectDetails/components/MultiPlantLocat
 import SinglePlantLocationInfo from '../ProjectDetails/components/SinglePlantLocationInfo';
 import styles from './ProjectsMap.module.scss';
 import { useDebouncedEffect } from '../../../utils/useDebouncedEffect';
+import { zoomOutMap } from '../../../utils/mapsV2/zoomToProjectSite';
 
 export type ProjectsMapDesktopProps = {
   isMobile: false;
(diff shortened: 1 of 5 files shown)


mariahosfeld (dismissed):
When using a direct link to the ploc, the time until the zooming in to the location on the map is too long. This needs to be improved.
Zooming out works as expected.

mariahosfeld (commented):
General functionality works, performance needs to be monitored.

mariahosfeld (approved):
Performance still needs to be monitored (takes a bit to zoom in when testing in an incognito window), but intended functionality works.

mohitb35 (commented):
While the PR seems to work in terms of functionality, I have 2 concerns:

1. Zoom performance seems laggy, especially after using the map for a while and going back to the home page or resetting to the default zoom. In addition, memory usage seems pretty high, but that may not be related to this.
2. The code seems to be getting more complicated and harder to review, with all the useEffects and dependencies. useEffects are getting triggered multiple times due to multiple dependencies that represent the same state or change at the same time. Perhaps not in this PR, but I think it would be good to simplify the logic that triggers map state changes and url parameter changes soon.

src/features/projectsV2/ProjectsMap/index.tsx
```diff
@@ -179,6 +180,25 @@ function ProjectsMap(props: ProjectsMapProps) {
...
+          handleViewStateChange({
+            ...map.getCenter(),
+            zoom: map.getZoom(),
+          });
+        });
+      }
+    });
+  }, [props.page]);
+
   const onMove = useCallback(
```
mohitb35, line 202: Would it make sense to adapt this useEffect as follows:

```
useEffect(() => {
    if (props.page === 'project-details') return;
    // Only execute zoom logic when map is fully loaded
    if (mapLoaded && mapRef.current) {
      try {
        const map = mapRef.current.getMap
          ? mapRef.current.getMap()
          : mapRef.current;
        zoomOutMap(map, () => {
          handleViewStateChange({
            ...map.getCenter(),
            zoom: map.getZoom(),
          });
        });
      } catch (error) {
        console.error('Failed to zoom out map on page change:', error);
      }
    }
  }, [props.page, mapLoaded]);
```

Reasoning:
1. Promise.resolve().then() is unreliable - This pattern only defers to the next microtask, not actual map initialization. Using the existing mapLoaded state or a proper callback from map initialization would be more reliable.
2. Adds error handling as zoomOutMap operation could fail without any error boundaries.

mohitb35 (changes requested):
@sunilsabatp

Please review my feedback

PR #2359, opened 2025-01-10
Icon changes

rupamkairi:
This covers Icons changes proposed in the following tasks.
https://process.hub.plant-for-the-planet.org/browse/ET-206
https://process.hub.plant-for-the-planet.org/browse/ET-207

For reference
Screenshot 2025-01-10 at 21 32 40
Screenshot 2025-01-10 at 21 33 02


mariahosfeld (changes requested):
@rupamkairi , please also change the icon in the edit profile form (see attached screenshot).
image

public/assets/images/icons/headerIcons/DefaultProfileImageIcon.tsx
```diff
@@ -1,13 +1,45 @@
...
-    <svg viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
-      <circle cx="12" cy="12" r="12" fill="#DFF4E8" />
-      <path
-        fillRule="evenodd"
-        clipRule="evenodd"
-        d="M16.4632 9.93092C16.4632 9.93092 15.6142 9.97331 15.423 10.8185C15.423 10.8185 15.4202 8.32812 13.5176 6.94321C13.5176 6.94321 12.9615 7.37839 12.9727 8.28421C12.9727 8.28421 11.3727 6.4935 7.93948 5.60212C6.98604 9.05149 7.74561 11.3021 7.74561 11.3021C6.96679 10.8394 6.31185 11.1034 6.31185 11.1034C6.56035 13.4433 8.71523 14.6912 8.71523 14.6912C7.88769 14.4342 7.42645 15.1483 7.42645 15.1483C9.48192 17.8293 12.2914 18.2031 14.3401 17.1436C14.4749 17.0738 14.6066 16.9978 14.7344 16.9159C16.6763 15.6714 17.7575 13.0519 16.4632 9.93092ZM14.3401 17.1436L13.4524 15.6061L10.8222 15.3421L10.8748 14.8956L13.1756 15.1267L11.9034 12.9232L9.27324 12.6591L9.32553 12.2122L11.6264 12.4434L9.97035 9.57501L10.3647 9.34735L12.0207 12.2157L12.971 10.1075L13.3841 10.2857L12.2977 12.6955L13.57 14.8991L14.5202 12.7909L14.9331 12.9686L13.8467 15.3784L16.1366 19.3447L15.7423 19.5724L14.3401 17.1436Z"
-        fill="#219653"
+    <svg
+      width="48"
+      height="48"
```
mohitb35, line 5: I suggest removing `width` and `height` - makes it easier for the svg to take the container dimensions while width and height remain in proportion.

There is a minor difference in svg dimensions within the `userDefaultIconContainer` due to these styles

```
.profileImageButton .userDefaultIconContainer svg {
    height: 100%;
}
```

Since width is specified as 48, the width remains as is, while the height adapts to 100% (i.e. 50px)

In this case it is not noticeable, but it's still a good idea to let the svg height and width adapt to maintain aspect ratio.

src/features/projectsV2/ProjectListControls/microComponents/ClassificationDropDown.tsx
```diff
@@ -6,13 +6,41 @@ import { useMemo } from 'react';
...
+  ),
+  'managed-regeneration': (
+    <ManagedRegeneration color="black" width={'20'} height={'20'} />
+  ),
+  'urban-planting': (
+    <UrbanRestoration color="black" width={'20'} height={'20'} />
+  ),
+  'other-planting': <OtherPlanting color="black" width={'20'} height={'20'} />,
+  mangroves: <Mangroves color="black" width={'20'} height={'20'} />,
+};
```
mohitb35, line 42: 1. If the width and height are constant, and these icons are only being used in one place, would it make sense to provide width and height default values of 20 to the icon components themselves? 
2. The color should not be black - it should match the text color (rgb(51,51,51)) which cascades from the following styles
```
.filterButton {
  ....
  color: rgba(51, 51, 51, 1);
}
```
I suggest using css to specify the svg color in this instance.

mohitb35 (changes requested):
@rupamkairi have left some minor feedback

src/features/projectsV2/ProjectListControls/microComponents/ClassificationDropDown.tsx
rupamkairi: Sure, let me try.

mariahosfeld:
@rupamkairi , could you remove the frame around this instance of the icon (path /profile)
image

rupamkairi:
> @rupamkairi , could you remove the frame around this instance of the icon (path /profile) image

That might require removing the background color (a gray background color is there.).
Sure on it.

mariahosfeld (commented):
Looks good to me.

public/assets/images/icons/headerIcons/DefaultProfileImageIcon.tsx
mohitb35, line 1: I noticed that this svg has some white space included inside the graphic. As a result, the icon leaves a little white space at the edge of the container. Do we want to go with that or reduce the space?

mohitb35, line 1: I noticed that this svg has some white space included inside the graphic. As a result, the icon leaves a little visible white space at the edge of the container element. Do we want to go with that or reduce the space?
Screenshot 2025-01-14 at 12 53 39 PM
Screenshot 2025-01-14 at 12 53 34 PM

rupamkairi: Made some tweaking to the svg itself. It is not exactly same to this, but its close. Please check & let me know for further tweaking.

Screenshot 2025-01-14 at 11 41 59

public/assets/images/icons/projectV2/PointMarkerIconsSymbol.tsx
```diff
@@ -0,0 +1,272 @@
+// Todo - move to a common location
+import type { IconProps } from '../../../../../src/features/common/types/common';
+
+export const Agroforestry = ({
+  color = 'defaultColor',
```
mohitb35, line 5: I don't think `defaultColor` does anything - you'd need to define a css variable for this to work.

A simpler approach would be to simply define a default color of #333, or use `currentColor` instead.

mohitb35 (approved):
LGTM

public/assets/images/icons/projectV2/PointMarkerIconsSymbol.tsx
```diff
@@ -0,0 +1,272 @@
...
+        fill={color}
+      />
+    </svg>
+  );
+};
```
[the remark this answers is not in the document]
mohitb35: ignore

mariahosfeld (approved):

PR #2365, opened 2025-01-14
refactor: refactor satellite view logic in useEffect

sunilsabatp:
This PR refactors the logic for setting the satellite view `based on the project purpose (conservation or trees)` and the `availability of plant locations(TreeMapper Data)`. The changes `simplify the useEffect hook` that previously handled these conditions and streamline the flow for determining whether the satellite view should be enabled or not.

`Changes:`
Replaced the previous if conditions and checks for plant locations with a simplified expression.
Satellite view is now set to true for conservation purposes and when there are no plant locations for tree purposes.
The logic for determining satellite view has been consolidated, making it more readable and maintainable.
`Impact:`
Improves readability and clarity of the logic determining when satellite view should be enabled.
Reduces redundant conditional checks, making the code easier to maintain in the future.

### DIFF
--- a/src/features/projectsV2/ProjectsMap/SingleProjectView.tsx
+++ b/src/features/projectsV2/ProjectsMap/SingleProjectView.tsx
@@ -108,11 +108,11 @@ const SingleProjectView = ({ mapRef }: Props) => {
   ]);
 
   useEffect(() => {
-    setIsSatelliteView(
+    const isSatelliteView =
       singleProject.purpose === 'conservation' ||
-        (singleProject.purpose === 'trees' &&
-          (!plantLocations || plantLocations.length === 0))
-    );
+      (singleProject.purpose === 'trees' && plantLocations?.length === 0);
+
+    setIsSatelliteView(isSatelliteView);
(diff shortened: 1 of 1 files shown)


norbertschuler (approved):
Seems to work for me. Somtimes when you switch between different projects (one with default sattelite view, the other without), the switch between sattelite and map or via verse happens while the zooming is already happening, but I think that might be ok.

mariahosfeld (commented):
When loading a project with a lot of plant locations for the first time, the initial load shows the sat layer, once the locations have been loaded, it just switches over to the other view. Is there a way to make this at least more smoothly?

sunilsabatp:
> When loading a project with a lot of plant locations for the first time, the initial load shows the sat layer, once the locations have been loaded, it just switches over to the other view. Is there a way to make this at least more smoothly?

I've refactored the logic to handle the initial state of plantLocations more efficiently. Now, the satellite view won't toggle unnecessarily during the first load. Instead, it will smoothly switch views based on the presence of plant location data @mariahosfeld

mariahosfeld (commented):
Looks good to me.

PR #2369, opened 2025-01-16
Migrate database connection from MySQL to PostgreSQL (Data explorer)

mohitb35:
Replace MySQL with PostgreSQL for database connections for data explorer APIS + implement a new query helper function, and adapt data explorer APIs accordingly. 

Also rotates the labels of Trees Planted graph when more than 15 columns are to be shown (or when space is less, whichever is earlier).

Pending

- [x] Uninstall serverless-mysql

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -96,6 +96,7 @@
         "read-more-react": "^1.0.10",
         "sass": "^1.62.1",
         "serverless-mysql": "^1.5.4",
+        "serverless-postgres": "^2.1.1",
         "slick-carousel": "^1.8.1",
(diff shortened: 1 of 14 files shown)


mariahosfeld (commented):
Looks good to me. But could you make sure that the preselection for the projects in the DataExplorer is the same as currently on production?

mohitb35:
> Looks good to me. But could you make sure that the preselection for the projects in the DataExplorer is the same as currently on production?

That order is as per the `/app/profile/projects?scope=map` API, so production should remain unchanged.

mariahosfeld (approved):

PR #2386, opened 2025-01-28
intervention condition added

shyambhongle:
Changes in this pull request:
In this PR  intervention filter condition has been added to the PlantLocation component to ensure that sample trees are hidden on the map when an intervention without sample trees is selected.

@sunil-sabat


sunilsabatp (approved):

mariahosfeld (commented):
Looks good to me.

PR #2394, opened 2025-01-31
Tax receipt/modify donor data form

sunilsabatp:
Key Features :

- [x] Profile Information Section – Users can update their first name, last name, company name, and Tax Identification Number (TIN).
- [x] Address List with Selection – Users can select a primary address from a list of saved addresses.
- [x] Primary Address Toggle – A checkbox allows users to mark an address as primary.
- [x] Edit Address Option – Each address has an edit button for modifications.
- [x] Add New Address Button – Enables users to add additional addresses.
- [x] Save & Return Button – Saves changes and navigates back to the previous page.

 [Figma design](https://www.figma.com/design/lT7AVzzAUM0NyThCyvacob/MyForest-Dashboard?node-id=4824-1474&t=DuzjcX4UtTCalfV4-0)
 
 Important points:

- A donor cannot directly access this page unless they come from the verification page.
- If the donor updates any credentials on this page (such as address or name), their profile will also be updated, and the same data will be rendered on the receipt once they click the `"Save and Proceed"` button.
- If the donor updates any credentials on this page (such as address or name), `hasDonorDataChanged `will be set to true.

 
 To Do :
 

- [x]  Merge the feedbacks addressed in the base branch


mohitb35:
When I am not logged in initially, and I try to modify contact details, I end up seeing the AccessDeniedLoader instead of the screen to modify contact details.

Screenshot 2025-02-13 at 12 32 49 PM

mohitb35:
Because the verification page does not lie in the dashboard, and the modify contact details page lies within the dashboard, there is a layout shift which does not look great.

I feel we should not keep the modify contact details page within the dashboard (at least if the user is coming from the verify receipt page.

sunilsabatp:
> When I am not logged in initially, and I try to modify contact details, I end up seeing the AccessDeniedLoader instead of the screen to modify contact details.
> 
> Screenshot 2025-02-13 at 12 32 49 PM

This is a one-way process; you cannot directly jump to this page unless you come from the verification page.

sunilsabatp:
> Because the verification page does not lie in the dashboard, and the modify contact details page lies within the dashboard, there is a layout shift which does not look great.
> 
> I feel we should not keep the modify contact details page within the dashboard (at least if the user is coming from the verify receipt page.

I agree,  can  we add this as a future enhancement ?

mohitb35:
General comment:

Remove files like `DonorDataForm` which are not being used.

mohitb35 (changes requested):
@sunilsabatp Please go through my feedback

src/features/user/DonationReceipt/DonorContactManagement.tsx
```diff
@@ -0,0 +1,183 @@
...
+        logoutUser,
+      });
+      if (res) setDonorAddresses(res);
+    } catch (error) {
+      setErrors(handleError(error as APIError));
+    }
+  }, [user, token, contextLoaded, tenantConfig.id, logoutUser, setErrors]);
+
+  useEffect(() => {
+    updateDonorAddresses();
```
mohitb35, line 121: You probably don't need this initially, as the profile has the `addresses` property.

src/features/user/DonationReceipt/microComponents/DonorAddress.tsx
mohitb35, line 1: Rename to DonorAddressList for clarity

src/features/user/DonationReceipt/microComponents/DonorContactForm.tsx
```diff
@@ -0,0 +1,282 @@
...
+              guid: checkedAddressGuid,
+              address1,
+              address2,
+              country,
+              zipCode,
+              city,
+            },
+            hasDonorDataChanged: true,
+          };
+        });
```
mohitb35, line 176: USe `updateDonationReceiptData` instead

src/features/user/DonationReceipt/DonorContactManagement.tsx
sunilsabatp: We need this logic "`For example`, if a user visits the donor contact management page and adds a new address, then returns to the verification page, and subsequently goes back to the donor contact management page, the `donorAddresses `state gets set to the address from the profile initially. As a result, we won't see the updated address."

mariahosfeld (dismissed):
After confirming, the download-button should not only be in the lower right corner, but as big as the original confirm-button and centered.

src/features/user/DonationReceipt/microComponents/DonorContactForm.tsx
```diff
@@ -0,0 +1,280 @@
...
+          hasDonorDataChanged: true,
+          amount,
+          currency,
+          donations,
+          dtn,
+          year,
+          challenge,
+          downloadUrl,
+          verificationDate,
+        });
```
mohitb35, line 174: Why are properties that are not being updated being passed in to `updateDonationReceiptData`, it already considers the previous state.

src/features/user/Settings/EditProfile/AddressManagement/AddAddress.tsx
```diff
@@ -70,8 +73,25 @@ const AddAddress = ({
...
+                      : addr
+                  )
+                : prev.addresses;
+
+            return {
+              ...prev,
+              addresses: [...updatedAddresses, res],
+            };
+          });
+          updateUserAddresses();
```
mohitb35, line 94: After updating the user object, is `updateUserAddresses()` actually needed? The user object will contain the updated addresses.

src/features/user/DonationReceipt/microComponents/DonorContactForm.tsx
```diff
@@ -0,0 +1,280 @@
...
+            name={'tin'}
+            control={control}
+            rules={{ required: t('donorInfo.tinRequired') }}
+            label={t('donorInfo.tin')}
+          />
+        )}
+      </InlineFormDisplayGroup>
+
+      <section className={styles.donorAddressSection}>
+        {donorAddresses.map((address) => {
```
mohitb35, line 234: Consider whether `donorAddresses` is needed, as the `user` will have updated address information.

public/static/locales/en/donate.json
```diff
@@ -43,6 +43,7 @@
     "cityRequired": "City is required",
     "countryRequired": "Country is required",
     "companyRequired": "Company Name is required",
+    "tinRequired": "Tin is required",
```
mohitb35, line 46: This seems unused

src/features/user/DonationReceipt/utils.ts
```diff
@@ -40,3 +46,68 @@ export const getVerificationDate = () => {
...
+ * Compares the receipt address with the profile address to determine if they match.
+ * Since the receipt address lacks a GUID, it performs a field-by-field comparison.
+ * (Note: The GUID is only available once the donor has completed the process on the donor contact management page.)
+ *
+ * @param profileAddress - The address stored in the user's profile.
+ * @param receiptAddress - The receipt address to be compared.
+ * @returns {boolean} - Returns `true` if both addresses match, otherwise `false`.
+ */
+export const isMatchingAddress = (
+  profileAddress: Record<string, any>,
```
mohitb35, line 60: Can you address the TS warning by specifying a different type than `any` here?

src/features/user/DonationReceipt/utils.ts
```diff
@@ -1,4 +1,9 @@
-import type { ReceiptData, ReceiptDataAPI } from './donationReceiptTypes';
+import type { Address, User } from '@planet-sdk/common';
+import type {
```
mohitb35, line 2: Clean up unused imports

mohitb35 (changes requested):
@sunilsabatp - Please review my feedback

pages/_app.tsx
```diff
@@ -302,12 +303,14 @@ const PlanetWeb = ({
...
-                                          ) : null}
-                                          {pageContent}
+                                          <DonationReceiptProvider>
+                                            {isMap ? (
+                                              <MapHolder
+                                                setshowVideo={setshowVideo}
+                                              />
+                                            ) : null}
+                                            {pageContent}
+                                          </DonationReceiptProvider>
```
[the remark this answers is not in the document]
mohitb35: Have added a note in #2409

src/features/user/DonationReceipt/microComponents/DonorContactForm.tsx
sunilsabatp: The `updateDonationReceiptData` is overwriting the previous state incorrectly, The issue here is that `formatReceiptData(data)` does not merge with the previous state; it creates a new object with defaults ('', 0, null) for missing properties.

Since formatReceiptData assigns defaults for missing properties, any omitted values (e.g., amount, paymentDate) will get replaced by these defaults, even though they exist in prevState. That is why I had to pass those properties

mohitb35: How about if we accept the existing state into `formatReceiptData`?

```
export const formatReceiptData = (
  data: Partial<ReceiptDataAPI>,
  existingData?: ReceiptData
): ReceiptData => {
  // If we have existing data, start with it as the base
  const baseData: Partial<ReceiptData> = existingData || {};
  
  // Extract basic fields that exist in both types
  const commonFields: Partial<ReceiptData> = {
    dtn: data.dtn ?? baseData.dtn ?? '',
    year: data.year ?? baseData.year ?? '',
    challenge: data.challenge ?? baseData.challenge ?? '',
    currency: data.currency ?? baseData.currency ?? '',
    amount: data.amount ?? baseData.amount ?? 0,
    verificationDate: data.verificationDate ?? baseData.verificationDate ?? null,
    downloadUrl: data.downloadUrl ?? baseData.downloadUrl ?? '',
    donations: data.donations ?? baseData.donations ?? [],
  };

  // Transform donor data if provided, otherwise keep existing
  const donorView: DonorView = {
    tin: data.donor?.tin ?? baseData.donor?.tin ?? null,
    name: data.donor?.name ?? baseData.donor?.name ?? '',
    type: data.donor?.type ?? baseData.donor?.type ?? null,
  };

  // Transform address data if provided, otherwise keep existing
  const addressView: AddressView = {
    city: data.donor?.city ?? baseData.address?.city ?? '',
    country: data.donor?.country ?? baseData.address?.country ?? '',
    zipCode: data.donor?.zipCode ?? baseData.address?.zipCode ?? '',
    address1: data.donor?.address1 ?? baseData.address?.address1 ?? '',
    address2: data.donor?.address2 ?? baseData.address?.address2 ?? null,
    guid: data.donor?.guid ?? baseData.address?.guid ?? null,
  };

  // Construct the final ReceiptData object
  return {
    ...commonFields,
    donor: donorView,
    address: addressView,
    hasDonorDataChanged: data.hasDonorDataChanged ?? baseData.hasDonorDataChanged ?? false,
    operation: data.verificationDate === null ? 'verify' : 'download',
  };
};
```

and pass the existing value in from the `updateDonationReceipt` function in DonationReceiptContext

```
const updateDonationReceiptData = (data: Partial<ReceiptDataAPI>) => {
    const formattedData = formatReceiptData(data, donationReceiptData);
    setDonationReceiptData(formattedData);
  };
```

Then you can simply pass in the modified data in DonorContactForm
```
const donorDetails = getUpdatedDonorDetails(updatedUser, checkedAddressGuid);
      updateDonationReceiptData({
        donor: {
          ...donorDetails,
          type: updatedUser.type === 'individual' ? 'individual' : 'organization',
          guid: checkedAddressGuid,
        },
        hasDonorDataChanged: true
      });
```

src/features/user/Settings/EditProfile/AddressManagement/AddAddress.tsx
```diff
@@ -106,6 +127,9 @@ const AddAddress = ({
         defaultAddressDetail={defaultAddressDetail}
         processFormData={addAddress}
```
mohitb35, line 128: Type mismatch for `defaultAddressDetail` prop

src/features/user/Settings/EditProfile/AddressManagement/AddAddress.tsx
mohitb35: While this has been made optional, I don't think its needed for src/features/user/Settings/EditProfile/AddressManagement/index.tsx as well.

src/features/user/DonationReceipt/microComponents/DonorContactForm.tsx
mohitb35: This is done with a slightly different approach, but the implementation is complex and prone to errors. The default receipt data handling is split between two places (default function and format function) instead of being centralized. Please simplify. Using the formatReceiptData function as suggested before would make the logic simpler.

src/features/user/DonationReceipt/microComponents/DonorContactForm.tsx
```diff
@@ -79,7 +77,7 @@ const DonorContactForm = ({
   const tAddressManagement = useTranslations('EditProfile.addressManagement');
   const t = useTranslations('DonationReceipt');
   const { user, contextLoaded, token, setUser, logoutUser } = useUserProps();
-  if (!user || !donationReceiptData) return null;
+  if (!user) return null;
```
mohitb35, line 80: Does this make it possible to access this page without donationReceiptData? Does #2411 take care of this?

src/features/user/Settings/EditProfile/AddressManagement/EditAddress.tsx
mohitb35, line 1: Is `updateUserAddresses` needed at all? Since the user object contains the necessary information already?

src/features/user/Settings/EditProfile/AddressManagement/AddAddress.tsx
sunilsabatp: Created a [issue](https://github.com/Plant-for-the-Planet-org/planet-webapp/issues/2417) it

src/features/user/Settings/EditProfile/AddressManagement/EditAddress.tsx
sunilsabatp: Created  a issue for this

mohitb35 (approved):
LGTM after adding in #2411

PR #2396, opened 2025-02-05
hide filter option condition added

shyambhongle:
In this PR, I have added a filter condition to hide options when an intervention is not available in the project.

- If only one intervention is present, only that option will be shown.
- If multiple interventions are available, the default selection will be "All Interventions," along with the specific interventions present.
Note: I have removed the default "Single/Multi Intervention" option, as it was confusing and seemed unnecessary.

### DIFF
--- a/src/features/projectsV2/ProjectsMap/InterventionDropDown/InterventionList.tsx
+++ b/src/features/projectsV2/ProjectsMap/InterventionDropDown/InterventionList.tsx
@@ -31,30 +31,35 @@ const InterventionList = ({
     setSelectedInterventionType(key);
   };
 
+  const shouldRenderIntervention = (interventionValue: string) => {
+    const showAllIntervention = interventionValue === 'all';
+    const showExisitingIntervention = existingIntervention.includes(interventionValue);
+    if (showAllIntervention && existingIntervention.length === 1) {
+      return false;
+    }
+    return showExisitingIntervention || showAllIntervention;
+  };
+
 
 
   return (
-    <ul className={`${styles.interventionListOptions} ${!hasProjectSites ? styles.interventionListOptionsAbove : styles.interventionListOptionsBelow}`}>
-      {interventionList.map((intervention, index) => {
-        const exists = existingIntervention.includes(intervention.value);
-        if (!exists && intervention.value !== 'all') {
+    <ul
+      className={`${styles.interventionListOptions} ${!hasProjectSites ? styles.interventionListOptionsAbove : styles.interventionListOptionsBelow
+        }`}
+    >
+      {interventionList.map((intervention) => {
+        if (!shouldRenderIntervention(intervention.value)) {
           return null;
         }
-        if (intervention.value == 'all' && existingIntervention.length === 1) {
-          return
-        }
+
         return (
           <li
-            className={`${styles.listItem} ${intervention.value === selectedInterventionData?.value
-              ? styles.selectedItem
-              : ''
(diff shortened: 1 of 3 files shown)


src/features/projectsV2/ProjectsMap/InterventionDropDown/index.tsx
```diff
@@ -54,20 +56,29 @@ const InterventionDropdown = ({
...
     setIsMenuOpen((prev) => !prev);
   };
+
+  const showVisibleOption = () => {
+    if (existingIntervention.length === 1) {
+      return findMatchingIntervention(existingIntervention[0]);
+    }
+    return findMatchingIntervention(selectedInterventionType);
+  }
+
```
sunilsabatp, line 71: - You can refactor the showVisibleOption function for better readability and maintainability like this:

```
const showVisibleOption = () => {
  const interventionToCheck = existingIntervention.length === 1
    ? existingIntervention[0]
    : selectedInterventionType;

  return findMatchingIntervention(interventionToCheck);
}
```
This reduces the conditional logic in the return statement by assigning the value to interventionToCheck first, making the function more concise.

- The name `showVisibleOption` doesn't clearly convey the actual purpose of the function. Based on what the function does—finding a matching intervention depending on certain conditions—it would be better to use a name that reflects that action.

src/features/projectsV2/ProjectsMap/InterventionDropDown/InterventionList.tsx
```diff
@@ -15,23 +15,34 @@ interface InterventionListProps {
...
   return (
     <ul className={`${styles.interventionListOptions} ${!hasProjectSites ? styles.interventionListOptionsAbove : styles.interventionListOptionsBelow}`}>
       {interventionList.map((intervention, index) => {
+        const exists = existingIntervention.includes(intervention.value);
+        if (!exists && intervention.value !== 'all') {
+          return null;
+        }
+        if (intervention.value == 'all' && existingIntervention.length === 1) {
+          return
+        }
```
sunilsabatp, line 45: The multiple conditionals inside the map() function could be simplified to make the logic easier to follow. Specifically, the combination of `existingIntervention.includes()` and the `intervention.value !== 'all' `condition can be separated for clarity.

  ```
const shouldRenderIntervention = (interventionValue: string) => {
    const showAllIntervention = interventionValue === 'all';
    const showExisitingIntervention= existingIntervention.includes(interventionValue);
    if (showAllIntervention && existingIntervention.length === 1) {
      return false;
    }
    return showExisitingIntervention || showAllIntervention ;
  };
```
  
  
```
 return (
    <ul
      className={`${styles.interventionListOptions} ${
        !hasProjectSites ? styles.interventionListOptionsAbove : styles.interventionListOptionsBelow
      }`}
    >
      {interventionList.map((intervention) => {
        if (!shouldRenderIntervention(intervention.value)) {
          return null;
        }

        return (
          <li
            className={`${styles.listItem} ${
              intervention.value === selectedInterventionData?.value ? styles.selectedItem : ''
            }`}
            onClick={() => handleFilterSelection(intervention.value)}
            key={intervention.value} // Use unique value as key
          >
            <p>{tProjectDetails(intervention.value)}</p>
          </li>
        );
      })}
    </ul>
  );
```

sunilsabatp (changes requested):

shyambhongle:
@sunilsabatp I have refactored code based on your feedback, could you please review it again.

sunilsabatp (approved):

PR #2398, opened 2025-02-11
feat: show total donation value.

sunilsabatp:
This PR ensures that the total donated amount is displayed when there are multiple donations. It introduces a condition to check if more than one donation exists and renders the total amount accordingly. Additionally, minor code optimizations improve readability and maintainability.


mohitb35 (approved):
LGTM.

PR #2408, opened 2025-02-18
Fix double locale bug in back navigation

mohitb35:
Resolve an issue with back navigation that caused double locale entries by improving the logic for handling paths that already contain locales.

To reproduce issue: 

- Open https://dev.pp.eco/en/yucatan in a new browser tab. After the page loads, press back. You should see the url https://dev.pp.eco/en/en and an error message ("Resource not found").

- Try out the same approach with the preview link in this PR (append /yucatan or /en/yucatan after the link), the issue should no longer be seen

### DIFF
--- a/src/utils/projectV2.ts
+++ b/src/utils/projectV2.ts
@@ -262,8 +262,9 @@ export const getLocalizedPath = (path: string, locale: string): string => {
     return `/${locale}`;
   }
 
-  // If path already starts with locale, return as is
-  if (cleanPath.startsWith(`/${locale}/`)) {
+  // If path already contains locale as a segment, return as is
+  const pathSegments = cleanPath.split('/').filter(Boolean);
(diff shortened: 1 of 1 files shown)


mariahosfeld (changes requested):
While I no longer encounter the "Ressource not found" issue, the map is not really zooming out now, and is basically "stuck" on Africa. Tested in normal browser-window as well as incognito.

mohitb35:
> While I no longer encounter the "Ressource not found" issue, the map is not really zooming out now, and is basically "stuck" on Africa. Tested in normal browser-window as well as incognito.

That's a non related existing bug on the map. It seems like this happens when the project details page is directly opened. I've added this as an issue and will look into it soon.

mariahosfeld (approved):
LGTM, Issue originally reported needs to be dealt with separately.

PR #2411, opened 2025-02-20
Tax receipt/verify email

sunilsabatp:
This PR ensures that unauthorized users cannot access the receipt link or modify the contact form, displaying appropriate error messages.


mohitb35 (changes requested):
Currently the /verify-receipt-data route is always protected (even if user is not logged in).

It should be possible to access the route as a non-logged in user.

src/features/user/DonationReceipt/DonorContactManagement.tsx
```diff
@@ -30,10 +31,14 @@ type StoredReceiptData = {
 
 const DonorContactManagement = () => {
   const t = useTranslations('DonationReceipt');
-  const router = useRouter();
   const { donationReceiptData, updateDonationReceiptData } =
     useDonationReceipt();
   const { user, token, contextLoaded, logoutUser } = useUserProps();
+
+  const isEligibleForEdit = user?.email === donationReceiptData?.donor.email;
+  if (!isEligibleForEdit) return <EditPermissionDenied />;
```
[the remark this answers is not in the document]
mohitb35: The specific issue is resolved, but has created another issue.

The flow does not work if the user is already logged in when on the /verify-receipt-data page, as we are only depending on session storage data, which only gets set when the user is not logged in at first.

mohitb35 (changes requested):
@sunilsabatp Please review the feedback.

src/features/user/DonationReceipt/DonationReceiptLayout.tsx
```diff
@@ -63,6 +76,7 @@ const DonationReceiptLayout = () => {
                 dtn,
                 year,
                 challenge,
+                donorEmail: data.donor.email,
               })
             );
         }
```
mohitb35, line 82: When are we removing `receiptData` from sessionStorage?

sunilsabatp: After confirming the receipt data.

mohitb35 (approved):
LGTM

PR #2422, opened 2025-03-03
Release Donation Receipt Feature

mariahosfeld:
**Includes:**

- #2416 
- #2423 
- #2424


shyambhongle (approved):
Looks good to merge
@mariahosfeld

rupamkairi (approved):
Looks good to me as far seen.

rupamkairi:
Just few Notes.

1. https://dev.pp.eco/en/verify-receipt-data?dtn=F5ZY68H2&challenge=RUPAM&year=2025
This page opens without Authentication but cant do any Operations (Modify/ View All) - It asks for authentication - Did not tried if it redirects afterwards.

2. In that above flow I do not have the hamburger menu.

3. Hamburger menu on Mobile overlaps the back icon on Modify Contact info step.
Screenshot 2025-03-03 at 20 02 54

These can be looked later as well I believe.

mariahosfeld:
> Just few Notes.
> 
> 1. /verify-receipt-data
>    This page opens without Authentication but cant do any Operations (Modify/ View All) - It asks for authentication - Did not tried if it redirects afterwards.
> 2. In that above flow I do not have the hamburger menu.
> 3. Hamburger menu on Mobile overlaps the back icon on Modify Contact info step.
>    Screenshot 2025-03-03 at 20 02 54
> 4. On Smaller phone (this is from iPhoneSE in simulator) the Modal have some scroll issues, can't scroll all the way to save new address.
> 
> These can be looked later as well I believe.

@rupamkairi , thanks for the feedback.
Regarding point 1: That is the expected behavior if the user is not logged in. After login they will be redirecte to the intended pages.
Regarding the rest:
You are correct, mobile is something we will deal with in another PR.

PR #2435, opened 2025-03-10
Corrects typos in translation files

mohitb35:
Corrects keys/text with typos in the translation files, based on [CSpell](https://cspell.org/) warnings. The main focus are the english translation files located in "public/static/locales/en". Changes for other locales are merely to match key changes in the english files, spell check is not carried out in those files.

CSpell is currently configured in VS Code via extension.


sunilsabatp (approved):
LGTM

PR #2437, opened 2025-03-11
Fix: Export complex objects (geometry and metadata) as JSON strings in DataExplorer

mohitb35:
### Problem

- When exporting data to Excel, complex objects like geometry and metadata are appearing as null in the resultant Excel file, despite these fields containing valid data in the API response.

### Solution

- Modified the PostgreSQL query to convert complex objects to JSON strings at the database level using to_json()::text, ensuring that all data is properly exported to Excel files.

### Changes

- Updated SQL query in the export API endpoint to stringify JSON objects before they're returned
- Used PostgreSQL's native to_json() function with a text cast to convert complex objects to strings
- Added null handling with CASE statements to gracefully handle null values

### Performance
This approach is more efficient than client-side processing as it:

- Reduces data transfer size
- Leverages database optimization
- Eliminates additional JavaScript processing in the API layer


mariahosfeld (commented):
Functionality works as expected, data points are now included.

sunilsabatp (approved):

PR #2443, opened 2025-03-13
Fetch tree-registrations from intervention table instead of contribution table

mohitb35:
Changes:

- Refactors contributions procedure to handle registrations separately and fetch them from the interventions table instead.
- Creates separate types related to registrations RegistrationsQueryResult and adapts MyContributionsSingleRegistration to have property names specific to registrations instead of common names for both contributions and registrations
- Adapts downstream components as per the updated types and property names


mariahosfeld:
Only issue I see is the date with tree registrations, the rest is working as expected

mariahosfeld (commented):
With the exception of the date issue, everything looks good.

sunilsabatp (approved):

PR #2449, opened 2025-03-18
refactor: use useServerApi() hook for fetching project and plant loca…

sunilsabatp:
Utilized `useServerApi `hook to centralize API calls and eliminate the need to manually pass `locale, tenantID`.
Added support for GET, POST, PUT, and DELETE requests with both authenticated and non-authenticated versions.


sunilsabatp:
unwanted PR

PR #2456, opened 2025-03-20
Release: Switch source table for tree registrations

mariahosfeld:
**Includes:**

- #2443


jmiridis (approved):
fetching country for registration is still pending

PR #2457, opened 2025-03-21
Include country in fetchRegistrations query

mohitb35:
Update the fetchRegistrations query to include the country field (recently added to interventions table).


sunilsabatp (approved):

mariahosfeld (commented):

PR #2458, opened 2025-03-21
Fix: simplify and correct showContentContainer conditional logic

mohitb35:
This PR refactors the conditional logic for displaying the content container in the ProjectsLayout component and fixes a minor bug in the display logic where the content container was not hiding when the explore projects toggle was switched off.

Changes
- Fix a bug where the content container was not hiding when the explore projects toggle was switched off.
- Simplify the logic that determines visibility of the content container

The bug fix ensures that:
- For project-list pages: Content correctly hides when the projects are toggled off.
- Additionally maintains correct behavior for hiding content when embedded with explicit hide parameter (project_details=false on the project details page, and project_list=false on the project list page)


mariahosfeld (commented):

sunilsabatp (approved):

PR #2466, opened 2025-03-24
Feature/use server api hook

sunilsabatp:
- Utilize the `useApi()` hook for api call.

pages covered :

- `/complete-signup`
- `/claim/[type]/[code]`
- `/profile/api-key`
-  `/profile/delete-account`


mohitb35 (approved):
LGTM. Just update the description as the /all route is not affected by this PR.

PR #2469, opened 2025-03-25
Feature/use api hook1

sunilsabatp:
Utilize the useApi() hook for api call.
pages covered :

1. `/profile/donation-link`
2. `/profile/edit`
3. `/profile/history`

### DIFF
--- a/pages/sites/[slug]/[locale]/profile/history.tsx
+++ b/pages/sites/[slug]/[locale]/profile/history.tsx
@@ -32,6 +32,7 @@ import { defaultTenant } from '../../../../../tenant.config';
 import { useRouter } from 'next/router';
 import { useTenant } from '../../../../../src/features/common/Layout/TenantContext';
 import getMessagesForPage from '../../../../../src/utils/language/getMessagesForPage';
+import { useApi } from '../../../../../src/hooks/useApi';
 
 interface Props {
   pageProps: PageProps;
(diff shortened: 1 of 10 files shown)


pages/sites/[slug]/[locale]/profile/history.tsx
Comment (mohitb35) [inline pages/sites/[slug]/[locale]/profile/history.tsx:1]: While the approach for fetchPaymentHistory works, I feel that we should use the api handler queryParams instead of trying to create the end url itself. The code could also be clearer with a bit of cleaning up.

Perhaps something like below:

```
async function fetchPaymentHistory(next = false): Promise { setIsDataLoading(true); setProgress(70);

try { if (next && paymentHistory?._links?.next) { // Parse the next URL to extract query parameters const nextUrl = new URL(paymentHistory._links.next, window.location.origin); const path = nextUrl.pathname;

  // Extract query parameters from the URL
  const queryParams: Record<string, string> = {};
  nextUrl.searchParams.forEach((value, key) => {
    queryParams[key] = value;
  });
  
  const newPaymentHistory = await getApiAuthenticated<PaymentHistory>(path, {
    queryParams,
  });
  
  setPaymentHistory({
    ...paymentHistory,
    items: [...paymentHistory.items, ...newPaymentHistory.items],
    _links: newPaymentHistory._links,
  });
} else {
  // For initial load or filter change
  const queryParams: Record<string, string> = {
    limit: '15',
  };
  
  // If filter is selected, add it to query params
  if (filter) {
    queryParams.filter = filter;
  }
  
  const newPaymentHistory = await getApiAuthenticated<PaymentHistory>(
    '/app/paymentHistory',
    { queryParams }
  );
  
  setPaymentHistory(newPaymentHistory);
  setAccountingFilters(newPaymentHistory._filters);
}

setProgress(100);
setTimeout(() => {
  setProgress(0);
  setIsDataLoading(false);
}, 1000);
} catch (err) { setErrors(handleError(err as APIError)); redirect('/profile'); } }
```

mohitb35 (changes requested):

mohitb35 (approved):
LGTM

PR #2473, opened 2025-03-26
fix: restrict project fetch when registering a tree for non-TPO users

sunilsabatp:
This fix ensures that projects are only fetched when registering a tree if the user is a TPO (Tree Planting Organization). Previously, the endpoint was being called unnecessarily for non-TPO users. The logic has been updated to check the user type before triggering the project fetch, improving performance and reducing unnecessary API calls.

### DIFF
--- a/src/features/user/RegisterTrees/RegisterTreesWidget.tsx
+++ b/src/features/user/RegisterTrees/RegisterTreesWidget.tsx
@@ -235,7 +235,7 @@ function RegisterTreesForm({
     if (contextLoaded && user?.type === 'tpo') {
       loadProjects();
     }
-  }, [contextLoaded, user]);
+  }, [contextLoaded]);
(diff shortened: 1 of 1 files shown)


src/features/user/RegisterTrees/RegisterTreesWidget.tsx
```diff
@@ -232,10 +232,10 @@ function RegisterTreesForm({
   }
 
   React.useEffect(() => {
-    if (contextLoaded) {
+    if (contextLoaded && user?.type === 'tpo') {
       loadProjects();
     }
-  }, [contextLoaded]);
+  }, [contextLoaded, user]);
```
mohitb35, line 238: @sunilsabatp the API is called twice at times. Probably because of the dependencies.

PR #2480, opened 2025-04-03
Fixes non-working project grid navigation link (mangroves page)

mohitb35:
Change the link for project grid navigation so that it navigates (scrolls) correctly to the projects list.


mohitb35:
Closed as duplicating #2482

PR #2487, opened 2025-04-08
Add URL parameters for project filters

mohitb35:
- Introduce URL parameters to preselect project filters, allowing users to filter projects based on classification and donation status using url query params
- Update logic to automatically select the 'allProjects' tab when filters are applied.

Usage:
1. The following classification filters can be applied using query parameters to the index route: 'large-scale-planting', 'agroforestry', 'natural-regeneration', 'managed-regeneration', 'urban-planting', 'mangroves', 'other-planting' . 
    - e.g. `?filter=mangroves`
    - e.g. `?filter=agroforestry&filter=urban-planting`
2. Only donatable projects can be filtered out using `?donatable_projects_only=true`


sunilsabatp (approved):

mariahosfeld (approved):
LGTM.

PR #2490, opened 2025-04-10
Feature/use api hook8

sunilsabatp:
1. Applied consistent use of custom API hook across multiple features:

- embed-modal
- targets-modal (Forest Progress)
- support-pin
- upload image (Register Tree)
- subscription actions: cancel, pause, edit, reactivate

2. Defined and updated TypeScript types for API payloads
3. Improved type safety and maintainability of subscription-related code

### DIFF
--- a/src/features/common/types/payments.d.ts
+++ b/src/features/common/types/payments.d.ts
@@ -119,9 +119,14 @@ export interface PaymentHistory {
   _links: Links;
   _filters: Filters;
 }
-
+export interface Project {
+  id: string;
+  name: string;
+  amount: number;
+}
 export interface Subscription {
   id: string;
+  lastUpdate: string;
   totalDonated: number;
   amount: number;
   destination: Destination;
@@ -132,30 +137,17 @@ export interface Subscription {
   cycleType: string;
   frequency: string;
   currency: string;
-  startDate?: string | null;
+  startDate: string | null;
   currentPeriodStart: string;
   currentPeriodEnd: string;
-  trialPeriodStart: string;
-  trialPeriodEnd: string;
-  endsAt?: string | null;
+  trialPeriodStart: string | null;
+  trialPeriodEnd: string | null;
+  endsAt: string | null;
   status: string;
   isSynced: boolean;
-  pauseUntil?: string | null;
+  pauseUntil: string | null;
   donorName?: string;
   bankAccount: BankAccount | null;
-}
-
-export interface ModifyDonations
-  extends Omit<
-    Subscription,
-    | 'startDate'
-    | 'currentPeriodStart'
-    | 'currentPeriodEnd'
-    | 'trialPeriodStart'
-    | 'trialPeriodStart'
-    | 'endsAt'
-    | 'donorName'
-  > {
-  lastUpdate: string;
-  bankAccount: BankAccount;
+  project: Project;
+  type: string;
 }
--- a/src/features/user/Account/EditModal.tsx
+++ b/src/features/user/Account/EditModal.tsx
@@ -50,6 +50,13 @@ interface EditModalProps {
   fetchRecurrentDonations: (next?: boolean | undefined) => void;
 }
 
+interface ModifySubscriptionApiPayload {
+  nextBilling?: string | null;
+  centAmount?: number;
+  frequency?: string;
+  [key: string]: unknown;
+}
+
 type FormData = {
   amount: number;
   frequency: string;
@@ -86,13 +93,6 @@ export const EditModal = ({
     setDisabled(false);
   }, [editModalOpen]);
 
-  interface ModifySubscriptionApiPayload {
-    nextBilling?: string | null;
-    centAmount?: number;
-    frequency?: string;
-    [key: string]: unknown;
-  }
-
   const onSubmit = async (data: FormData) => {
     setDisabled(true);
     const payload: ModifySubscriptionApiPayload = {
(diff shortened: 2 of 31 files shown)


src/features/user/RegisterTrees/RegisterTrees/UploadImages.tsx
mohitb35, line 1: 1. Remove unused imports
2. I see `postAuthenticatedRequest` here, which PR replaces those?

src/features/user/Account/ReactivateModal.tsx
```diff
@@ -27,12 +25,11 @@ export const ReactivateModal = ({
...
 }: ReactivateModalProps) => {
   const [disabled, setDisabled] = React.useState(false);
-  const { tenantConfig } = useTenant();
+  const { putApiAuthenticated } = useApi();
   const { theme } = React.useContext(ThemeContext);
-  const { token, logoutUser } = useUserProps();
   const { setErrors } = React.useContext(ErrorHandlingContext);
   const t = useTranslations('Me');
-  const bodyToSend = {};
+  const payload = {};
```
mohitb35, line 32: Why do we send an empty payload?

src/features/user/Account/PauseModal.tsx
```diff
@@ -78,25 +80,26 @@ export const PauseModal = ({
...
     setDisabled(true);
-    const bodyToSend = {
-      pauseType:
-        option == 'pauseForMonth' || option == 'pauseUntilDate'
-          ? 'custom-date'
-          : 'infinite', //custom-date | infinite
+    const pauseType: PauseType =
+      option === 'pauseForMonth' || option === 'pauseUntilDate'
+        ? 'custom-date'
+        : 'infinite';
```
mohitb35, line 86: Why was this moved out? 

Perhaps we should follow the same approach for both payload properties?

src/features/user/Account/EditModal.tsx
```diff
@@ -99,15 +86,16 @@ export const EditModal = ({
     setDisabled(false);
   }, [editModalOpen]);
 
-  interface BodyToSendType {
+  interface ModifySubscriptionApiPayload {
     nextBilling?: string | null;
     centAmount?: number;
     frequency?: string;
+    [key: string]: unknown;
   }
```
mohitb35, line 94: Please move this out of the component.

src/features/common/types/payments.d.ts
```diff
@@ -132,30 +137,17 @@ export interface Subscription {
...
-    | 'startDate'
-    | 'currentPeriodStart'
-    | 'currentPeriodEnd'
-    | 'trialPeriodStart'
-    | 'trialPeriodStart'
-    | 'endsAt'
-    | 'donorName'
-  > {
-  lastUpdate: string;
-  bankAccount: BankAccount;
```
mohitb35, line 160: Why was this removed?

mohitb35 (changes requested):
@sunilsabatp Please review my feedback

src/features/user/RegisterTrees/RegisterTrees/UploadImages.tsx
sunilsabatp: The postAuthenticatedRequest part is handled in` #2492` .

src/features/user/Account/ReactivateModal.tsx
sunilsabatp: 1. The payload is not required for this request.
2. Since we have a strict type definition for PUT requests to accept a payload, we provide an empty object to avoid any type errors.

src/features/common/types/payments.d.ts
sunilsabatp: The following properties should be present in the `ModifyDonations `object, but they were intentionally omitted. However, if we include these properties — `startDate`, `currentPeriodStart`, `currentPeriodEnd`, `trialPeriodStart`, `trialPeriodEnd`, and `endsAt `— the structure becomes identical to the Subscription interface. Therefore, I have removed `ModifyDonations `and used the Subscription interface instead.

src/features/user/Account/EditModal.tsx
```diff
@@ -126,6 +126,7 @@ export const EditModal = ({
           queryParams: { scope: 'modify' },
           payload,
         });
+        console.log('res', res);
```
mohitb35, line 129: Resolve the unresolved TODO, and remove the console log

src/features/user/Account/EditModal.tsx
```diff
@@ -118,27 +106,29 @@ export const EditModal = ({
...
-          data: bodyToSend,
-          token,
-          logoutUser,
+        const res = await putApiAuthenticated<
+          Subscription,
+          ModifySubscriptionApiPayload
+        >(`/app/subscriptions/${record?.id}`, {
+          queryParams: { scope: 'modify' },
+          payload,
         });
```
mohitb35, line 128: There's a warning on line 132 as `response` does not exist within `res`

`window.open(res.response.confirmationUrl, '_blank');`

src/features/user/Account/ReactivateModal.tsx
mohitb35: I think we need to simplify the type structure defined in useApi.ts and apiClient.ts a bit. It has a few inconsistencies. But that is a separate matter.

mohitb35 (changes requested):
@sunilsabatp Please go through the feedback. Once these minor points are resolved, this is good to go.

mohitb35 (approved):
LGTM

PR #2492, opened 2025-04-10
Feature/use api hook9

sunilsabatp:
Refactor API integrations for improved consistency and type safety.

- Replaced `postRequest `and `postAuthenticatedRequest `with `postApi `and `postApiAuthenticated `hooks across multiple modules for consistent API handling.
- Defined and used specific payload types (e.g., `SpeciesApiPayload`, `RedeemCodeApiPayload`, `TreeMapperApiPayload`) to enhance type safety and code clarity.

PR #2491, opened 2025-04-10
Feature/use api hook 10

mohitb35:
Replaces `getAuthenticatedRequested` with `useApi().getApiAuthenticated`  in `PlanetCash` and `PlantingLocation` components

Pages affected
- /profile/planetcash
- /profile/treemapper/import (1st step)

PR #2479, opened 2025-04-02
Feature/use api hook5

sunilsabatp:
Utilized `useApi()`  hook for api calls

Affected pages :

- `/profile/redeem/[code]`
- `/s/[id]`
- `/t/[profile]`
- `/profile/treemapper`
- `/profile/treemapper/data-explorer`
- `/profile/treemapper/import`
- `/profile/treemapper/my-species`

- [x] Also Covered unauthenticated getRequest
- [x] Defined type for the projects(Analytics)

PR #2493, opened 2025-04-10
Release: use api hook migration part 2

mohitb35:
Includes:
- #2488 
- #2491 
- #2492 
- #2471 
- #2479
- #2490 
- #2494
- #2472


sunilsabatp (approved):
LGTM

mariahosfeld (changes requested):
Issue while testing:

- [ ] switching currency only takes effect after manual refresh (seen on main screen, tested via chrome and safari on mobile)

mariahosfeld (approved):
Looks good to me.

mariahosfeld:
> Issue while testing:
> 
> * [ ]  switching currency only takes effect after manual refresh (seen on main screen, tested via chrome and safari on mobile)

Will be handled in different release.

PR #2494, opened 2025-04-14
refactor(usePropContext):  handle user profile fetch with impersonati…

sunilsabatp:


src/features/common/Layout/UserPropsContext.tsx
```diff
@@ -98,36 +104,66 @@ export const UserPropsProvider: FC = ({ children }) => {
     logout({ returnTo: returnUrl });
   };
 
+  const fetchUserProfile = useCallback(
+    async (impersonationData?: ImpersonationData) => {
+      const header = {
+        'tenant-key': `${tenantConfig.id}`,
+        'X-SESSION-ID': await getsessionId(),
+        Authorization: `Bearer ${token}`,
```
mohitb35, line 112: Avoid using `await` within object creation. It would be better to keep this before composing the header object.

src/features/common/Layout/UserPropsContext.tsx
```diff
@@ -98,36 +104,66 @@ export const UserPropsProvider: FC = ({ children }) => {
...
+            method: 'GET',
+            headers: setHeaderForImpersonation(header, impersonationData),
+          }
+        );
+
+        if (!response.ok) {
+          const errorData = await response.json();
+          throw new APIError(response.status, errorData);
+        }
+        return response.status === 204 ? true : await response.json();
```
mohitb35, line 128: Why are we handling a 204 response status here? How can that scenario occur while fetching the user profile. 

Even if such a scenario occurs, should we not treat it as an error when user profile data is missing?

src/features/common/Layout/UserPropsContext.tsx
```diff
@@ -98,36 +104,66 @@ export const UserPropsProvider: FC = ({ children }) => {
...
+      } catch (err) {
+        console.error('Error during API call:', err);
+        throw err;
```
mohitb35, line 131: The function catches errors but then immediately logs and rethrows them. This pattern means errors are logged twice - once in fetchUserProfile and again in the calling functions. Consider either logging or rethrowing, but not both.

src/features/common/Layout/UserPropsContext.tsx
```diff
@@ -98,36 +104,66 @@ export const UserPropsProvider: FC = ({ children }) => {
...
+      }
+    },
+    [token, tenantConfig.id, locale, getsessionId]
```
mohitb35, line 134: Is getsessionId needed in the dependency array? It is an imported function and will not change.

src/features/common/Layout/UserPropsContext.tsx
```diff
@@ -98,36 +104,66 @@ export const UserPropsProvider: FC = ({ children }) => {
...
   };
 
+  const fetchUserProfile = useCallback(
+    async (impersonationData?: ImpersonationData) => {
+      const header = {
+        'tenant-key': `${tenantConfig.id}`,
+        'X-SESSION-ID': await getsessionId(),
+        Authorization: `Bearer ${token}`,
+        'x-locale': locale,
+      };
```
mohitb35, line 114: Explicitly define the return type for fetchUserProfile, and take care of any TS warnings that arise while doing so.

mohitb35 (changes requested):
@sunilsabatp Please review my feedback.

src/features/common/Layout/UserPropsContext.tsx
sunilsabatp: We usually won't see this scenario for the GET request, however I've handled the error scenario

src/features/common/Layout/UserPropsContext.tsx
sunilsabatp: I removed the try/catch block from the `fetchUserProfile()` function since the caller function already has a try/catch block. If an error occurs, it will propagate to the caller function, where the error is being handled.

sunilsabatp: I have removed the try-catch block from fetchUserProfile since the calling function handles all possible error scenarios in both places where it is used.

mohitb35 (approved):
LGTM

PR #2498, opened 2025-04-25
Refetch projects data on currency change

mohitb35:
Ensure that projects data is refetched when the currency code changes.

This fixes the bug where the user needed to refresh the page to see the updated currency (after selecting a different currency from the language/currency dialog).


mariahosfeld (commented):
While it is also there in the current deployment, is there a way to get rid of the UI behavior on the map (after switch no projects pins at first, then pins appear and slight zoom happens)? Otherwise currency switch now works.

mohitb35:
> While it is also there in the current deployment, is there a way to get rid of the UI behavior on the map (after switch no projects pins at first, then pins appear and slight zoom happens)? Otherwise currency switch now works.

Let's tackle that separately, as the issue occurs with language change as well.

The projects-archive pages do not have this issue, so we may have a ready solution.

sunilsabatp (approved):

PR #2507, opened 2025-04-30
Add locale + tenant as a query parameter to /app/projects API calls

mohitb35:
Include locale in API calls to prevent caching issues when the locale changes, ensuring the correct response is fetched based on the current locale.

Includes tenant in API calls to prevent browser side caching issues (mostly while testing out different tenants in the same browser over a short period).


mariahosfeld (commented):
Functionality works. Loading time with switch a bit longer than before, but this trade of for seeing the correct language should be fine.

mohitb35:
@mariahosfeld Have updated this PR, this should resolve the project data caching issue.

sunilsabatp:
@mohitb35  When I opened the deployment link, I encountered the error below. Could you please address this?
image

sunilsabatp (approved):

PR #2509, opened 2025-05-02
Correctly initialize TenantProvider to avoid default tenant usage

mohitb35:
Initialize the TenantProvider with the correct tenant configuration to ensure the appropriate tenant-key header is used in API calls.

To tenant: visit https://planet-app-sf.herokuapp.com/en/home directly (check network tab, correct tenant key should be passed in header). Compare with https://weareams.web.plant-for-the-planet.org/en/home.


mariahosfeld (commented):

sunilsabatp (approved):

Issue #2503, opened 2025-04-29
Improve UI behavior when switching language/currency.


mariahosfeld:
While it is also there in the current deployment, is there a way to get rid of the UI behavior on the map (after switch no projects pins at first, then pins appear and slight zoom happens)? Otherwise currency switch now works.

_Originally posted by @mariahosfeld in https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/2498#pullrequestreview-2793513605_

PR #2515, opened 2025-05-09
fix: prevent marker flicker by removing loading check before renderin…

sunilsabatp:
- This PR addresses the issue where project markers briefly disappear when switching languages or re-fetching projects.
- The root cause was a conditional `isLoading `check in `MultipleProjectsView `that prevented marker rendering during data fetch.

[Issue](https://github.com/Plant-for-the-Planet-org/planet-webapp/issues/2503)

NOTE: Test the PR only on the landing page; other pages may break due to mismatches in the tenant header configuration.


mariahosfeld (changes requested):
Tested the following scenarios:
1. change of language via URL -> as this is reload, "flicker" will not go away
2. change of language via pop-up

Needed changes:
- [x] grey layer behind the pop-up needs to be hidden (?) after changing the language, currently it stays as an opaque layer on top
image

sunilsabatp:
> Tested the following scenarios:
> 
> 1. change of language via URL -> as this is reload, "flicker" will not go away
> 2. change of language via pop-up
> 
> Needed changes:
> 
> * [x]  grey layer behind the pop-up needs to be hidden (?) after changing the language, currently it stays as an opaque layer on top
>   image

Couldn't reproduce it.

mohitb35 (approved):

mariahosfeld (approved):

Issue #2468, opened 2025-03-25
Navbar Profile Image dimension


sunilsabatp:
The profile image in the navigation bar has incorrect dimensions. It appears stretched or improperly sized compared to the other elements in the navbar.

Is it on a Production or Development Build?
[x] Both 

To Reproduce
Go to the homepage.
Look at the profile image in the navbar.
Observe that the image is not displayed with the correct aspect ratio or size.

Expected behavior
The profile image should be displayed with the correct dimensions and aspect ratio, ensuring it aligns properly with the other navbar elements.

What actually happens
The image appears misaligned or incorrectly sized compared to other elements.

Screenshot 
<img width="832" alt="Image" src="https://github.com/user-attachments/assets/8486ed88-6e4f-4be0-a912-d2fd50e9c90f" />

Additional context
This issue might be caused by incorrect CSS styles applied to the profile image, such as width, height, or object-fit. It should be checked in the styles applied to the image component.

PR #2530, opened 2025-05-21
fix: update img styles to include width for consistency; adjust for x…

sunilsabatp:
[Issue](https://github.com/Plant-for-the-Planet-org/planet-webapp/issues/2468)


mariahosfeld (commented):
Test showed correct dimensions.

PR #2535, opened 2025-05-26
Updates external website url handling and validation

mohitb35:
1. Update the handling of website URLs to ensure they are recognized as external links
2. Enforce 'http(s)' protocol in the validation regex for the Edit Profile and Basic Details forms.


sunilsabatp:
The error message is a bit generic. It would be more helpful to explicitly instruct the user to add `https://` as a prefix to the URL if the user does not provide it.
image

mohitb35:
> The error message is a bit generic. It would be more helpful to explicitly instruct the user to add `https://` as a prefix to the URL if the user does not provide it. image

Fixed

sunilsabatp (approved):

PR #2536, opened 2025-05-28
Ensure Map-Dependent Views Render Only After Map Load Completion

sunilsabatp:
This PR addresses a rendering issue where the map would occasionally:

- Zoom out to the maximum level unexpectedly after the first render
- Display low-resolution tiles
- Show floating or misaligned markers before the map had fully loaded

Changes made:

- Introduced mapLoaded state using useState
- Used the `onLoad `event of the `<Map />` component to set `mapLoaded `to `true `once the map is fully initialized
- Updated conditions for rendering `SingleProjectsView `and `MultipleProjectsView `to wait until `mapLoaded `is `true`


sunilsabatp:
redundant PR

PR #2555, opened 2025-06-17
refactor: simplify measurement rendering by using latest values inste…

sunilsabatp:
This PR refactors the logic for rendering plant location measurements:

Why ?
- The history array was often empty or unavailable, leading to no data being shown. Using the  available values from `location.measurements `ensures consistency and simplifies the UI logic.


mariahosfeld (commented):
Functionality looks good.

PR #2564, opened 2025-06-20
Use centralized SetState type from common.d.ts

sunilsabatp:
This PR removes the duplicate `SetState `type definition from` src/features/common/Layout/BulkCodeContext.tsx` and replaces it with the canonical definition from` src/features/common/types/common.d.ts.`

[Issue](https://github.com/Plant-for-the-Planet-org/planet-webapp/issues/2550)


mohitb35 (approved):
LGTM

PR #2566, opened 2025-06-23
Rationalize colors for snackbar

mohitb35:
Align snackbar colors with the design system and eliminate redundant snackbar state management in the CompleteSignup and EmbedModal components.


mariahosfeld (approved):
LGTM!

PR #2579, opened 2025-06-27
Fix validation logic in DonationReceiptContext for donor/address changes

mohitb35:
Update the validation logic to recalculate the `isValid` state when donor or address changes, ensuring the Confirm button in the verify receipt screen appears correctly after correcting invalid contact information.


mariahosfeld (commented):
LGTM!

Issue #2367, opened 2025-01-15
To consider: handle unselection of a point location when an other intervention with Point geometry is selected?


mohitb35:
Do we not want to handle unselection of a point location when an intervention with Point geometry is selected? Similar to how single trees allow clicking on them again to unselect the point?

e.g. XZ9L5P (type = maintenance)

_Originally posted by @mohitb35 in https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/2343#discussion_r1914402232_

mohitb35:
Context - it is currently possible to click on a single tree registration or a multi tree registration with point geometry to unselect them (if already selected i.e. we are clicking again on the same selected plant location with a point geometry).

Do we want to expand this behaviour to other types of interventions, when point geometry is present.

mohitb35:
Context - it is currently possible to click on a single tree registration or a multi tree registration with point geometry to unselect them (if already selected i.e. we are clicking again on the same selected plant location with a point geometry).

Do we want to expand this behaviour to other types of interventions, when point geometry is present.

PR #2603, opened 2025-07-14
feat: support deselecting other point interventions on map click

sunilsabatp:
[Issue](https://github.com/Plant-for-the-Planet-org/planet-webapp/issues/2367)


mariahosfeld (changes requested):
While testing, I was not able to get the project information back by clicking on the map if it was a single tree location. Same for other point locations.

sunilsabatp:
Duplicate PR

Issue #2355, opened 2025-01-09
Refactor dropdown open logic (intervention and site dropdown)


mohitb35:
The logic to keep only one menu open is unclear, partly due to using terms like `enable` and `disable` instead of `open`/`close`/`toggle`. The names `disableInterventionMenu`, `enableInterventionFilter`, `disableInterventionFilter` need rethinking.

Please revisit this logic across the `ProjectSiteDropdown`, `InterventionDropdown`, `ProjectsContext` and `MapControls`

_Originally posted by @mohitb35 in https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/2343#discussion_r1903794488_

PR #2604, opened 2025-07-14
Hotfix/site intervention dropdown

sunilsabatp:
[Issue](https://github.com/Plant-for-the-Planet-org/planet-webapp/issues/2355)


mariahosfeld (changes requested):
The intervention dropdown now contains all types, but it should only show the ones that were recorded by the projects. If the project did not record a fence, this option should not show up in the dropdown. Please revert this change.

mariahosfeld (commented):
Functionality does not seem to be affected by changes.

PR #2605, opened 2025-07-16
Multi Plant Location - update area display logic and improve type safety

mohitb35:
1. Remove dependency on sample trees for displaying plant location area 
2. Improve type safety in the OtherInterventions interface
3. Adds stronger typing for PlantLocationHeader and MultiPlantLocationInfo
4. Refactors code to remove unnecessary conditions.

### DIFF
--- a/public/static/locales/de/projectDetails.json
+++ b/public/static/locales/de/projectDetails.json
@@ -30,7 +30,7 @@
     "progressReports": "Fortschrittsberichte",
     "readMore": "Mehr",
     "readLess": "Weniger",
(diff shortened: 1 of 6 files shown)


sunilsabatp:
project Slug -> `ke-whcr`
The species lists are not rendering because the API returns `scientificName ` as null
image

mohitb35:
> project Slug -> `ke-whcr` The species lists are not rendering because the API returns `scientificName ` as null image

Yes, aware of that. Checking to see if `otherSpecies` is an acceptable fallback. However, that is for a different PR.

PR #2624, opened 2025-07-28
Fixes partially hidden mobile swiper for intervention info

mohitb35:
Adds an offset from bottom equal to footer height while positioning MobileInfoSwiper, to avoid the footer partially hiding the swiper.

Before:
Before

After:
After


mariahosfeld (changes requested):
Name of the picture is a bit squished between bottom of picture and the index of the pictures. Please add some padding/distance here. Otherwise looks good.

mariahosfeld (approved):
LGTM!

PR #2625, opened 2025-07-29
Fix layout issues in image slider descriptions

mohitb35:
Address layout inconsistencies for image descriptions in the image slider across various components, improving both mobile and non-mobile views.


mariahosfeld (commented):
LGTM!

PR #2667, opened 2025-08-21
Hotfix/zoom to project site validation

sunilsabatp:
[Fixes](https://github.com/Plant-for-the-Planet-org/planet-webapp/issues/2652)

Context :
- `zoomInToProjectSite `directly accessed `geoJson.features[selectedSite] `without validating index or geometry, which could lead to runtime errors.
- `getFeaturesAtPoint `queried all `INTERACTIVE_LAYERS`, including ones not yet added to the map, causing` layer does not exist `errors.
image

Solutions :
- Added guard clause in `zoomInToProjectSite `to validate `selectedSite `and ensure geometry exists.
- Updated `getFeaturesAtPoint `to filter only layers that exist in the map before querying features.

Result :
- Prevents runtime crashes due to invalid indices or missing geometry.
- Eliminates errors when querying non-existent layers.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -39,6 +39,7 @@
     "MAXAR",
     "MODIS",
     "MONTHNAME",
+    "moveend",
     "multiday",
     "multipolygons",
     "Oceanforce",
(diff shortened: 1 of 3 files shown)


mohitb35 (approved):
LGTM

PR #2693, opened 2025-09-05
fix: standardize casing for modal state setters in Recurrency components

sunilsabatp:
This PR standardizes the casing and naming convention of modal state setter functions across the Recurrency components.


mohitb35 (approved):
LGTM

PR #2694, opened 2025-09-09
refactor: replace hardcoded colors with design system colors

sunilsabatp:
- Removed hardcoded color values from the projectV2.
- Replaced them with corresponding design system color tokens for consistency.


mohitb35 (approved):
LGTM

PR #2698, opened 2025-09-16
Refactor router usage and optimize localizedPath function

mohitb35:
Improve performance by optimizing router dependency usage across multiple components and enhance the localizedPath function with useCallback.

### DIFF
--- a/src/features/common/Layout/UserLayout/UserLayout.tsx
+++ b/src/features/common/Layout/UserLayout/UserLayout.tsx
@@ -299,7 +299,7 @@ const UserLayout = ({ children }: { children: ReactNode }) => {
         router.push(localizedPath('/login'));
       }
     }
-  }, [contextLoaded, user, router]);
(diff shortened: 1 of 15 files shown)


src/features/common/Layout/UserLayout/UserLayout.tsx
```diff
@@ -299,7 +299,7 @@ const UserLayout = ({ children }: { children: ReactNode }) => {
         router.push(localizedPath('/login'));
       }
     }
-  }, [contextLoaded, user, router]);
+  }, [contextLoaded, user]);
```
sunilsabatp, line 302: I think `router.asPath`  needs to be added as a dependency here. Is there any specific reason you haven’t added it?

mohitb35: This is supposed to run just once, not every time the path changes.

sunilsabatp (approved):
LGTM !

PR #2740, opened 2025-10-23
fix: add null checks in addMissingMonths to prevent undefined errors

sunilsabatp:
[Resolves](https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/2725#issuecomment-3432854690)

### Issue
`TypeError: Cannot read properties of undefined (reading 'month')` in `addMissingMonths` function when processing empty or malformed data.

### Changes
- Added null/undefined checks before accessing `firstData.month` and `lastData.month`
- Return empty array for invalid data 
- Added warning logs for debugging

### Result
Component now handles invalid data gracefully by rendering an empty chart instead of crashing.


mohitb35 (approved):
LGTM

PR #2750, opened 2025-11-03
fix: safely access giftDetails.giverName using optional chaining

sunilsabatp:
Added optional chaining operator to `giftDetails?.giverName` to prevent` null/undefined `reference errors and updates type for SingleGiftReceived. This ensures the component gracefully handles cases where giftDetails or giverName may not exist by falling back to the anonymousUser translation


mohitb35 (approved):
LGTM

PR #2751, opened 2025-11-04
Refactor contribution statistics calculation

mohitb35:
Optimize the calculation of contribution statistics in the ContributionStats component by utilizing projectListResult. This change deprecates unnecessary properties from the API response and simplifies the response transformation in the useMyForestApi hook.


sunilsabatp (approved):
LGTM

PR #2783, opened 2025-12-03
Fix progress calculation for tree projects

mohitb35:
1. Update the progress calculation to utilize `unitsContributed` and `unitsTargeted` for tree projects, ensuring accurate progress representation. `countPlanted` will be deprecated is no longer used here.
2. Adds missing styles for progress bars for `regular`/`nonDonatable` projects

- Adjusted progress calculation logic for tree projects.
- Updated dependencies in the `useMemo` hook for better performance.
- Adds missing progress bar styles so that bar is visible for regular projects


mariahosfeld (approved):
LGTM!

PR #2809, opened 2025-12-23
Normalize registration GeoJSON to Point geometry for clustering

sunilsabatp:
This PR normalizes registration GeoJSON to always use `Point `geometry. Since both clustered and single marker views are rendered from the `Supercluster response`—and Supercluster ignores polygon geometries—`registration entries with polygon geometry were not rendered earlier.`

Polygon-based registration locations are now converted to a representative point, fixing missing registration markers in both clustered and single marker views.


mohitb35 (approved):
LGTM

PR #2850, opened 2026-01-24
Make subtitle prop optional in DashboardView component

### LINKED ISSUE #1758: DashboardView - make subtitle prop optional
It seems a little odd to need to explicitly define subtitle as `null`.This could be made optional. 

_Originally posted by @mohitb35 in https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/1712#discussion_r1170804136_

mohitb35:
The subtitle prop in the DashboardView component has been made optional, allowing for more flexible usage. Several instances of the DashboardView component have been updated to reflect this change by removing the subtitle prop where it was previously set to null.


sunilsabatp (approved):
LGTM

PR #2857, opened 2026-01-30
Feature/migrate project use context to zustand store

### LINKED ISSUE #2980: Project details page crashes for projects whose sites have null geometry (unguarded turf area())
**Describe the bug**

The project details page throws an uncaught client-side exception and fails to render for any project whose site has `geometry: null`. Turf's `area()` is called on such a site without a guard, so it throws `TypeError: Cannot read properties of null (reading 'type')`, which bubbles up as a Next.js "A client-side exception has occurred" and blanks the page.

This is a pre-existing bug (reproduces on `develop`); it is not caused by the Zustand migration in #2857. It was found while testing that PR.

Both. Reproduced on a production build (Vercel preview) and present in `develop`.

**To Reproduce**

1. Open a project whose only site has `geometry: null`, e.g. https://app-staging.plant-for-the-planet.org/en/koenigswald-ontananga
2. Open the browser console.
3. The page does not render; console shows `TypeError: Cannot read properties of null (reading 'type')` originating from `ProjectSiteDropDown/index.tsx` (the `siteList` `useMemo`), followed by `A client-side exception has occurred`.

Affected staging projects (4, all `purpose=trees`, single site with null geometry):
- `koenigswald-ontananga`
- `reversing-deforestation-for-smallholder-farmers`
- `reforestation-and-environmental-education`
- `trees-for-peace-an-ayudh-europe-campaign`

**Expected behavior**

The project page renders. Sites without geometry contribute `0` area (or are skipped) rather than crashing the page.

**What actually happens**

`area(site)` throws on the null geometry and the exception is uncaught, so the whole project details page fails to render.

**Root cause**

`src/features/projectsV2/ProjectsMap/ProjectSiteDropDown/index.tsx` (the `siteList` `useMemo`):

```ts
return projectSites.map((site, index: number) => {
  const calculatedSiteAreaInM2 = area(site); // throws when site.geometry === null
  ...
});
```

`area()` reads `geometry.type`; when `geometry` is `null` it throws. There is no try/catch or null guard here, so the error is uncaught and crashes the page.

The same turf-on-null-geometry pattern exists in two more places, but those are wrapped in try/catch, so they only log rather than crash:
- `src/features/projectsV2/ProjectDetails/components/ProjectInfo.tsx` (`projectAreaInHectares`) — logs `"Error calculating project area in hectares:"`.
- `src/features/projectsV2/ProjectDetails/components/MultiTreeInfo.tsx` — uses `area` similarly.

**Suggested fix**

Guard the geometry before calling `area()`, e.g. `site.geometry ? area(site) : 0`, in `ProjectSiteDropDown`. Apply the same guard in `ProjectInfo` and `MultiTreeInfo` for consistency, and consider centralizing a small `safeArea(feature)` helper.

**Desktop**
- OS: macOS
- Browser: Chrome

**Additional context**

Not a blocker for #2857 (the crashing code is unchanged by that PR); it was simply found while testing it.

sunilsabatp:
Overview
This PR migrates project-related state management from React Context to Zustand stores. It removes prop drilling, improves performance, and centralizes state and URL synchronization logic across the application.

## Store Architecture

Introduced four Zustand stores:

  - `projectStore`: Manages project list data, filters, and search state
  - `singleProjectStore`: Handles individual project details, site selection, and URL synchronization
  - `interventionStore`: Manages intervention selection, sample trees, hover state, and map interactions
  - `viewStore`: Centralizes UI-level state such as current page and selected view mode

viewStore enhancements:

- Moved `page` and `selectedMode` state into the store to centralize UI state
- Added initialization hooks to automatically derive page state from routes

##  State Management Improvements

Eliminated prop drilling by consuming state directly from Zustand stores

-   Centralized URL synchronization logic with dedicated store actions:
-   `selectSiteAndSyncUrl` – Site selection with URL updates
-   `selectInterventionSyncUrl` – Intervention selection with URL updates
-   `updateProjectDetailsPath` – Consistent shallow routing behavior

Added cleanup actions to ensure proper state lifecycle handling:

-   `clearFilterStates` – Reset search and classification filters
-   `clearInterventionSelectionAndHover` – Clear map interaction state
-   `clearProjectStates` – Reset project-level state on navigation

## Initialization Logic

- `useInitializeSingleProject` – Bootstraps project view from URL params (site)
-  `useInitializeIntervention` – Handles intervention selection (ploc) from URL with safe fallbacks
-  `useInitializeView` – Automatically determines page state based on the current route

## Component Refactoring

- Updated components to consume Zustand stores instead of React Context:
          `SingleProjectView`, `ProjectsMap`, `ProjectDetails`, `ProjectSiteList`, `InterventionDropdown`, `SitesDropdown`, `ProjectSnippet`, `ProjectInfo`, `SingleTreeInfo`

- Map layers and interaction handlers

## Helper Utilities

-  `getSiteIdFromIndex` / `getSiteIndexById` – Safe site resolution helpers
-  `getInterventionByHid` – Intervention lookup by hid
-   `isString` – Type guard for query params

## Bug Fixes & Edge Cases

Prevented site zoom when only an intervention is requested via URL

- Fixed map-to-list switching on mobile by syncing site selection
- Resolved multi-project view `flicker` during navigation
- Handled edge case where the URL requests both a site and an intervention ->Site selection is now prioritized to  prevent the map from freezing

##  Cleanup

  - Removed `ProjectContext` and `ProjectsProvider` entirely
  - Removed redundant props across the component tree
  - Improved `Zustand devtools` action naming for easier debugging

✨ Benefits
  - Better performance – Direct store access and fewer re-renders
  - Improved maintainability – Centralized logic with clear store boundaries
  - Enhanced debugging – Action-level tracing via Zustand devtools
  - Improved type safety – Consistent types across stores and consumers


mohitb35:
Heads-up: this PR reimplements project search in the new `useFilteredProjects` store selector using the pre-existing normalization (NFD strip + lowercase + plain `.includes()`). Since then, #2970 (approved and merged) substantially reworked search in `ProjectsContext.getSearchProjects`:

- hyphen → space normalization (`Plant-for-Ghana` ↔ `Plant for Ghana`)
- stroke/ligature folding NFD misses (`ø ł đ ð ß æ œ þ ı`)
- German digraph cross-matching (`München` ↔ `Muenchen` ↔ `Munchen`)
- a guard so a separator-only keyword (`-`, `---`) returns no results

When #2970 merges first, please port `normalizeForSearch` / `LATIN_FOLD` / `expandGermanDigraphs` / `normalizeDigraph` / the `fieldMatches` helper into the migrated selector — otherwise the migration silently reverts all of it. cc @sunilsabatp

src/hooks/useInitializeProject.ts
```diff
@@ -0,0 +1,79 @@
...
+        _scope: 'map',
+        currency: currencyCode,
+        //passing locale/tenant as a query param to break cache when locale changes,
+        //as the browser uses the cached response even though the x-locale header is different
+        locale,
+        tenant: tenantId,
+        'filter[purpose]': 'trees,conservation',
+      },
+    });
+  }, [currencyCode, locale, tenantId, currentPage]);
```
mohitb35, line 53: There are 2 regressions here:

1. `/app/projects` is called to fetch projects no matter which page is opened initially.
2. There can be 2 fetches on the list page (e.g. when the initial currency is not EUR). The calls race, resulting in the possibility that the list can show EUR prices when the selector says INR.

src/stores/singleProjectStore.ts
```diff
@@ -0,0 +1,236 @@
...
+          set(
+            {
+              fetchError: true,
+              isFetching: false,
+              singleProject: null,
+            },
+            undefined,
+            'singleProjectStore/project_fetch_error'
+          );
+        }
```
mohitb35, line 153: While an error is seen when the fetch fails (e.g. if the project is not found), we no longer redirect, and a permanent skeleton is seen. 

Old `loadProject` catch did `setErrors(...)` **and** `router.push(localizedPath('/'))` (confirmed at old `src/features/projectsV2/ProjectDetails/index.tsx:125-127`). The new store sets `fetchError: true, singleProject: null` and fires the toast but does **not** redirect, and **nothing consumes `fetchError`**. `ProjectDetails` renders `<Skeleton>` whenever `singleProject === null`. So a bad slug, a non-tree/conservation project (the store throws `ClientError(404)`), or a network error now leaves the user stuck on an infinite skeleton instead of being sent home.

We should either restore the redirect or add an error state that consumes `fetchError`

src/hooks/useInitializeSingleProject.ts
```diff
@@ -0,0 +1,86 @@
...
+    if (hasOnlyRequestedIntervention) return;
+
+    let siteIndex: number | null = null;
+    // If site is provided in the URL, try to resolve it
+    if (isString(requestedSite) && hasProjectSites) {
+      const index = getSiteIndexById(projectSites, requestedSite);
+      siteIndex = index !== -1 ? index : FIRST_SITE_INDEX;
+    } else {
+      // Default to first site (or null if no sites exist)
+      siteIndex = hasProjectSites ? FIRST_SITE_INDEX : null;
```
mohitb35, line 71: Minor (Low): `hasProjectSites` only checks `projectSites.length > 0` and ignores geometry. The old context used `hasNoSites = length === 0 || sites.every(s => s.geometry === null)` and selected no site in that case; here a project whose sites all have null geometry gets site 0 selected (and a ?site= for a geometry-less site). 

Impact is small — the map zoom is masked because getSitesGeoJson filters null-geometry sites out — so not a blocker. 

Worth reintroducing a shared `hasNoSites` helper and reusing it here, in `useInitializeIntervention`, and in exitMapMode. Note: currently unobservable on the affected projects because they hit a separate pre-existing crash first (#2980)

src/features/projectsV2/ProjectsMap/MapControls.tsx
```diff
@@ -84,72 +80,40 @@ const MapControls = ({
...
+    setSelectedMode('list');
+    // Mobile-only behavior: switching from map to list on project details
+    // resets intervention state, selects the first available site (or null),
+    // and syncs the site via shallow routing.
+    selectSiteAndSyncUrl(
+      hasProjectSites ? FIRST_SITE_INDEX : null,
+      locale,
+      router
+    );
   };
```
mohitb35, line 103: Old `exitMapMode` only did `setSelectedMode('list')` and left the selected site untouched. New version calls `selectSiteAndSyncUrl(hasProjectSites ? FIRST_SITE_INDEX : null, …)`, so a mobile map→list switch always snaps back to site 0, discarding the site the user was viewing.

src/features/projectsV2/ProjectDetails/components/ProjectInfo.tsx
```diff
@@ -129,7 +128,7 @@ const ProjectInfo = ({
   }, [metadata]);
 
   const location = useMemo(() => {
-    if (!tpo.address) return '';
+    if (!tpo) return '';
     const { address, zipCode, city, country } = tpo.address;
```
mohitb35: Weakened guard in `ProjectInfo` can throw on missing TPO address

Old code guarded `if (!tpo.address) return '';` before destructuring. New code guards only `if (!tpo) return '';` then destructures `tpo.address`. 

If the API omits `address` (the original guard implies it can), this throws "Cannot destructure property 'address' of undefined".

mohitb35 (approved):
@sunilsabatp

Mostly looked good. I came across and fixed some minor issues. 

Good to merge now.

PR #2873, opened 2026-02-11
Feature/migration usecontext to zustand planet cash

sunilsabatp:
This PR migrates PlanetCash state management from `React Context` to `Zustand` and refactors related components for cleaner logic and better maintainability.

## What’s Included

- Replaced `PlanetCashContext `with a new `planetCashStore (Zustand)`
- Migrated all PlanetCash components to use usePlanetCashStore
- Removed PlanetCash provider and context
- Renamed accounts to `planetCashAccounts `for clarity
- Added `useInitializePlanetCash` hook to reset store on profile change / impersonation
- Simplified fetch logic (early returns, moved cleanup to finally)
- Simplified conditional rendering in `Accounts` and `Transactions`
- Reduced `prop drilling` by reading values directly from the store
- Optimized store selectors to prevent unnecessary re-renders

## Impact

- No functional changes
- Cleaner architecture
- Improved readability
- Better state management consistency


mariahosfeld (commented):
Functionality on the Client Side seems to be as usual, so this does not impact the UX, looks good to me.

src/hooks/useInitializePlanetCash.ts
```diff
@@ -0,0 +1,18 @@
+import { useEffect } from 'react';
+import { usePlanetCashStore, useUserStore } from '../stores';
+
+/**
+ * Hook to handle PlanetCash store lifecycle.
+ * Resets PlanetCash state when the user profile changes.
+ */
```
mohitb35, line 7: Small thing on the doc block. The first line does not add anything the hook name already says, and the second restates the three lines below it.

The part a reader cannot work out from the code is which situations actually change `profileId`. It reads like it is only about switching profiles, but it also covers first load, login, and logout.

```suggestion
/**
 * PlanetCash data belongs to a single user profile, so the store is re-initialized whenever that profile changes.
 * `profileId` changes on first load, on login, on logout, and when entering or leaving impersonation.
 */
```

src/hooks/useInitializePlanetCash.ts
```diff
@@ -0,0 +1,18 @@
...
+export const useInitializePlanetCash = () => {
+  const profileId = useUserStore((state) => state.userProfile?.id);
+  const resetPlanetCashStore = usePlanetCashStore(
+    (state) => state.resetPlanetCashStore
+  );
+
+  useEffect(() => {
+    // Reset PlanetCash store on impersonation mode active
```
mohitb35: This names one trigger out of four. The effect also runs on first load, on login, and on logout, so impersonation is the least of what it covers. The doc block above is the better place for it, so I would drop this line.

```suggestion
```

src/features/user/PlanetCash/index.tsx
```diff
@@ -83,30 +94,33 @@ export default function PlanetCash({
...
-          '/app/planetCash'
-        );
-        redirectIfNeeded(accounts);
-        const sortedAccounts = sortAccountsByActive(accounts);
-        setIsPlanetCashActive(accounts.some((account) => account.isActive));
-        setAccounts(sortedAccounts);
-      } catch (err) {
-        setErrors(handleError(err as APIError));
-      }
+  const fetchAccounts = async () => {
```
mohitb35, line 97: `fetchAccounts` lost its `useCallback` here. Looks safe to me, since the effect below depends only on `isAuthReady` and never on the callback, so nothing changes. Was that deliberate?

src/features/user/PlanetCash/index.tsx
```diff
@@ -83,30 +94,33 @@ export default function PlanetCash({
...
+    // If accounts already exist, just handle redirects
```
mohitb35: "already exist" reads as "the user has accounts", but the check passes for an empty array too, so what it really means is "we already fetched".

```suggestion
    // If accounts were already fetched, just handle redirects
```

src/features/user/PlanetCash/screens/Transactions.tsx
```diff
@@ -34,6 +36,10 @@ const Transactions = ({
   const [isDataLoading, setIsDataLoading] = useState(false);
   const [isModalOpen, setIsModalOpen] = useState(false);
   // store: state
+  const hasPlanetCashAccount = usePlanetCashStore(
+    (state) =>
+      state.planetCashAccounts !== null && state.planetCashAccounts.length > 0
+  );
```
mohitb35, line 42: Nice change, this stops the list refetching when an account object is replaced. Worth stating what it changes though.

The effect that calls `fetchTransactions` used to depend on the accounts array itself, so it refetched whenever that array was replaced, including by `updateAccount` from `TopUpManagement`. It now depends on this boolean, so it only refetches when the user goes from having no accounts to having at least one, or back the other way.

I could not find a way to observe the difference today, since `TopUpManagement` renders on the Accounts tab and this is a separate tab. Was that the intent?

mohitb35 (commented):
Reviewed the refreshed diff after the develop merge. The migration is faithful: I checked the new store against the deleted context value by value, and the consumers, the init timing, and the render branches all match. Nothing here blocks merge.

Four small comments below. Two are comment wording, two are just clarifying intent.

Separately, I found three pre-existing problems with how the account list is fetched. None of them are from this PR, `develop` behaves the same way, so please don't fix them here. They are tracked in #3083, including the double request you may have noticed in local dev.

mohitb35 (approved):
Approving. The migration is faithful, and I checked the new store against the deleted context value by value, plus the consumers, the init timing, and the render branches. CI is green across the board.

Two small things before you merge, both with one-click suggestions on the diff:

- [x] - the doc block on `useInitializePlanetCash`
- [x] - the "already exist" comment in `fetchAccounts`

And two questions that are just for the record, not blockers: the `useCallback` removal on `fetchAccounts`, and the narrowed refetch trigger in `Transactions`. I believe both are safe, I would just like to know they were deliberate.

Nothing else here is yours to fix. Everything substantive I found predates this PR, including the double request in local dev and the CodeRabbit note about the create-account fallback. All of it is tracked in #3083, with the four symptoms and a suggested implementation.

Thanks for keeping this one self-contained. It made the migration easy to verify against the old context.

src/features/user/PlanetCash/screens/Accounts.tsx
```diff
@@ -2,30 +2,32 @@ import type { ReactElement } from 'react';
...
+      <>
+        {planetCashAccounts.map((account, index) => (
+          <AccountDetails account={account} key={index} />
+        ))}
+      </>
+    );
+  }
+
+  // This will never be seen, as the user is being redirected to create a new PCA when none exists
+  return <NoPlanetCashAccount />;
```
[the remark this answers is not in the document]
mohitb35: Real observation, and it is worse than cosmetic: `NoPlanetCashAccount` carries a live "create PlanetCash" button that pushes to `/new`, and since `useEffect` runs after paint, that frame does get painted. So an account owner can briefly be offered the create flow.

Not from this PR though. `develop` has the identical branch, so it is pre-existing.

I would not take the proposed patch as-is. Returning the loader whenever the list is `null` also covers the failed-request case, which turns an API failure into a spinner that never stops. That swaps one wrong state for another.

The underlying problem is that `Accounts` cannot tell "no data yet" from "no accounts" from "the request failed", because all three are `null`. Tracked as symptom 4 in #3083, along with the fix: a `status` field on the store, where `idle` covers this pre-request render and `error` is distinct from an empty list.

src/features/user/PlanetCash/index.tsx
sunilsabatp: Yes, this was deliberate. `fetchAccounts` isn’t used as a dependency or passed anywhere that relies on `referential stability`, so wrapping it in `useCallback` doesn’t provide any benefit here. The effect still only runs when `isAuthReady` changes, so removing `useCallback` doesn’t change the behavior.

src/features/user/PlanetCash/screens/Transactions.tsx
sunilsabatp: Yes, that was the intent  to avoid refetching transactions when an existing account object is updated, since only the presence of an account matters here.

PR #2874, opened 2026-02-11
Feature/migration usecontext to zustand payout

sunilsabatp:
This PR migrates the Manage Payouts module from the existing context-based state management to a dedicated `Zustand` store.

## Changes included:

- Added `useManagePayoutStore` (Zustand store)
- Migrated `accounts` and `payoutMinAmounts` from `usePayouts` to the new store
- Added derived selectors (`hasAccounts`, `hasPayoutMinAmounts`) to prevent redundant API calls
- Simplified fetch logic using early returns
- Prevented unnecessary prop drilling
- Simplified JSX rendering logic in the Overview component
- Removed redundant payout wrapper and old context after migration

## Result

- Cleaner state management
- Better separation of concerns
- Reduced redundant API calls
- No functional changes


src/stores/useManagePayout.ts
mohitb35: Store filename breaks the naming convention

Every other store in `src/stores` is `<feature>Store.ts`: `planetCashStore.ts`, `currencyStore.ts`, `userStore.ts`, `myForestStore.ts`.

This one is named after a hook and drops the `Store` suffix. The interface `ManagePayoutStore` and the export `useManagePayoutStore` already follow the convention, so only the filename is out of step.
Suggested name: `managePayoutStore.ts`.

mohitb35 (commented):
Two things need fixing before this merges. Read the third section before you write the fix.

## 1. Payout accounts are never cleared when the profile changes

Nothing resets this store. There is no reset action and no caller for one, so `accounts` stays filled when you switch profile.

What happens: impersonate TPO-A and open payouts. Exit, impersonate TPO-B, open payouts. The guard in `fetchAccounts` sees A's data and skips the fetch. B's session then shows A's IBAN, holder name and BIC.

This is not from your PR. The old `PayoutsContext` had no reset either. On #2873 we agreed to leave pre-existing problems alone and file them instead. This one is different: it shows bank details under the wrong profile. And since the develop sync, the fix is close to a copy of `useInitializePlanetCash`, which now sits in the same folder.

## 2. `hasAccounts` treats an empty list as "never fetched"

The old guard was `if (!accounts)`. It fetched only when the value was `null`. An empty array is truthy, so it did not fetch again.

`accounts !== null && accounts.length > 0` is false for `null` and for `[]`. So a TPO with no bank accounts fetches `/app/accounts` on every mount. Overview, Schedule and Add Bank Details are separate routes, so that is every tab click. It happens exactly when the list is empty, which is during setup.

## 3. Why the obvious fix will not work

Adding the reset hook on its own does nothing useful, and it will look like it worked.

`StoreInitializer` sits above the page in `_app`, so its effect runs first and clears `accounts`. The effect that calls `fetchAccounts` runs next, but it reads `hasAccounts` from the render before the reset. So it still sees `true` and returns early. The store is now empty and nothing fetches.

`userId` in the deps does not help, because `userId` changed in that same render.

PlanetCash has this exact problem today. It is problem 2 in #3083.

## Suggested change

Four edits. They only work together, so please do not ship 1 and 2 without 4.

- [x] **1.** `src/stores/managePayoutStore.ts`: add an `initialState` object and a `resetManagePayoutStore` action, the same way `planetCashStore.ts` does it.
- [x] **2.** Add `src/hooks/useInitializePayout.ts`, close to a copy of `useInitializePlanetCash`, and call it from `StoreInitializer` after `useInitializePlanetCash()`.
- [x] **3.** Replace the `hasAccounts` selector with `const hasFetchedAccounts = useManagePayoutStore((state) => state.accounts !== null);` The name matters. The guard answers "have we fetched", not "are there any". The `?.` on the next line is dead after the null check, and goes away with this edit.
- [x] **4.** Add it to the deps of the effect that calls `fetchAccounts`: `[isAuthReady, userId, hasFetchedAccounts]`.

Step 4 is what makes the reset work. After a reset the flag turns `false`, the effect runs again, the guard passes, and the new profile's accounts load. It cannot loop. Once the fetch lands the flag is `true` again and the next run returns early.

Worth checking once it is in: switch profiles with the Network tab open and confirm a fresh `GET /app/accounts` fires.

## Out of scope for this PR, but should be tackled in a quick followup

- [x] `Overview` shows the "add bank details" button before the first request starts. CodeRabbit flagged this too. Treating `null` as loading there trades a wrong empty state for a spinner that never stops when the request fails. It is pre-existing, PlanetCash has the same problem, and a proper fix needs the request status on the store. That belongs in a separate followup PR, #3084 is created to track this. 

The rest of the PR looks good. The screens behave the same as before, `AddBankAccount` appends to the list instead of replacing it, and `key={account.id}` in `Overview` is a real improvement.

mohitb35 (changes requested):
@sunilsabatp Please go through the [feedback above](https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/2874#pullrequestreview-5005080314).

src/features/user/ManagePayouts/screens/EditBankAccount.tsx
```diff
@@ -142,7 +147,6 @@ const EditBankAccount = (): ReactElement | null => {
       </FormHeader>
       <BankDetailsForm
         account={accountToEdit}
-        payoutMinAmounts={payoutMinAmounts}
         isProcessing={isProcessing}
         handleSave={handleSaveAccount}
       />
```
mohitb35, line 152: nit: 

It might be worth adding `key={accountToEdit.id}` while rendering `BankDetailsForm` in `EditBankAccount` - just to force it to unmount and mount a new form if the account changes.

This is not possible at this point, just a latent weakness. Adding a key would make this issue impossible rather than just unreachable.

mohitb35 (approved):
[Feedback](https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/2874#pullrequestreview-5005080314) addressed.

#3084, #3085 issues do not block merge.

src/features/user/ManagePayouts/screens/Overview.tsx
```diff
@@ -3,25 +3,25 @@ import type { ReactElement } from 'react';
...
 const Overview = ({ isDataLoading }: Props): ReactElement | null => {
-  const { accounts } = usePayouts();
+  const accounts = useManagePayoutStore((state) => state.accounts);
 
-  return isDataLoading ? (
-    <BankAccountLoader />
-  ) : accounts && accounts.length > 0 ? (
+  if (isDataLoading) return <BankAccountLoader />;
+
+  if (!accounts || accounts.length === 0) return <NoBankAccount />;
```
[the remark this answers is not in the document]
mohitb35: #3084 tracks this, pre-existing issue.

PR #2878, opened 2026-02-26
Refactor bulk-codes module: migrate from Context API to Zustand store

sunilsabatp:
This PR refactors the bulk-codes module by replacing the Context API with a centralized Zustand store (`BulkCodeStore`) to improve state management, readability, and maintainability.

Key Changes:

- Introduced BulkCodeStore (Zustand) for centralized state management.
- Migrated all bulk-code related pages and components from `useBulkCode` context to `useBulkCodeStore`.
- Moved project fetching and filtering logic into the store.
- Removed unnecessary prop drilling (e.g.,`projectList`, `planetCashAccount`).
- Added derived selectors such as `isBulkMethodSet` for cleaner validation.
- Simplified validation logic using guard clauses.
- Removed redundant state (`bulkGiftData`).
- Added `resetBulkStore` action for proper cleanup.
- Removed BulkCode context provider completely.

Benefits

- Cleaner and more maintainable architecture
- Reduced prop drilling
- Centralized business logic
- Improved readability and separation of concerns
- Easier debugging with Zustand devtools


mariahosfeld (commented):
Functionality looks good to me, tested generic codes as well as custom certificates. All functions work as before, so the UX has not changed with this update.

src/features/user/BulkCodes/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+  if (!Array.isArray(projects) || projects.length === 0) {
+    return [];
+  }
+
+  const allowedCHFProjects = ['yucatan'];
+
+  return projects.filter((project) => {
+    const isValidUnitCost = project.unitCost > 0;
+    const isAllowedPurpose =
+      project.purpose === 'trees' || project.purpose === 'conservation';
```
mohitb35, line 29: **Priority: high.** Already fixed on the branch, no action needed. Posted for the reasoning.

I pushed a fix for this one to the branch, so this is only the reasoning.

The filter as submitted tested `classification`, but the rule in develop tests `purpose`. That test was changed deliberately in #2888, which narrowed bulk codes to restoration and conservation projects only. So moving the filter into `utils.ts` brought back the rule that #2888 had removed.

The effect was that funds projects became selectable again. Their classifications are `academy`, `endowment`, `funding`, `membership` and `sponsorship`, so the three that are neither `membership` nor `endowment` passed the new test. Those projects also have `unitType` of `currency`, while `IssueCodesForm` multiplies units by `unitCost` and presents the result as a count of units.

`filterEligibleProjects` now uses the `purpose` test again, and the JSDoc cites #2888 so the reason stays next to the code.

The CHF check and the `unitCost` check both moved across correctly and I left them alone. Your simplification of the CHF branch is right, the repeated `currency === 'CHF'` in develop was redundant.

src/features/user/BulkCodes/index.tsx
```diff
@@ -74,43 +68,9 @@ export default function BulkCodes({
...
-        setErrors(handleError(err as APIError));
-      }
-    }
-  }, [planetCashAccount?.currency, locale]);
-
   useEffect(() => {
-    fetchProjectList();
-  }, [fetchProjectList]);
+    fetchProjectList(getApi);
+  }, [fetchProjectList, getApi]);
```
mohitb35: **Priority: medium.**

Two things about these dependencies.

`fetchProjectList` is a store action so it never changes, but `getApi` is a fresh function on every render, because `useApi` does not wrap anything in `useCallback`. So this effect runs on every render of `BulkCodes`.

That is also the only reason the fetch happens at all. `planetCashAccount` is not in the list, and `fetchProjectList` returns early while the account is null. Develop tracked the account explicitly, its `useCallback` had `planetCashAccount?.currency`. So if `useApi` ever wraps its functions in `useCallback`, which is a normal thing to do, the project list will quietly stop loading and nothing here will say why.

The second part is in the store. The guard in `fetchProjectList` is `projectList !== null`, and `projectList` stays null until the request resolves, so every render inside that window starts another request for the same country. `currencyStore` already handles this with an `isFetching` flag that it sets before the await and clears in a `finally`.

Could we add `planetCashAccount` to these dependencies and copy the `isFetching` guard into `bulkCodeStore`? Then the fetch happens because the code says so, rather than because a dependency happens to be unstable.

src/stores/bulkCodeStore.ts
```diff
@@ -0,0 +1,103 @@
...
+        }
+      },
+
+      setProject: (project) => set({ project }),
+
+      setBulkMethod: (bulkMethod) => set({ bulkMethod }),
+
+      setPlanetCashAccount: (planetCashAccount) => set({ planetCashAccount }),
+
+      resetBulkStore: () =>
```
mohitb35: **Priority: medium.**

`resetBulkStore` clears `bulkMethod` and `project`, and `IssueCodesForm` calls it after a successful issue. Nothing clears `planetCashAccount` or `projectList`.

Those two outlive a profile switch. `BulkCodes` only sets the account when it is still null, so once it is set it is never corrected. After entering impersonation, the impersonated user sees the previous profile's country project list and prices, and `paymentOptions` is queried with the previous country.

There is no money risk here, I checked that. The donation payload carries no account id, so the server picks the account from the impersonated user.

2873 and 2874 both added an initialize hook for exactly this, `useInitializePlanetCash` and `useInitializePayout`, each resetting its store when the profile id changes. Could we add `useInitializeBulkCode` in the same shape and register it in `StoreInitializer`?

src/stores/bulkCodeStore.ts
```diff
@@ -0,0 +1,103 @@
...
+  setBulkMethod: (bulkMethod: BulkCodeMethods | null) => void;
+  setPlanetCashAccount: (planetCashAccount: PlanetCashAccount | null) => void;
+  setProject: (project: CountryProject | null) => void;
+  resetBulkStore: () => void;
+}
+
+export const useBulkCodeStore = create<BulkCodeStore>()(
+  devtools(
+    (set, get) => ({
+      bulkMethod: null,
```
mohitb35: **Priority: low.**

A few small consistency points against `planetCashStore` and `managePayoutStore`.

**1. No `initialState` object.**
Both siblings declare `const initialState = {...}`, spread it with `...initialState`, and reset with `set(initialState, ...)`. This store lists its initial values here and then repeats a subset of them inside `resetBulkStore`. Using `initialState` would also keep the reset complete on its own, so nobody has to remember to add new fields to it later.

**2. No devtools action names.**
None of the `set` calls in this file pass an action name, so devtools shows them all as anonymous. `currencyStore` names each one, for example `currencyStore/fetch_start`.

**3. `resetBulkStore` name.**
Reads as if the store were called `bulkStore`. `resetBulkCodeStore` would match `resetPlanetCashStore` and `resetManagePayoutStore`.

src/features/user/BulkCodes/components/ProjectSelector.tsx
```diff
@@ -12,21 +11,19 @@ import {
...
   useErrorHandlingStore,
 } from '../../../../stores';
+import { useBulkCodeStore } from '../../../../stores/bulkCodeStore';
 import { useTranslations } from 'next-intl';
 
 interface ProjectSelectorProps {
-  projectList: CountryProject[];
   project: CountryProject | null;
   setProject?: (project: CountryProject | null) => void;
+  active?: boolean;
```
mohitb35: **Priority: low.**

Could this line come out? `active` used to be the prop here, and `disabled` replaced it in 7ef69257e, which renamed it across `ProjectSelector`, `ProjectSelectAutocomplete` and `IssueCodesForm`. So the two are the same switch, one older name and one current.

Nothing reads `active` in this component and neither caller passes it. `IssueCodesForm` passes `disabled={true}` and `SelectProjectForm` passes neither, so `disabled` is doing the whole job.

The reason I would rather remove it than leave it as a harmless extra: it now sits directly above the prop that replaced it, so the next person could reasonably wire `active` up and end up with two flags for one behaviour.

Same question as on the project filter really. That rule was also changed deliberately, in #2888, and this PR brought back the earlier version of it too. Both times the older version reappeared in a file that already had the newer one on this branch. Is it possible parts of these files were rebuilt from an older copy rather than edited in place? If so, worth a quick look for anything else that came along with them.

src/stores/bulkCodeStore.ts
```diff
@@ -0,0 +1,103 @@
...
+export interface Recipient {
+  units: number;
+  recipientName: string;
+  recipientEmail: string;
+  message: string;
+  notifyRecipient: boolean;
+  // occasion: string;
+}
+
+export interface BulkGiftImportData {
```
mohitb35: **Priority: low.**

`BulkGiftImportData` was only needed by `bulkGiftData`, which this PR removes, and nothing imports it now. Safe to drop.

`Recipient` above it is still used, `IssueCodesForm` imports it, so that one stays.

pages/sites/[slug]/[locale]/profile/bulk-codes/[method]/[id].tsx
```diff
@@ -41,76 +41,73 @@ export default function BulkCodeIssueCodesPage(): ReactElement {
...
   //store: state
   const isAuthReady = useAuthStore((state) =>
     Boolean(state.token && state.isAuthResolved)
   );
   const userProfile = useUserStore((state) => state.userProfile);
+  const isBulkMethodSet = useBulkCodeStore(
+    (state) => state.bulkMethod !== null
+  );
+  const projectList = useBulkCodeStore((state) => state.projectList);
+  const project = useBulkCodeStore((state) => state.project);
```
Comment (mohitb35) [inline pages/sites/[slug]/[locale]/profile/bulk-codes/[method]/[id].tsx:51]: **Priority: low.**

`project` is read here but never used, which is the one new lint warning on the branch.

It is left over from the `if (!project)` guard that this rewrite removed. Without the guard, `checkContext` fetches `paymentOptions` again whenever its dependencies change, even when the project is already set. In practice they settle after the first run, so it costs nothing today, but the guard was what said so.

Either drop the variable or put the guard back. Putting it back is the smaller change and keeps this file closer to develop.

src/features/common/Layout/UserLayout/UserLayout.tsx
```diff
@@ -278,14 +278,9 @@ const UserLayout = ({ children }: { children: ReactNode }) => {
...
 
-  // While auth state is resolving
-  if (!isAuthResolved) {
-    return <UserProfileLoader />;
-  }
-
   //Auth resolved but no user (redirect in effect)
   if (!userProfile) {
-    return null;
+    return <UserProfileLoader />;
```
mohitb35: **Priority: high.** This is the main reason for requesting changes. A question first, so no change until you have answered it.

Could you say what this change is for? I would like to check I am reading it right before suggesting anything.

My guess is that it is the workaround for the flicker in #3044, where a profile refetch flipped `isAuthResolved` back to false and remounted the user area, which wiped the bulk codes success screen. If that is it, the cause is now fixed in develop. #3055 moved ownership of the flag so it is set once and never reset, so a refetch no longer flickers.

The reason I am asking rather than suggesting a revert is that the two conditions are not interchangeable, and dropping the `isAuthResolved` branch has a side effect that is visible to users.

#3055 also changed when the flag flips. It is now set in the `finally` of the profile fetch, so it becomes true after `userProfile` has already been set. That leaves a window of at least one render where the profile exists and the flag is still false. Develop's `!isAuthResolved` branch holds the loader through that window. Without it, the page renders its children there instead.

That is what causes a flash on a bad bulk codes link. Opening `/en/profile/bulk-codes/lemony` shows the select project form once or twice before the redirect lands, where develop shows nothing in between. The page renders the step and only redirects from an effect, so there is always a gap. This change is what makes the gap visible.

Worth knowing that the window applies to every screen inside `UserLayout`, not only bulk codes, so any screen that paints before its own guards have settled would show the same way.

So, two questions. Is the flicker still happening for you after #3055? And if it is not, can we drop this hunk here, so the shared layout goes back to develop's behaviour?

If you do want the spinner rather than the blank, that seems reasonable to me, but it would be clearer as its own change with the `isAuthResolved` branch kept, so it gets reviewed as a UX decision on its own.

pages/sites/[slug]/[locale]/profile/bulk-codes/[method]/index.tsx
```diff
@@ -25,33 +24,37 @@ import {
...
+
+      const isValidMethod =
+        method === BulkCodeMethods.GENERIC || method === BulkCodeMethods.IMPORT;
+
+      isValidMethod
+        ? setBulkMethod(method)
+        : router.push(localizedPath('/profile/bulk-codes'));
     }
-  }, []);
+  }, [router.isReady]);
```
Comment (mohitb35) [inline pages/sites/[slug]/[locale]/profile/bulk-codes/[method]/index.tsx]: **Priority: high.** Same issue as the `UserLayout` question, at the page level.

This follows on from the `UserLayout` question. Dropping that hunk removes the flash, but the gap in this page stays, and it does not depend on layout timing, so it is worth closing either way.

The page renders the step and only redirects from an effect, so a bad method paints the select project form first. Deciding validity during render closes it:

```js
const { method } = router.query;
const isValidMethod =
  method === BulkCodeMethods.GENERIC || method === BulkCodeMethods.IMPORT;

useEffect(() => {
  if (!router.isReady) return;
  if (!isValidMethod) {
    router.push(localizedPath('/profile/bulk-codes'));
    return;
  }
  if (!isBulkMethodSet) setBulkMethod(method);
}, [router.isReady, isValidMethod, method]);

if (!isInitialized || !router.isReady || !isValidMethod) return <></>;
```

Two smaller things this also settles. It adds the `router.query.method` dependency CodeRabbit asked about. And it replaces the ternary, which reads a little oddly for two statements that are not values.

Adding `[router.isReady]` here was the right call by the way. Develop passed an empty array, so this only worked because the route is `fallback: 'blocking'` and the query is ready on the first render.

`[id].tsx` has the same shape, an unconditional step with `checkContext` redirecting from an effect, so an invalid method or a project that is not found should flash the issue codes form in the same way. Worth a quick test there too.

mohitb35 (changes requested):
Thanks, this reads well and the port itself is faithful. I checked all six context values one by one. The four that moved are correct, and both drops are right, nothing reads `bulkGiftData` or `totalUnits` any more. Removing the prop drilling from `ProjectSelector` is a clear improvement, and moving the fetch into the store follows what `currencyStore` and `singleProjectStore` already do.

I pushed one fix to the branch, on the project filter, and left the reasoning in a comment on that file.

**Each comment starts with a priority so you can work through them in order.**

- **High.** Two comments, and they are the reason for requesting changes. A bad bulk codes link, for example `/en/profile/bulk-codes/lemony`, flashes the select project form once or twice before redirecting. Develop shows nothing in between. I traced it to the `UserLayout` change rather than to anything in bulk codes, so that comment asks about the intent first rather than assuming, because the answer may change what we do. The third high comment is the filter fix I already pushed, kept only for the reasoning.
- **Medium.** Two comments, both on the store. The project list fetch works today, but for a reason the code does not state, and it can start several requests for the same country. And nothing resets the store when the profile changes, unlike the two stores from 2873 and 2874.
- **Low.** Five comments. Store conventions, a re-added prop, two dead or duplicated exports, and one unused variable. Take or leave any of these.
- **For information.** One comment, pointing at #3087, which I filed for a pre-existing type naming problem. Nothing to do here.

Two of them, the project filter and the `active` prop, are cases where an earlier deliberate change came back. There is a question in the `ProjectSelector` comment about whether parts of these files were rebuilt from an older copy. That question matters more than either fix.

Two of CodeRabbit's three points are already handled, the empty project list in f503981 and the router readiness check. The remaining one about dependency lists is partly addressed and the rest is pre-existing, so nothing is owed there.

Nothing blocks on the type or lint baselines. `tsc` stays at 126 and there is one new lint warning, which is the unused `project` in `[id].tsx`.

src/stores/bulkCodeStore.ts
```diff
@@ -0,0 +1,103 @@
...
+import type { BulkCodeMethods } from '../utils/constants/bulkCodeConstants';
+import type { ApiConfigBase } from '../hooks/useApi';
+
+import { create } from 'zustand';
+import { devtools } from 'zustand/middleware';
+import { useErrorHandlingStore } from './errorHandlingStore';
+import { handleError } from '@planet-sdk/common';
+import { filterEligibleProjects } from '../features/user/BulkCodes/utils';
+
+export type ApiRequestFn = <T>(
```
mohitb35: **Priority: low.**

`myForestStore` already declares an `ApiRequestFn` with this same shape, so `src/stores` now has two exported types with one name.

The other four fetching stores, `currencyStore`, `projectStore`, `singleProjectStore` and `interventionStore`, inline the signature in their action type instead of naming it.

I think the named alias reads better than inlining it four more times, so my suggestion is to keep one copy somewhere shared and import it in both stores, rather than to inline it here.

src/stores/bulkCodeStore.ts
```diff
@@ -0,0 +1,103 @@
...
+  url: string,
+  config?: ApiConfigBase
+) => Promise<T>;
+
+export interface PlanetCashAccount {
```
mohitb35, line 16: **Priority: none, for information only.** Filed as #3087.

No action needed in this PR, this is a pointer so the thread exists.

This `PlanetCashAccount` shares its name with a different type in `features/common/types/planetcash.d.ts`, the full account with `balance`, `creditLimit` and the rest. `Recipient` below has the same problem against the upload row type in `BulkCodesTypes.d.ts`, which is why `IssueCodesForm` imports one of them under an alias.

Neither clash is new, `BulkCodeContext` declared both types too, so this is not something the migration introduced. Filed as #3087 for a follow-up.

src/features/common/Layout/UserLayout/UserLayout.tsx
sunilsabatp: I checked this after #3055, and I am no longer able to reproduce the flicker issue. It looks like the original problem has already been resolved by the isAuthResolved changes.

In that case, I agree that we should keep the develop behaviour and remove this workaround change rather than keeping the extra branch. Thanks for pointing out the side effect with the bulk codes flow.

mohitb35 (approved):
LGTM

PR #2906, opened 2026-04-07
Feature/restor api integration

shyambhongle:
Fixes #

Restor API integration

## Changes in this pull request:
- Integration of Restor API

@mohitb35

PR #2910, opened 2026-04-13
Release fix RESTOR and deactivate register trees

mariahosfeld:
**Includes:**

- #2906

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -164,6 +164,11 @@ const nextConfig = {
         destination: 'https://www.plant-for-the-planet.org/de/tagdesbaumes/',
         permanent: true,
       },
+      {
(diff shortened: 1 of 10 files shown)


mohitb35 (approved):
LGTM

mohitb35:
Some comments (non blocking)

1. A newly created site could not be synced (I got a 400 error).
en/profile/projects/proj_ygwN4yizPNwaQa218Gy5g0qe?type=project-sites
After investigation, realized this is a conservation project - perhaps the sync button should not be shown here at all?

2. It may also be an improvement if we don’t consider sites that were already synced as an error scenario. 

3. It may be worth letting the Restor team know - we get “SITE_AREA_TOO_SMALL” errors for extremely large sites.

PR #2935, opened 2026-06-03
TreeMapper API integration

shyambhongle:
Intervnetion APIs are replaced with new TreeMapper Backend APIs.


mariahosfeld (dismissed):
- [x] At least on staging I need more than one try to open the Yucatan-project via URL (at least on first load of platform)
- [x] No TreeMapper polygons are shown and interventions endpoint returns a 404

mariahosfeld (commented):
Functionality looks good to me now.

mariahosfeld (approved):

PR #2954, opened 2026-06-18
fix(auth): route email_not_verified denials to verify-email

sagararyal:
The Auth0 PostLogin Action now denies unverified logins with 'email_not_verified' instead of '401'. Match the new message so the verify-email routing (Navbar) and the login wait-state still trigger, instead of falling through to a generic alert and logout loop.


mariahosfeld:
Pre-merge we most likely need to take care about this issue globally in the systems: https://github.com/Plant-for-the-Planet-org/planet-donations/pull/583#issuecomment-4747641805

mohitb35 (approved):
LGTM

PR #2957, opened 2026-06-20
Map performance: GPU markers, drop the second WebGL engine, virtualize the list

sagararyal:
## What

A performance pass over the projects map, plus a few correctness fixes uncovered along the way. Headline: the explore map no longer ships a second WebGL engine, renders its markers on the GPU, and virtualizes the project list.

### Map performance
- **Drop the redundant `mapbox-gl` engine.** The explore route loaded *both* `maplibre-gl` and `mapbox-gl` (~532 KB gzipped) because `SingleMarker` and the profile `ContributionsMap` imported from the bare `react-map-gl-v7` entry (which pulls mapbox-gl) instead of `react-map-gl-v7/maplibre`. Swapped all of them, then removed the now-dead deps — `mapbox-gl`, `react-map-gl@5`, `deck.gl`, `luma.gl`, `@mapbox/mapbox-gl-draw`, `mapbox-gl-compare` (none were imported). `mapbox-gl` is fully evicted from installs.
- **GPU marker layer.** The explore map rendered ~259 projects as HTML `<Marker>` DOM nodes, each reprojected on the JS main thread on every pan/zoom frame (plus a per-marker drop-shadow filter). Replaced with one maplibre GeoJSON source + symbol layer. The custom per-category pins are **preserved** (rasterized to map images via `map.addImage`) and the drop-shadow is baked into the texture. Reprojection now runs in the GPU vertex shader; scales to ~1M points.
- **Virtualize the project list** with `react-virtuoso` — only the ~visible cards mount and recycle on scroll, instead of all ~326 at once.
- **Lazy-load project card images** so off-screen cards don't fetch eagerly.
- **Remove the stale basemap sprite reference.** The style pointed at a vendored ESRI sprite; the sprite reference has been removed from `root.json` and the vendored files deleted. The basemap renders without POI icons — acceptable given the app's primary markers are the project markers.

### Fixes
- **Redis degrades gracefully** when `REDIS_URL`/`REDIS_TOKEN` are unset (logs + runs without cache) instead of throwing at module load, which was crashing the multi-tenancy middleware and every tenant page in local dev. Production still hard-fails on a real misconfiguration.
- **Interventions overlay** failures no longer surface a generic "check your internet" popup and bounce the user to the homepage — it's an optional map overlay, so it now logs and renders the project without it. A guard for missing `TREEMAPPER_URL` is also added.
- **Marker hover popup flicker** — the popup opened over the pin/cursor when it rendered above the marker, triggering a mouseleave/reopen loop; it is now offset clear of the pin and stays open while hovered.
- **Remove the `theme.ts` top-level await** that made the theme module (and every importer) async — webpack `topLevelAwait` warning, a dev full-reload on every theme change, and a config fetch blocking module load. Tenant fonts now come synchronously from the tenant store.
- **Storybook loses all CSS variables after theme refactor** — `preview.js` used `React.lazy` to import the theme as a workaround for the async module; after the refactor the default export became a function that was never called. Fixed by a direct synchronous `getGlobalStyles()` call.
- **Last project card hidden behind bottom nav** in virtualized list — `padding-bottom: 65px` on the Virtuoso root had no effect on its internal scroll container. Fixed with a `Footer` spacer component.

### DX / config
- Default `TREEMAPPER_URL` to `dash.treemapper.app` for local dev (the interventions overlay endpoint).
- Commit `.claude/launch.json` (dev-server presets) and gitignore per-user Claude files.

## Notes for review
- ~The large additions are mostly the **vendored basemap sprite JSON/PNGs** under `public/data/styles/sprites/`. This copies an ESRI basemap asset — please confirm it fits the ArcGIS basemap terms (the app already uses the ESRI basemap and shows Map Credits). Alternative if preferred: repoint the style to the live `basemaps.arcgis.com` sprite URL rather than vendoring it.~
- The Redis change keeps the production guard (throws on a real misconfig in prod); only local/non-prod degrades.

## Testing
Verified each change live (against staging, logged in): `mapbox-gl` no longer loads on the explore route; markers render as the GPU layer with the correct custom pins + shadow and a stable hover popup; the list virtualizes (~5 cards mounted, recycles on scroll); no `topLevelAwait` warning. Suggest an `ANALYZE=true npm run build` before/after for the bundle-size delta.


mohitb35 (approved):
LGTM

PR #2965, opened 2026-07-06
refactor: clean up legacy react-map-gl type definitions

### LINKED ISSUE #2960: Remove dead types from map.d.ts
## Context

After removing the v5 `react-map-gl` package in the map performance refactor, a large block of types in `src/features/common/types/map.d.ts` is no longer imported anywhere in the codebase. All `ViewState` and `MapStyle` consumers now import directly from `react-map-gl-v7/maplibre`. The dead code was pre-existing but is now more visible.

## Types to remove

- `ClusterMarker`
- `MarkerProps`
- `ClusterMarkerProps`
- `SingleMarkerProps`
- `CustomPopupMarkerProps`
- `TestPointProps`
- `TestClusterProps`
- `InteractionState`
- `ViewState`
- Both `ViewportProps` declarations
- `MapStyle`, `RequiredMapStyle`, and the full `MapStyle` sub-tree: `Sources`, `Esri`, `Layer`, `Paint`, `FillTranslate`, `Layout`, `Metadata`, `ArcgisQuickEditor`, `Colors`, `ArcgisEditorExtent`, `SpatialReference`

## Steps

1. Verify no consumers remain (grep for each type name)
2. Delete the dead declarations from `map.d.ts`
3. If the file becomes empty, delete it entirely

sunilsabatp:
Resolves #2960 
## What & Why

After the `react-map-gl` v5 removal and migration to `react-map-gl-v7/maplibre`, several legacy map-related type definitions in `src/features/common/types/map.d.ts` were no longer being used anywhere in the codebase.

These types were previously retained for compatibility with the old map implementation, but all current consumers now import the required types directly from `react-map-gl-v7/maplibre`. Keeping the unused declarations adds unnecessary maintenance overhead and can make it harder to understand which types are actually in use.

This PR removes the dead type definitions.

## Impact

- No functional changes.
- Reduces dead code and maintenance overhead.
- Makes map type usage more explicit by relying on the official types from `react-map-gl-v7/maplibre`.
- Improves codebase clarity following the map library migration.


mohitb35 (approved):
LGTM

PR #2970, opened 2026-07-09
fix(projects): accent-, hyphen- and German-digraph-insensitive project search

sagararyal:
## What

Makes project search forgiving of accents, hyphens, and German umlaut spellings, so a project is found regardless of how the user types it.

### Before
`getSearchProjects` stripped diacritics + lowercased, but treated hyphens literally and missed letters Unicode NFD doesn't decompose. So these failed:
- "Plant for Ghana" did not match **Plant-for-Ghana**
- "Muenchen" did not match **München**
- "Łódź", "Straße", "Øster" stayed accent-sensitive

### After
One shared `normalizeForSearch`, applied to the keyword and every searched field (name, location, TPO, country):
- **Hyphen → space** (+ collapse repeats): "Plant for Ghana" ↔ "Plant-for-Ghana"
- **Diacritic strip** (NFD + combining-mark removal): "Yucatan" ↔ "Yucatán"
- **Stroke/ligature fold** for letters NFD leaves behind: ø ł đ ð ß æ œ þ ı (Łódź → lodz, Straße → strasse)
- **German digraphs**: each field is matched against both its accent-stripped and its ae/oe/ue-expanded form, so "München" / "Munchen" / "Muenchen" all cross-match
- **Separator-only keyword** (e.g. `-`, `---`) returns no results, since it normalizes to an empty string that would otherwise match every project

## Language coverage (verified)

| Language | Status |
|----------|--------|
| Spanish | ✅ ñ, áíóú, ü |
| French | ✅ é-è-ê, ç, ï, œ, æ |
| Czech | ✅ č ř š ž ě ů ý |
| German | ✅ accents **and** ae/oe/ue digraphs |

The German digraph path is purely additive (an extra match path), so it cannot remove matches for the other languages.

## Notes for review
- Logic verified with executable checks against representative names in all four languages (all pass, no false positives).
- No local tsc/lint was available in my environment — please rely on CI for type/lint.
- Post-approval changes: separator-only keyword guard, plus a small internal refactor (fold regex derived from the fold map). No change to the normalization approach that was reviewed.

## Also in this PR
- Updates `validate-branch-name` in `package.json` to allow the `fix/` branch prefix (and its error message), so this branch passes the husky naming hook.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,8 +28,8 @@
     "find-typos": "cspell --words-only --unique . | sort --ignore-case >> project-words.txt"
   },
   "validate-branch-name": {
-    "pattern": "^(main|develop){1}$|^(feature|hotfix)/[a-z0-9-_]+$",
-    "errorMsg": "Invalid branch name. \n 1.Branch names can contain lowercase characters, numbers, hyphen and underscore. \n 2.Except for 'main' and 'develop', branch names must begin with 'feature/' or 'hotfix/' "
+    "pattern": "^(main|develop){1}$|^(feature|hotfix|fix)/[a-z0-9-_]+$",
+    "errorMsg": "Invalid branch name. \n 1.Branch names can contain lowercase characters, numbers, hyphen and underscore. \n 2.Except for 'main' and 'develop', branch names must begin with 'feature/', 'fix', or 'hotfix/' "
   },
   "engines": {
     "node": "22.x",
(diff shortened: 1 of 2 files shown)


mariahosfeld (commented):

mohitb35 (approved):
LGTM

PR #2984, opened 2026-07-22
Feature/reusable icon button

sunilsabatp:
## Summary

Introduces a reusable, accessible `IconButton` primitive and migrates existing icon-only controls to use it. This resolves follow-up **FU-001** from the accessibility remediation work (**A11Y-002: icon-only buttons and links without accessible names**).

The component encapsulates the A11Y-002 pattern so it is consistently applied across the codebase:

- `label` is required and applied as `aria-label` (omitting it fails type checking).
- The icon (`children`) is automatically hidden from assistive technologies.
- `type="button"` is applied by default and can be overridden when needed.
- All native `<button>` / `<a>` props are forwarded.
- `className` is merged with a minimal base reset, keeping layout and styling ownership at the call site.
- Uses an `elementType` discriminated union (`button` by default, or `link`), mirroring `WebappButton`.
- Internal links are localized, while external URLs render as plain `<a>` elements.

### New `IconButton` Component

- Added `IconButton` (`src/features/common/IconButton/`) with:
  - Button and link variants
  - Minimal shared styling
- Keeps layout concerns with the consuming component to avoid visual regressions.

### Storybook

- Added Storybook stories covering:
  - Button
  - Button with click handler
  - Internal link
  - External link
- Uses Storybook action spies (`fn()`) for click handling.
- Made structural props (`elementType`, `href`, `target`) read-only in Controls to prevent invalid story states.

### isExternalURL util

`isExternalUrl` (shared util) is scheme-based and deterministic — it doesn't read `window`, so external/internal link classification is identical during SSR and on the client. This removes a potential hydration mismatch on external links (e.g. the `Footer` logos) and also fixes a pre-existing SSR hook-count mismatch in `WebappButton`, which now shares the util.

### Call Site Migrations

Migrated **22 icon-only controls** to use `IconButton`, removing duplicated accessibility markup and icon wrappers (**net -218 / +156 lines**).

#### Updated Locations

- ErrorPopup
- CookiePolicy
- SignInButton (mobile)
- CarouselSlider
- RedeemCode
  - Enter
  - Successfully
  - Failed
- DirectGift
- ProjectSearchAndFilter
  - Search
  - Filter
- ActiveSearchField
- ProjectSnippet ImageSection
- ImageSlider
- ImageSliderModal
- Account modals
  - Cancel
  - Edit
  - Pause
  - Reactivate
- DonationInfoPopover
- DonorAddressList
- TargetsModal
- ShareModal (5 social buttons)
- Footer logo links

### Accessibility Cleanup

- Account modals now use native button behavior and no longer require:
  - `role`
  - `tabIndex`
  - `onKeyPress`

### Documentation

Updated `docs/accessibility/FOLLOW_UPS.md`:

- Marked **FU-001** as implemented.
- Documented the intentionally excluded cases:
  - `UserProfileButton` (image-primary, not icon-only)
  - Footer social links (already satisfy A11Y-002 via inline `svg[aria-hidden]`)

## Testing

- No visual or behavioral changes are intended; all migrations are like-for-like replacements.
- Type checks pass.
- ESLint passes.
- Storybook verification:
  - All four stories render correctly.
  - Icons render at the expected size.
  - Click interactions are logged in the Actions panel.


src/features/common/IconButton/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+const isExternalUrl = (url: string): boolean =>
+  /^[a-z][a-z\d+.-]*:/i.test(url) || url.startsWith('//');
+
+/**
+ * Accessible icon-only button or link.
+ * Requires an accessible label, hides the icon from screen readers,
+ * and forwards all native button/link props.
+ */
+const IconButton = forwardRef<
```
rupamkairi: **Speculative Generality:** `forwardRef<HTMLButtonElement | HTMLAnchorElement, ...>` is added here, but none of the 22 migrated call sites pass a `ref`, and the sibling `WebappButton` (same discriminated-union shape) doesn't forward one either. Suggest dropping it until a real caller needs it — easy to re-add later, and it currently just adds a generic-heavy signature with no consumer.

```suggestion
const IconButton = (props: IconButtonProps): ReactElement => {
```

(and drop the closing `)` / generic params accordingly, plus the two `ref={ref as Ref<...>}` forwards on lines 92/100/120)

src/features/common/IconButton/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+      'aria-label' | 'children' | 'className'
+    > {
+  elementType: 'link';
+  href: string;
+}
+
+export type IconButtonProps = IconButtonAsButtonProps | IconButtonAsLinkProps;
+
+const isExternalUrl = (url: string): boolean =>
+  /^[a-z][a-z\d+.-]*:/i.test(url) || url.startsWith('//');
```
rupamkairi: **Duplicated Code:** This regex-based check duplicates `isExternalURL` in `WebappButton/index.tsx:45-54`, which instead does `new URL()` + `host` comparison against `window.location.host`. Same concept, two different algorithms — they'd give different answers for an absolute same-origin URL. Worth extracting one shared `isExternalUrl` helper (e.g. `src/utils/url.ts`) both components import, so the external/internal decision can't silently drift between them.

```suggestion
import { isExternalUrl } from '../../../utils/url';
```
(remove the local `isExternalUrl` definition and update `WebappButton/index.tsx` to import + call the same shared helper instead of its own `isExternalURL`)

rupamkairi (approved):
2 Low Impact Review Points flagged by Claude. 
I am approving it, As those are not functional Issues.

PR #2986, opened 2026-07-22
fix(a11y): make signup address autocomplete keyboard operable (A11Y-006)

sunilsabatp:
## Summary

Addresses **A11Y-006: Address suggestion listbox is not keyboard operable** by replacing the custom address suggestion list in `SignupAddressField` with MUI's `Autocomplete` component.

The previous implementation relied on a bespoke `role="listbox"` and mouse-only interactions, preventing keyboard users from navigating or selecting address suggestions and providing limited screen reader support.

## Changes

* Replaced the custom address suggestion list with MUI `Autocomplete`.
* Added built-in keyboard support for navigating and selecting suggestions.
* Improved accessibility semantics using the combobox pattern provided by MUI.
* Added proper active option management and screen reader announcements via MUI's accessibility features.
* Removed the unused `addressSuggestions` translation key.
* Added a follow-up item (**FU-002**) to track consolidation of duplicated address autocomplete logic.

## Accessibility Impact

This change addresses:

* **WCAG 2.1.1 – Keyboard (Level A)**
* **WCAG 4.1.2 – Name, Role, Value (Level A)**

Users can now:

* Navigate suggestions using Arrow Up/Down.
* Select a suggestion using Enter.
* Dismiss suggestions using Escape.
* Receive proper active-option announcements from assistive technologies.

### Keyboard

* Type into the address field and verify suggestions appear.
* Navigate suggestions using Arrow Up/Down.
* Select a suggestion using Enter.
* Close the suggestions list using Escape.

### Screen Reader

* Verify the field is announced as a combobox.
* Verify the active suggestion is announced while navigating the list.

### Regression

* Verify address lookup and selection continue to work as expected.
* Verify manual address entry remains supported.


src/features/user/CompleteSignup/components/SignupAddressField.tsx
```diff
@@ -109,46 +110,45 @@ const SignupAddressField = ({
...
-              setAddressSuggestions([]);
-              onBlur();
+            onChange={(_, newValue) => {
+              const selected =
+                typeof newValue === 'string' ? newValue : newValue?.text ?? '';
+              onChange(selected);
+              if (selected) {
+                handleAddressSelection(selected);
+              }
             }}
```
rupamkairi, line 141: Behavior change beyond A11Y-006 — free-text Enter can overwrite city/zipCode with geocoder guesses

In the old code, `handleAddressSelection` fired only when a user clicked a real suggestion (`onMouseDown`). Now `onChange` fires for `freeSolo` string commits too, so a user who types an arbitrary address and presses Enter sends that string to `getAddressDetailsFromText`, which may re-populate `city/zipCode` with guessed values. The finding only asked that keyboard selection of suggestions work — restrict the detail lookup to actual suggestion objects:

```suggestion
            onChange={(_, newValue) => {
              const selected =
                typeof newValue === 'string' ? newValue : newValue?.text ?? '';
              onChange(selected);
              // Only fetch address details when a real suggestion object is
              // chosen, not when free text is committed via Enter.
              if (newValue && typeof newValue !== 'string') {
                handleAddressSelection(newValue.text);
              }
            }}
```

rupamkairi (approved):
Left a Review. Please Take a  Look before adding in. 
If needs please test the case too.

PR #2989, opened 2026-07-23
feat(payments): Payment History redesign (shadcn Payments hub)

sagararyal:
## What

Redesigns donor Payment History as a shadcn/ui **Payments hub** at `/profile/payments`, replacing the legacy MUI history and recurrency pages. This is the first app surface built MUI-free. It targets **`feature/shadcn-base`**, the branch that adds the shadcn/ui + Tailwind v3 foundation (preflight off, alongside MUI) that this consumes.

Uses the clean `GET /app/payments`, `/app/payments/{guid}`, `/app/subscriptions`, and `/app/profile/paymentMethods` endpoints.

## Highlights

- **Transactions**: chip filters (All / Completed / Pending / Failed / PlanetCash), desktop master-detail plus mobile Sheet, infinite scroll, shareable `?txn=` and `?filter=` deep-links.
- **Detail**: amount, friendly payment method, per-project breakdown, fundraiser link, dedication, certificate.
- **Recurring**: shadcn record cards plus Edit / Pause / Cancel / Reactivate modals (was MUI).
- **Payment methods**: list and remove (card / SEPA).
- **PlanetCash** transactions redirect into the hub, pre-filtered.
- Legacy `/profile/history` and `/profile/recurrency` removed; sidebar simplified to **Payments** and **Donation receipts** (plus **Payouts** for ROs).
- Dates localized by route locale; new `Payments` i18n namespace across 7 locales.

## Known gaps (backend-dependent; the FE handles these defensively)

- [ ] **Pending/failed dates** not all visible. The `created` fallback is on the `/app/payments` list (treecounter-platform #534) but not yet on the serving env, and pending donations still have no date.
- [ ] **Recurring fundraiser name** not shown when a subscription funds a fundraiser: `/app/subscriptions` exposes no fundraiser field.
- [ ] **Composite-donation certificates** not returned by `/app/payments/{guid}`.
- [ ] **Payment-method default** not indicated: `isDefault` is never set by the API (the FE already renders a Default badge when it is true), and there is no set-default endpoint.

## Deferred (out of scope for this PR)

- [ ] Payouts (RO) view: needs a tpo test account to build/verify.
- [ ] Add payment method: needs Stripe Elements (webapp has no Stripe integration).
- [ ] Native-speaker translation review (current strings are machine-translated).

## Testing

Verified live against staging (logged-in donor): transactions list, filters, and detail; recurring modals; payment-methods list and remove; deep-links; the PlanetCash redirect. `lint:errors` is green and the production build compiles (the SSG step needs Redis creds, which CI provides).


sagararyal:
## Review triage

Triaged all CodeRabbit and CodeFactor findings against the current code and applied the 9 that hold in `54ac32aba`. Two differ from the original suggestion:

- **next-intl labels**: confirmed real. next-intl v4 returns the key path (e.g. `Me.complete`) instead of throwing on a missing message, so the `try/catch` fallback was dead and unmapped status/method codes rendered raw keys. Used literal-key maps rather than `t.has()`, because `t.has(dynamicKey)` also trips the typed-messages `TS2345`. This additionally cleared 3 pre-existing dynamic-key type errors.
- **Pagination**: the `requestIdRef` data guard already covers stale responses, so no wrong next-link is appended. The real defect was a stuck `isLoadingMore` flag that blocked all further pagination after a mid-flight filter change. Fixed by resetting it when the query changes.

### Skipped, with reason

1. **`IssueCodesForm` redirect `?txn=${res.uid}`**: not applied. `res.uid` is the donation's human reference (e.g. `001565746`), while `?txn=` is fetched via `GET /app/payments/{guid}` using the `don_…` guid (row clicks pass `item.guid`). Forwarding the reference would resolve to a 404 and surface an error toast right after a successful issuance. The old page's `?ref=` was a client-side highlight param, not a fetch key, so it does not carry over. Kept the bare `/profile/payments` redirect.
2. **CodeFactor "Complex Method" (`PaymentDetailContent`, `RecurringRecord`)**: left as-is. Both are already factored (Section/DetailRow helpers, literal-key label switches). A further split would relocate the branching rather than reduce it, and would thread translators and props through new sub-components, adding regression surface under `ignoreBuildErrors: true`. Advisory and non-blocking.

Everything else from the review is applied. tsc is down to 120 (from 134 over the branch) and `lint:errors` is clean.

PR #2992, opened 2026-07-24
fix(a11y): use native table for donation receipt donations list

sunilsabatp:
## Summary

Addresses **A11Y-018: Custom ARIA table has header outside the table role** by replacing the role-based donations table with a native HTML table.

The previous implementation rendered the header row and total outside the element with `role="table"`, preventing screen readers from correctly associating column headers with data cells. This change adopts semantic table markup and gives the table an accessible name.

## Changes

- Replaced the role-based table structure with a native `<table>`.
- Added semantic table elements:
  - `<thead>`
  - `<tbody>`
  - `<tfoot>`
  - `<th scope="col">`
  - `<td>`
- Added an `aria-label` to the table so screen readers announce it.
- Converted the `DonationsTable` SCSS from flexbox to a table layout, using `border-collapse: separate` with `border-spacing: 0` so the table's horizontal padding still applies (with `collapse`, table padding is ignored, which left the columns flush and misaligned with the donor details section).
- Adjusted section padding so the donations table and donor details align consistently.
- Narrowed the total row type guard to ensure `amount` and `currency` are non-null, resolving a pre-existing TypeScript error.

## Accessibility

This change resolves **A11Y-018** by ensuring:

- Column headers are correctly associated with data cells.
- The table has an accessible name.
- Screen readers can navigate the table using native table semantics.
- The implementation complies with **WCAG 1.3.1 – Info and Relationships (Level A)**.

## Testing

- Verified the donations table renders correctly with the updated semantic markup
- Verified column alignment matches the donor details section.
- Verified the table is announced with its name and column headers by screen readers

### DIFF
--- a/src/features/user/DonationReceipt/DonationReceipt.module.scss
+++ b/src/features/user/DonationReceipt/DonationReceipt.module.scss
@@ -62,15 +62,16 @@
 .receiptDataSection {
   width: 100%;
   border-radius: 16px;
-  padding: 20px;
+  padding: 16px;
   background-color: $dsWhite;
 }
 
 .donationsTable {
   width: 100%;
   table-layout: fixed;
-  border-collapse: collapse;
-  padding: 6px 16px 16px;
+  border-collapse: separate;
+  border-spacing: 0;
+  padding: 8px 8px 24px 8px;
 
   @include xsPhoneView {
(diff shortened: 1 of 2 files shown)


mohitb35 (approved):
LGTM

PR #2993, opened 2026-07-24
fix(a11y): add autocomplete attributes to signup and profile fields

sunilsabatp:
## Summary

Addresses **A11Y-016: Missing `autocomplete` on identity and address fields** by adding appropriate `autoComplete` attributes to profile and signup form inputs.

Previously, several name, email, and address fields did not expose their input purpose to browsers, preventing users from benefiting from autofill and increasing the amount of manual typing required. This change adds the appropriate autocomplete tokens while preserving the existing UI and behavior.

## Changes

- Added `autoComplete` attributes to identity fields:
  - `given-name`
  - `family-name`
  - `email`
- Added `autoComplete` attributes to address fields:
  - `street-address`
  - `address-level2`
  - `postal-code`
- Updated the following components:
  - `EditProfileForm`
  - `FullNameInput`
  - `CompleteSignup`
  - `SignupAddressField`

## Accessibility

This change resolves **A11Y-016** by enabling browser autofill and exposing the purpose of form fields to assistive technologies.

**WCAG**
- **1.3.5 – Identify Input Purpose (Level AA)**

## Testing

- Verified the appropriate `autoComplete` attributes are applied to all affected inputs.
- Confirmed browser autofill recognizes supported fields.
- Verified there are no visual or functional regressions.

PR #2996, opened 2026-07-24
Release: accessibility fixes

mohitb35:
Includes:

- #2995
- #2994 
- #2993 
- #2992
- #2987
- #2986
- #2985
- #2984
- #2998 
- #2999


sunilsabatp (approved):

mariahosfeld (changes requested):
When using the keyboard to navigate, we need to have a highlight-effect to indicate to the user where they are in the selection.
Example: Time Travel - only after pressing enter, I see which year I had selected, while using tab I have no indication where I am in the navigation.
Please check all places that are now accessible for this.

mariahosfeld (approved):

PR #2998, opened 2026-07-26
fix(a11y): expose expanded state on disclosure toggles (A11Y-021)

sunilsabatp:
## Summary

Addresses **A11Y-021: Disclosure toggles missing `aria-expanded`** by exposing the expanded/collapsed state of disclosure controls to assistive technologies.

Previously, the UI tracked the open/closed state in React, but that state was not reflected through ARIA attributes. As a result, screen reader users could activate the controls but had no indication whether the associated content was currently expanded or collapsed.

## Changes

- Added `aria-expanded` to all disclosure controls so screen readers announce their current state.
- Added `aria-controls` to associate each toggle with the content region it expands/collapses.

### AboutProject
- Added `type="button"` to prevent unintended form submission.
- Added `aria-expanded` and `aria-controls` for the collapsible description.

### MapFeatureExplorer
- Added `aria-expanded` and `aria-controls` to the desktop explore toggle.
- Threaded an optional panel ID through `MapSettings` so both desktop and mobile panels can be referenced correctly.
- Moved the `<h3>` outside the button and replaced it with a styled `<span>` to avoid placing a heading inside an interactive control while preserving the existing visual appearance.

### NavLink
- Added `aria-expanded` to both submenu toggle buttons since either control can expand/collapse the submenu.
- Added an accessible label to the arrow button (`Me.toggleSubMenu`).
- Wrapped submenu items in a container that serves as the `aria-controls` target.
- Simplified the repeated submenu condition into a single `hasSubMenu` flag.

## Why

These changes expose the current expanded/collapsed state to assistive technologies, allowing screen reader users to understand whether each disclosure widget is open or closed.

This addresses:
- **WCAG 4.1.2 – Name, Role, Value (Level A)**
- **WCAG 1.3.1 – Info and Relationships (Level A)**

## Testing

- Confirmed `aria-controls` references only mounted elements.
- Verified there are no visual or behavioural changes.

### DIFF
--- a/docs/accessibility/ACCESSIBILITY_FINDINGS.md
+++ b/docs/accessibility/ACCESSIBILITY_FINDINGS.md
@@ -992,6 +992,9 @@ useEffect(() => {
 | **Component / Feature** | AboutProject, MapFeatureExplorer, UserLayout submenu |
 | **Effort** | Small |
 | **Priority Score** | 50 / 100 |
+| **Status** | 🟡 Remediated in code (2026-07-26) — pending verification |
+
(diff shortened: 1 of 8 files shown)


mohitb35 (approved):
LGTM

PR #2837, opened 2026-01-16
Feature/migration usecontext to zustand user

sunilsabatp:
## Overview
This PR completes the migration of user state management from `React Context` to `Zustand stores`, improving performance, reducing prop drilling, and establishing a more maintainable state management architecture.

## Key Changes
1. Store Architecture
    - `AuthStore`: Manages authentication tokens, session state, and `isAuthResolved `flag
    - `UserStore`: Manages user profile, language preferences, and profile refetch triggers

 2. Authentication & Token Management
    - Migrated `token handling` from Context to Zustand (authStore)
    - Created reusable `useAuthSession hook` to decouple authentication logic
    - Simplified auth initialization through `StoreInitializer `component
    - Moved profile loading state to `isAuthResolved `in AuthStore for better separation of concern 
    
 3. User Profile Management
    - Replaced `useUserProps `Context hook with direct Zustand store selectors across the application
    - Migrated `fetchUserProfile `from Context to UserStore with improved error handling:
             - Returns fetched profile data for better composability
             - Handles impersonation-specific 403 errors locally without triggering global error handling
             - Propagates impersonation errors to calling components
             - Treats all other API errors as global profile errors

    - Renamed `refetchUserData `to `shouldRefetchUserProfile` for clarity
    
4. Language & Localization
    - Removed `setUserLanguage `action and centralized language initialization via `initializeLocale`
    - Replaced `userLang ` from UserProps with `userLanguage ` from UserStore
    
 5. Impersonation
    - Migrated `isImpersonationModeOn` from Context to Zustand store
    - Decoupled `fetchUserProfile` from auth store token for better flexibility
    - Replaced `loadUser `with store-based `fetchUserProfile`
    
 6. Developer Experience
    - Introduced barrel exports (`src/features/stores/index.ts`) for unified store imports
    - Added devtools naming with 'storeName/action_name' pattern for better debugging
    - Replaced generic parameter names with descriptive ones (e.g., `value`→ `userProfile`)
    
 7. Cleanup
    - Removed redundant `UserPropsContext `and `UserPropsProvider`
    - Simplified effect logic across components by leveraging store selectors
    - Extracted `transformProfileToForestUserInfo `helper to reduce code duplication
    
Technical Improvements :
Before (Context):

`const { user, setUser, isImpersonationModeOn } = useUserProps();`

After (Zustand):
```
const userProfile = useUserStore((state) => state.userProfile);
const setUserProfile = useUserStore((state) => state.setUserProfile);
[...]

PR #3018, opened 2026-07-30
Fix localStorage language writer

### LINKED ISSUE #3015: localStorage.language is never written, so donations and login screens lose the user's language
**Describe the bug**
Nothing writes `localStorage.language` any more, but 21 places still read it. So the app falls back to English in several places, and donation links lose the user's language.

Production. Live since 2026-01-28.

**To Reproduce**
1. Clear `localStorage.language`, or use a fresh browser profile
2. Open the site in a non-English language, for example `/de`
3. Check `localStorage.language`. It is not set, and switching language does not set it
4. Start a donation from a project. The donation URL carries `locale=null`
5. Open the login screen. It is in English, not German

**Expected behavior**
`localStorage.language` follows the current locale, and updates when the user switches language.

**What actually happens**
The key is never written. Readers get `null` and use their English fallback.

Impact, worst first:

1. **Donation language is broken.** `src/utils/getDonationUrl.ts:12` reads the key and puts it into `&locale=${language}`, so donations opened from the webapp lose the user's language
2. **Auth0 screens are always English.** Ten call sites pass `ui_locales: localStorage.getItem('language') || 'en'`, including `UserPropsContext.tsx`, `Navbar/microComponents/SignInButton.tsx` and `pages/sites/[slug]/[locale]/login.tsx`
3. **Dates render in English.** `src/utils/countryCurrency/getFormattedDate.ts` plus `SingleReview`, `TopProjectReports`, `ContributionSummary` and `RegistrationSummary`

**Additional context**

Cause: `QueryParamsProvider` carried an effect that had nothing to do with query params.

`useEffect(() => { if (localStorage.getItem('language') !== locale) localStorage.setItem('language', locale); }, [locale])`

#2820 deleted that file on 2026-01-28 and did not carry the effect over. The PR described itself as a refactor with no functional changes.

Current state, checked today:

- `grep -rn "setItem('language'" src pages` returns 0
- `grep -rn "getItem('language')" src pages` returns 21

Why this went unreported for six months: the symptom differs per user. Anyone who used the site before 2026-01-28 still has the key, frozen at whatever language they had then, so they get a wrong language that never updates. Everyone since gets the English fallback.

**Planned fix**

Add a `useInitializeLanguage` hook that restores the effect, keyed on `useLocale()`, and call it from `StoreInitializer`, which is mounted once in `_app.tsx` and already hosts seven similar init hooks. Language switching is a route change, so `useLocale()` updates and the effect refires, which is the same mechanism the original relied on.

To be picked up after #3011 is released, on its own branch off `develop`.

mohitb35:
Fixes #3015.

Restore the functionality to synchronize the application's language setting with localStorage. This change ensures that the language preference is correctly stored and updated based on the current locale. 

This makes sure that legacy readers of localStorage.language get the correct value, fixing 
- the missing locale in donation links
- Auth0 screens always appearing in English
- Dates appearing in English format

This is a minimal fix for the immediate issue.

Left for later
- migrating away from localStorage.language across the app, 
- minor issues in refreshing date format when language is changed (due to conflicts with current WIP like #2837, #2989)

- Added a custom hook to manage language synchronization with localStorage.
- Integrated the language initialization into the StoreInitializer component.

### DIFF
--- a/src/features/common/StoreInitializer/StoreInitializer.tsx
+++ b/src/features/common/StoreInitializer/StoreInitializer.tsx
@@ -7,6 +7,7 @@ import { useInitializeIntervention } from '../../../hooks/useInitializeIntervent
 import { useInitializeView } from '../../../hooks/useInitializeView';
 import { useInitializeSingleProject } from '../../../hooks/useInitializeSingleProject';
 import { useInitializeTenant } from '../../../hooks/useInitializeTenant';
+import { useInitializeLanguage } from '../../../hooks/useInitializeLanguage';
 
 interface StoreInitializerProps {
   tenantConfig?: Tenant;
@@ -31,6 +32,7 @@ export const StoreInitializer = ({
 }: StoreInitializerProps) => {
   useInitializeTenant(tenantConfig);
   useInitializeParams();
+  useInitializeLanguage(); // Sync route locale to legacy localStorage readers.
   useInitializeCurrency();
   useInitializeView(isMobile);
   useInitializeProject();
--- a/src/hooks/useInitializeLanguage.ts
+++ b/src/hooks/useInitializeLanguage.ts
@@ -1,0 +1,12 @@
+import { useEffect } from 'react';
+import { useLocale } from 'next-intl';
+
+export const useInitializeLanguage = () => {
+  const locale = useLocale();
+
+  useEffect(() => {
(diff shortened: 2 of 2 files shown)


mariahosfeld (approved):
LGTM!

PR #3023, opened 2026-08-04
chore: add shadcn/ui + Tailwind v3 foundation

sunilsabatp:
## Summary

This PR adds the initial **Tailwind CSS v3** and **shadcn/ui** foundation alongside the existing **MUI, Emotion, and SCSS** stack.

The goal is to support a gradual UI migration while keeping existing pages and styles unchanged.

## Changes

* Added Tailwind CSS v3.
* Added shadcn/ui configuration.
* Disabled Tailwind preflight to avoid applying a global CSS reset.
* Added shadcn design tokens in `src/styles/globals.css`.
* Mapped the shadcn primary color to the existing application primary color:

  * `--primary: #68B030`
  * Matches the existing `--primary-color`.
* Added the `@/*` to `src/*` path alias.
* Added `components.json` for shadcn/ui configuration.
* Added the shared `cn()` utility in `src/lib/utils.ts`.
* Added the shadcn `Button` as the first reusable UI component.
* Added the required Tailwind and PostCSS configuration.

## Motivation

The application currently uses MUI, Emotion, and SCSS.

This setup allows new components and gradually migrated pages to use Tailwind and shadcn/ui without requiring an immediate rewrite of the existing UI.

## Backward Compatibility

Tailwind preflight is disabled:

```ts
corePlugins: {
  preflight: false,
}
```

This prevents Tailwind from applying its default global reset to existing elements and reduces the risk of changes to:

* Existing MUI components
* SCSS-based pages
* Typography
* Form controls
* Layouts and spacing

## Design Tokens

The shadcn tokens in `src/styles/globals.css` mirror the application’s existing colors.

The primary token is based on the existing web application primary color:

```css
--primary: #68B030;
```

## Usage Example

```tsx
import { Button } from "@/components/ui/button"

<Button>Continue</Button>
```

## Validation

* Verified that the existing home page remains pixel-identical.
* Confirmed that existing MUI, Emotion, and SCSS styles continue to work.
* Confirmed that Tailwind utility classes are generated.
* Confirmed that the shadcn `Button` uses the application design tokens.
* Confirmed that imports using the `@/` alias resolve correctly.

## Scope

This PR only introduces the shared Tailwind and shadcn/ui foundation.

Existing pages and MUI components are not migrated as part of this PR. Additional components and page migrations will be handled gradually in separate PRs.


jmiridis:
@sunilsabatp what is required to merge this PR?

jmiridis:
Reviewed. The shape of this is right: preflight off, tokens mapped to the real webapp colors, PostCSS config replicating the Next defaults instead of clobbering them, and only one component added. That is the correct way to land Tailwind next to MUI.

I found two things that were broken, both a side effect of turning preflight off. I pushed the fixes to this branch in 27b4fe953 rather than leaving them as comments, since the second one would have blocked every component added after `Button`. Please review that commit.

## 1. Every shadcn border rendered invisible

The outline variant uses `border border-input`. Tailwind's `.border` sets `border-width` only. `border-style` normally comes from preflight, which is disabled here, so nothing declared a style:

* `<div>` based components (Card, Input, Table, Select, Dialog, all coming next) default to `border-style: none`
* `<button>` hits `button { border: 0px }` in `src/theme/global.scss:291`, and the shorthand resets style to `none` too

Verified in a real browser against the compiled stylesheet, before and after:

```
before  #outline (button)  border-style: none   -> nothing renders
before  #card    (div)     border-style: none   -> nothing renders
after   #outline (button)  border-style: solid  border-width: 1px
after   #card    (div)     border-style: solid  border-width: 1px
after   #primary (button)  border-style: none   border-width: 0px  (untouched)
```

The fix in `globals.css` is keyed on the border utility classes, not on `*`:

```css
[class~='border'],
[class*='border-'] {
  border-style: solid;
  border-color: hsl(var(--border));
}
```

An attribute selector matches class specificity, so it outranks the `button` element rule in `global.scss`. A universal selector would lose to it, which is why the usual preflight-lite snippet does not work in this codebase. Nothing outside shadcn is affected, since no existing code uses `border-*` utility classes.

## 2. The dark token block was dead code

`darkMode: 'class'` makes Tailwind look for `.dark`. This app never renders that class. `ThemeProvider` produces `theme-light` / `theme-dark` (`src/theme/themeContext.tsx`) and `Layout` puts it on the wrapper div (`src/features/common/Layout/index.tsx:24`). Changed to `darkMode: ['class', '.theme-dark']` and renamed the CSS block. `dark:` utilities now compile to `:is(.theme-dark *)`.

## 3. Smaller items in the same commit

* `globals.css` now imports **after** `global.scss` in `_app.tsx`. On equal specificity the later sheet wins, so Tailwind utilities should come last.
* Added `globals.css` to `.storybook/preview.js`, otherwise shadcn components render unstyled in Storybook.
* Dropped the `./components/**/*.{ts,tsx}` content glob. That directory does not exist at the repo root.

## Left for you

**`postcss-preset-env: ^11`** is a large jump from what Next 14 bundles internally (v7 era), and `stage: 3` in v11 covers a wider feature set than the config it is imitating. It processes the output of every existing SCSS file, so it is the one line here with app-wide blast radius. I did not touch it. Could you diff a production build against `develop` to confirm the emitted CSS is unchanged?

**Radius, a design question rather than a bug.** `--radius: 0.5rem` gives `rounded-md` buttons, while the existing MUI buttons are pills (`borderRadius: 28` in `src/theme/themeStyles.ts:111`). During the migration both shapes will sit on the same page. Worth deciding deliberately whether shadcn matches the current pill shape or whether we move to the squarer look, so the mixed state reads as intentional.

## Not blocking

Nothing imports `Button` yet, so the risk to `develop` is close to zero. The real test is #2989, which stacks 82 files on this branch, and both issues above would have shown up there first.

Note that #3006 adds a global focus-visible ring, which overlaps with shadcn's own `focus-visible:ring-2`. Whichever of the two merges second should check for a doubled ring.

sunilsabatp:
> Reviewed. The shape of this is right: preflight off, tokens mapped to the real webapp colors, PostCSS config replicating the Next defaults instead of clobbering them, and only one component added. That is the correct way to land Tailwind next to MUI.
> 
> I found two things that were broken, both a side effect of turning preflight off. I pushed the fixes to this branch in [27b4fe9](https://github.com/Plant-for-the-Planet-org/planet-webapp/commit/27b4fe953b67ac397349312d9126ffdca59d95c0) rather than leaving them as comments, since the second one would have blocked every component added after `Button`. Please review that commit.
> 
> ## 1. Every shadcn border rendered invisible
> The outline variant uses `border border-input`. Tailwind's `.border` sets `border-width` only. `border-style` normally comes from preflight, which is disabled here, so nothing declared a style:
> 
> * `<div>` based components (Card, Input, Table, Select, Dialog, all coming next) default to `border-style: none`
> * `<button>` hits `button { border: 0px }` in `src/theme/global.scss:291`, and the shorthand resets style to `none` too
> 
> Verified in a real browser against the compiled stylesheet, before and after:
> 
> ```
> before  #outline (button)  border-style: none   -> nothing renders
> before  #card    (div)     border-style: none   -> nothing renders
> after   #outline (button)  border-style: solid  border-width: 1px
> after   #card    (div)     border-style: solid  border-width: 1px
> after   #primary (button)  border-style: none   border-width: 0px  (untouched)
> ```
> 
> The fix in `globals.css` is keyed on the border utility classes, not on `*`:
> 
> ```css
> [class~='border'],
> [class*='border-'] {
>   border-style: solid;
>   border-color: hsl(var(--border));
> }
> ```
> 
> An attribute selector matches class specificity, so it outranks the `button` element rule in `global.scss`. A universal selector would lose to it, which is why the usual preflight-lite snippet does not work in this codebase. Nothing outside shadcn is affected, since no existing code uses `border-*` utility classes.
> 
> ## 2. The dark token block was dead code
> `darkMode: 'class'` makes Tailwind look for `.dark`. This app never renders that class. `ThemeProvider` produces `theme-light` / `theme-dark` (`src/theme/themeContext.tsx`) and `Layout` puts it on the wrapper div (`src/features/common/Layout/index.tsx:24`). Changed to `darkMode: ['class', '.theme-dark']` and renamed the CSS block. `dark:` utilities now compile to `:is(.theme-dark *)`.
> 
> ## 3. Smaller items in the same commit
> * `globals.css` now imports **after** `global.scss` in `_app.tsx`. On equal specificity the later sheet wins, so Tailwind utilities should come last.
> * Added `globals.css` to `.storybook/preview.js`, otherwise shadcn components render unstyled in Storybook.
> * Dropped the `./components/**/*.{ts,tsx}` content glob. That directory does not exist at the repo root.
> 
> ## Left for you
> **`postcss-preset-env: ^11`** is a large jump from what Next 14 bundles internally (v7 era), and `stage: 3` in v11 covers a wider feature set than the config it is imitating. It processes the output of every existing SCSS file, so it is the one line here with app-wide blast radius. I did not touch it. Could you diff a production build against `develop` to confirm the emitted CSS is unchanged?
> 
> **Radius, a design question rather than a bug.** `--radius: 0.5rem` gives `rounded-md` buttons, while the existing MUI buttons are pills (`borderRadius: 28` in `src/theme/themeStyles.ts:111`). During the migration both shapes will sit on the same page. Worth deciding deliberately whether shadcn matches the current pill shape or whether we move to the squarer look, so the mixed state reads as intentional.
> 
> ## Not blocking
> Nothing imports `Button` yet, so the risk to `develop` is close to zero. The real test is #2989, which stacks 82 files on this branch, and both issues above would have shown up there first.
> 
> Note that #3006 adds a global focus-visible ring, which overlaps with shadcn's own `focus-visible:ring-2`. Whichever of the two merges second should check for a doubled ring.

I diffed the production CSS against the pre-Tailwind baseline.

Initially, there were some changes in the existing SCSS output. The cause was that our custom `postcss.config.js` was not using the same browser targets that Next.js 14.2.35 uses internally.

I added the same browser targets:

```js
browsers: ['chrome 64', 'edge 79', 'firefox 67', 'opera 51', 'safari 12'],
```

After rebuilding and diffing again, all of the unexpected prefix differences disappeared, including `backdrop-filter`, text-decoration prefixes, Swiper styles, `object-fit`, `column-gap`, and `min/max-content`.

The only remaining v11-specific difference is an additional `-moz-available` fallback for an existing `height: stretch` rule. It is additive and does not change the existing behavior.

So the issue was the missing browser-target configuration, not `postcss-preset-env@11` itself. With the targets pinned, the existing emitted CSS is effectively unchanged apart from the expected new Tailwind CSS and that harmless Firefox fallback. So we can go ahead with `postcss-preset-env@11`.

Regarding the radius, good point. The current `--radius: 0.5rem` value was originally added by Sagar.

@Sagar, could you confirm whether the intention is to keep the current `0.5rem` shadcn radius and move toward the squarer style, or whether we should match the existing MUI pill-style buttons during the migration ?

sunilsabatp:
> @sunilsabatp what is required to merge this PR

 @jmiridis  From my side, the PostCSS issue is resolved after adding Next.js 14’s browser targets and verifying the production CSS diff. The remaining item is the radius/design decision, which needs confirmation from Sagar. Once that is decided this should be ready to merge.

PR #3050, opened 2026-08-10
fix(api): guard against malformed impersonationData in localStorage

### LINKED ISSUE #3041: P1 — Guard the impersonation-data parse so a bad value cannot break all API calls
Part of #3030 (follow-up epic) · tracked as **P1**.

**Review ref:** #9 · **Size:** small · **Risk:** low, but touches every request · **Links to:** A6 (the urgent part of that cleanup)

**Scope**
`setHeader.ts` runs `JSON.parse(localStorage.getItem('impersonationData'))` without a
try/catch, and in `useApi` this runs **before** the request's try block. A corrupt
`impersonationData` value throws and escapes, which makes **every** API request in the app fail
until the bad key is cleared.

**Change**
Wrap the parse in a try/catch and treat a bad value as "no impersonation".

**Where**
`src/utils/apiRequests/setHeader.ts:9-11`, called from `src/hooks/useApi.ts:143`.

**Why its own PR**
It is small but affects every request, so it deserves a focused review rather than being buried.

**For the tracker** · Area: api · Labels: `bug`, `security`, `priority: medium` · Depends on: none
**Done when:** a corrupt `impersonationData` value is caught and treated as "no impersonation"; API requests still succeed when a bad key is present.

### LINKED ISSUE #3056: setHeaderForImpersonation can send an incomplete or mismatched impersonation credential pair
`setHeaderForImpersonation` builds the two impersonation headers field by field, so it can send an incomplete credential pair, and it can take the email from one source and the pin from another.

Found while writing tests for the function. The tests for the three cases below are in the same PR and currently fail on purpose.

## The code

[src/utils/apiRequests/setHeader.ts](../../src/utils/apiRequests/setHeader.ts) resolves each header on its own:

```ts
if (impersonationData?.targetEmail || impersonationDataFromLocal?.targetEmail) {
  header['X-SWITCH-USER'] = impersonationData?.targetEmail || impersonationDataFromLocal?.targetEmail;
}
if (impersonationData?.supportPin || impersonationDataFromLocal?.supportPin) {
  header['X-USER-SUPPORT-PIN'] = impersonationData?.supportPin || impersonationDataFromLocal?.supportPin;
}
```

There is no check that both fields are present, and no check that they came from the same place.

## How the API responds

The API only impersonates when both headers are sent and the pin matches the target named in `X-SWITCH-USER`.

A pin sent without an email is a silent no-op in any environment: no impersonation is attempted and the response is the caller's own profile. The UI disagrees, because [src/hooks/useInitializeUser.ts](../../src/hooks/useInitializeUser.ts) sets the impersonation flag from the presence of the localStorage key alone.

An email without a pin, or a mismatched pair, is rejected with a 403 in production. On non-production, where validation is deliberately looser, it can be granted instead — including with a pin that was never supplied for that target.

## Can it happen?

Not through the form. [ImpersonateUserForm.tsx](../../src/features/user/Settings/ImpersonateUser/ImpersonateUserForm.tsx) marks both fields required and guards with `if (data.targetEmail && data.supportPin)`.

But `enterImpersonation` stores what the API returned, not what the form validated:

```ts
enterImpersonation({
  targetEmail: res.email,
  supportPin: res.supportPin,
});
```

If a profile response omits `supportPin`, `JSON.stringify` drops the key and localStorage ends up holding email-only data. That is case 1 above.

Case 3 needs a pin to survive in localStorage while a new email arrives by argument.

Hand editing localStorage reaches all three. Impersonation is a support-staff feature, so the people who can do that are the people who use it.

## Suggested fix

Treat a source as a unit, and send nothing unless the pair is complete:

```ts
const data = impersonationData ?? impersonationDataFromLocal;
if (data?.targetEmail && data?.supportPin) {
  header['X-SWITCH-USER'] = data.targetEmail;
  header['X-USER-SUPPORT-PIN'] = data.supportPin;
}
```

One source wins whole. No partial sends, and no cross-source mixing.

## Relationship to PR #3050

PR #3050 rewrites this function to wrap the `JSON.parse` in a `try/catch`. It keeps the per-field `||` logic, so it does not change any of the three behaviours here.

Whether this is folded into #3050 or done as a follow-up after it merges is still open.

## Tests

Five tests in `src/utils/apiRequests/setHeader.test.ts` assert the corrected behaviour and fail until this is fixed.

Incomplete pair from localStorage:

- `sets no headers when the stored data has no pin`
- `sets no headers when the stored data has no email`

Incomplete pair from the argument:

- `sets no headers when the argument has no pin`
- `sets no headers when the argument has no email`

Fields taken from two different sources:

- `does not combine fields from the argument and localStorage`

They are plain failing tests rather than skipped ones, so the gap stays visible.

### The argument path is guarded today

`src/stores/userStore.ts` is the only caller that passes the second argument, and only `ImpersonateUserForm` reaches it, after checking both fields. So incomplete argument data cannot arrive through the app right now.

The two argument-only tests still earn their place. The comment above that call site already anticipates other callers:

> If another caller passes `impersonationData`, its 403 will also be handled locally.

A future caller without the form's check would hit exactly this, and the contract should hold for whoever calls it rather than depending on one form validating first.

### Not owned by this issue

A sixth test in the same file, `sets no headers when the stored data is not valid JSON`, belongs to PR #3050. I ran this suite against that branch: it turns that one green and leaves these five red, so the two split cleanly.

### Effect on CI

The suite is red until this is fixed. `.github/workflows/test.yml` runs the tests on every pull request and on pushes to `develop`, but is deliberately not a required check for this reason. Make it required once this and #3050 close.

sunilsabatp:
## Summary

Improves impersonation handling so malformed, incomplete, or stale impersonation data cannot break API requests or cause the UI and request headers to disagree.

This PR now centralizes impersonation storage and validation, safely handles corrupted `localStorage` data, ensures impersonation headers are only sent as a valid email/PIN pair, and prevents an invalid explicitly supplied argument from falling back to stale stored data.

## Changes

* Added a shared `src/utils/impersonation.ts` module that owns:

  * `ImpersonationData`
  * `IMPERSONATION_STORAGE_KEY`
  * `isValidImpersonationData`
  * `readStoredImpersonationData`

* Removed the old `src/utils/apiRequests/impersonation.tsx` type-only module.

* Updated `setHeaderForImpersonation` to:

  * safely handle malformed JSON in `localStorage` instead of throwing
  * validate impersonation data before using it
  * only send impersonation headers when both `targetEmail` and `supportPin` are valid
  * treat email and PIN as a single pair rather than resolving them independently
  * use stored impersonation data only when no explicit argument is provided
  * send no impersonation headers when an explicit argument is provided but is invalid, rather than falling back to stale stored data

* Updated `useInitializeUser` to derive impersonation mode from validated stored data instead of only checking whether the `impersonationData` key exists.

* Updated `userStore.ts`, `ImpersonateUserForm.tsx`, `setHeader.ts`, and `useInitializeUser.ts` to use the shared impersonation utilities.

## Why

Previously, malformed `impersonationData` could throw during `JSON.parse` and break unrelated API requests going through the shared request path.

Different parts of the application also had different definitions of whether impersonation was active. For example, invalid stored data could result in no impersonation headers being sent while the UI still showed impersonation mode as enabled.

There was also a case where an explicitly supplied but incomplete impersonation argument could fall back to previously stored credentials. This could cause stale impersonation data to be used for a request that had explicitly supplied different data.

Centralizing the validation and storage logic gives the application one consistent definition of valid impersonation data.

## Behaviour changes

* Malformed impersonation JSON in `localStorage` no longer throws or blocks API requests.
* Missing or empty `targetEmail` / `supportPin` values are treated as invalid impersonation data.
* Impersonation headers are only sent as a complete pair:

  * `X-SWITCH-USER`
  * `X-USER-SUPPORT-PIN`
* Email and PIN are no longer combined from different sources.
* A valid explicitly supplied `impersonationData` argument takes precedence over stored data.
* If no argument is supplied, valid stored impersonation data may still be used.
* If an argument is explicitly supplied but is invalid, the function sends no impersonation headers instead of falling back to `localStorage`.
* The impersonation UI state now uses the same validation rules as request-header generation.

Valid impersonation flows and unrelated request headers remain unchanged.

## Testing

Ran the Vitest suite:

```text
Test Files  2 passed (2)
     Tests  19 passed (19)
```

This includes coverage for:

* malformed stored JSON
* missing or empty email/PIN values
* valid stored impersonation data
* valid explicitly supplied impersonation data
* preserving unrelated headers
* preventing fields from being combined across the argument and `localStorage`
* preventing invalid explicit arguments from falling back to stored impersonation data

## Resolves

* Resolves #3041
* Resolves #3056


src/utils/apiRequests/setHeader.ts
```diff
@@ -6,29 +6,25 @@ export const setHeaderForImpersonation = (
...
-        impersonationDataFromLocal?.targetEmail;
-    }
+  let impersonationDataFromLocal: ImpersonationData | undefined;
+  try {
+    impersonationDataFromLocal = JSON.parse(
+      `${localStorage.getItem('impersonationData')}`
+    );
+  } catch {
+    impersonationDataFromLocal = undefined;
+  }
```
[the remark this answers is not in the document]
sunilsabatp: Addressed: https://github.com/Plant-for-the-Planet-org/planet-webapp/pull/3050/changes/56d7dd46125265e32b30768f85ac631635a28fc8

src/utils/apiRequests/setHeader.ts
```diff
@@ -1,34 +1,41 @@
...
+    impersonationData?.targetEmail || impersonationDataFromLocal?.targetEmail;
+  const supportPin =
+    impersonationData?.supportPin || impersonationDataFromLocal?.supportPin;
+
+  if (targetEmail) {
+    header['X-SWITCH-USER'] = targetEmail;
+  }
+  if (supportPin) {
+    header['X-USER-SUPPORT-PIN'] = supportPin;
   }
```
mohitb35: The shape guard is a good idea, and fixes the 2 localStorage cases from #3056.

Two more cases stay open, because each header is still resolved on its own.

The guard runs on the stored value but not on `impersonationData`. So the argument can still theoretically send just one value, and it can pair its email with a pin from localStorage:

```
stored = { targetEmail: 'stored@x.org', supportPin: '1111' }
arg    = { targetEmail: 'new@x.org',    supportPin: '' }

→ X-SWITCH-USER:      new@x.org
  X-USER-SUPPORT-PIN: 1111   ← stale pin, different user
```

The API needs both headers, and the pin must match the target in `X-SWITCH-USER`. Half a pair is a silent no-op or a 403. A mismatched pair is a 403 in production, but non-production validation is looser and can grant it.

The argument is the untrusted side in practice. `ImpersonateUserForm` checks both fields, but `enterImpersonation` stores `res.email` and `res.supportPin` from the API response. A response without `supportPin` writes email-only data. TypeScript does not catch it, because the parameter is typed `ImpersonationData`.

Suggest treating a source as a combined unit, so one source wins whole:

```suggestion
  const source = isValidImpersonationData(impersonationData)
    ? impersonationData
    : impersonationDataFromLocal;

  if (source) {
    header['X-SWITCH-USER'] = source.targetEmail;
    header['X-USER-SUPPORT-PIN'] = source.supportPin;
  }
```

This reuses the guard you already wrote, and applies it to both sources. No partial sends, and no cross-source mixing. It closes #3056.

src/utils/apiRequests/setHeader.ts
```diff
@@ -1,34 +1,41 @@
 import type { ImpersonationData } from './impersonation';
+
+const isValidImpersonationData = (data: unknown): data is ImpersonationData => {
+  return (
+    typeof data === 'object' &&
+    data !== null &&
+    typeof (data as ImpersonationData).targetEmail === 'string' &&
+    typeof (data as ImpersonationData).supportPin === 'string'
+  );
+};
```
mohitb35: Small gap: `typeof '' === 'string'`, so the guard accepts `{ targetEmail: '', supportPin: '' }`. An empty pin is never valid, and the #3056 tests use `''` as their "missing" value, so the fix in my other comment needs this tightened to pass.

Suggest checking for a non-empty string:

```suggestion
    typeof (data as ImpersonationData).targetEmail === 'string' &&
    (data as ImpersonationData).targetEmail !== '' &&
    typeof (data as ImpersonationData).supportPin === 'string' &&
    (data as ImpersonationData).supportPin !== ''
```

src/utils/apiRequests/setHeader.ts
```diff
@@ -1,34 +1,41 @@
 import type { ImpersonationData } from './impersonation';
+
+const isValidImpersonationData = (data: unknown): data is ImpersonationData => {
```
mohitb35: Non-blocking, but would be better to do in this small PR as a related change. 

The `'impersonationData'` key is now read or written in three layers, and none of them owns it:

- `src/stores/userStore.ts` lines 73 and 88 write and remove it
- `src/utils/apiRequests/setHeader.ts` line 21 reads and parses it
- `src/hooks/useInitializeUser.ts` line 59 checks it exists


Worth moving the guard (`isValidImpersonationData`) and the parse from `localStorage` into the module that already holds the type, `src/utils/apiRequests/impersonation.tsx`, so all three go through one path. 

- That file also has no JSX in it, so it could be a `.ts`. 
- Since it is no longer consumed only for apis, locating it at `src/utils/impersonation.ts` would fit better, matching how `src/utils/storeConfig.ts` owns its keys.

mohitb35 (changes requested):
The `try/catch` looks right, and it fixes #3041. Two notes before this merges.

**The PR body is out of date.** Commit `56d7dd461` adds the shape guard, which the description does not mention. It also changes behaviour beyond #3041: stored data that used to produce one header now produces none. Please add that, so a reviewer knows to test it. Add `Resolves #3056` too if you take the inline suggestion, since the guard already fixes two of the five cases in that issue.

**Please also fix `src/hooks/useInitializeUser.ts` line 59:**

```ts
setIsImpersonationModeOn(localStorage.getItem('impersonationData') !== null);
```

It checks that the key exists. It never checks that the value is usable.

After this PR, a stored value that fails the `isValidImpersonationData` guard sends no headers, so the API returns the real user's own profile. The key is still there, so the flag turns `true` anyway. The banner in `ImpersonationActivated.tsx` then reads "you are impersonating <your own address>", and `UserLayout.tsx` line 166 hides the menu item for starting a real impersonation.

This bug already exists on `develop` for a pin-only value. This PR just makes more inputs reach it.

Once `src/utils/impersonation.ts` exports the reader, the fix is one line:

```ts
setIsImpersonationModeOn(readStoredImpersonationData() !== undefined);
```

Then the flag and the headers cannot disagree.

mohitb35 (approved):
LGTM

Issue #3030, opened 2026-08-10
Tracking: user & auth Context-to-Zustand migration follow-ups


mohitb35:
Umbrella tracking issue for the follow-up work from the review of the **user & auth Context-to-Zustand migration** (PR #2837, which introduces `userStore` and `authStore`).

Full write-up with per-item detail: `docs/reviews/zustand-user-migration-followups.md`.

_`area` is written in each sub-issue's "For the tracker" footer as plain text, not as a label._

## Suggested order of work

**Phase 0 — small and independent (start here):**
- [x] #3041 P1 — guard the impersonation-data parse (do first; protects every request)
- [x] #3035 I5 — early return before hooks
- [x] #3032 I2 — RedeemPopup combined-boolean dependency
- [x] #3036 I6 — setHeader.ts cleanups
- [x] #3040 I10 — getDonationUrl unescaped values
- [x] #3034 I4 — clear token on logout
- [ ] #3039 I9 — impersonation auth-loss (investigate; feeds A5)
- [ ] #3037 I7 — profile error logs (blocked; park until A2)

**Phase 1 — the readiness reshape (core):**
- [x] #3044 A1 — one owner for the "auth is ready" flag (after 2878 merges)
- [ ] #3043 P3 — shared readiness selectors (after A1 and the four module PRs)

**Phase 2 — fetch and store cleanup:**
- [ ] #3042 P2 — route profile fetch through apiClient
- [ ] #3046 A3 — request sequencing
- [ ] #3045 A2 — return errors to the caller (unblocks I7)
- [ ] #3047 A4 — split fetchUserProfile + store shape (folds P2, I8)
- [ ] #3033 I3 — expired-token handling (alongside A1/A2)
- [ ] #3038 I8 — redundant GET after update (with A4)

**Phase 3 — impersonation consolidation:**
- [ ] #3048 A5 — one owner for impersonation (folds I1, I4; removes I9 symptoms)
- [ ] #3049 A6 — keep impersonation plumbing off the hot path

## Relationship to the stacked module PRs

Four module migrations sit stacked on the user migration: **#2873** (PlanetCash), **#2874** (Payout), **#2878** (Bulk-codes), **#2963** (Donation receipt). Two follow-ups wait on them:
- **P3** waits for all four (they rewrite the readiness checks in the same files P3 edits).
- **A1** waits for **#2878** (the bulk-codes flicker workaround ships there; A1 then removes it).

Everything else is independent of those four.

[...]

PR #3052, opened 2026-08-10
fix: move AddressManagement guard below hooks

### LINKED ISSUE #3035: I5 — Early `return` before hooks in AddressManagement
Part of #3030 (follow-up epic) · tracked as **I5**.

**Priority:** Low · **Review ref:** #18 · **Links to:** none (standalone)

**What is wrong**
An early `return null` sits **above** React hook calls (`useTranslations` / `useState` /
`useMemo`):
```ts
if (!userAddresses) return null;   // this is above the hooks
```
If `userAddresses` ever changes from empty to set while the component is mounted, React throws a
"rendered fewer hooks than expected" error (a Rules-of-Hooks violation).

**Where**
`src/features/user/Settings/EditProfile/AddressManagement/index.tsx:30`

**Why it matters**
Safe today because the surrounding layout guarantees the profile exists at mount, but it is
fragile.

**Suggested fix**
Move the early `return` **below** all the hook calls.

**For the tracker** · Area: user · Labels: `bug`, `refactor`, `priority: low` · Depends on: none
**Done when:** the early `return null` sits below every hook call in AddressManagement; no Rules-of-Hooks warning.

sunilsabatp:
Resolves:  #3035

## Summary

Fixes the Rules of Hooks issue in `AddressManagement` by moving the early `return null` check below all React hook calls.

Part of #3030, tracked as **I5**.

## Problem

`AddressManagement` returned early when `userAddresses` was unavailable:

```tsx
if (!userAddresses) return null;
```

This check was placed before hooks such as `useTranslations`, `useState`, and `useMemo`.

If `userAddresses` changed from unavailable to available while the component remained mounted, the component could call a different number of hooks between renders, violating React's Rules of Hooks.

## Changes

* Moved the `userAddresses` early-return guard below all React hooks.
* Ensured hooks are called unconditionally and in a consistent order.
* Preserved the existing component behavior and UI.

## Testing

* Verified that all hooks are called before the conditional return.
* Verified that the component still returns `null` when `userAddresses` is unavailable.
* Confirmed no unrelated behavior changes were introduced.

## Related

* Epic: #3030
* Follow-up item: **I5**

### DIFF
--- a/src/features/user/Settings/EditProfile/AddressManagement/index.tsx
+++ b/src/features/user/Settings/EditProfile/AddressManagement/index.tsx
@@ -25,9 +25,6 @@ import { useUserStore } from '../../../../../stores';
 const AddressManagement = () => {
   // store: state
   const userAddresses = useUserStore((state) => state.userProfile?.addresses);
-  // If addresses is null (not an empty array), it indicates a malformed API response
-  // Normal users without addresses will have an empty array, not null
-  if (!userAddresses) return null;
   const tAddressManagement = useTranslations('EditProfile.addressManagement');
   const [addressAction, setAddressAction] = useState<AddressAction | null>(
     null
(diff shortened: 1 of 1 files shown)


mohitb35 (approved):
LGTM

PR #3057, opened 2026-08-11
fix(security): percent-encode dynamic values in getDonationUrl

### LINKED ISSUE #3040: I10 — getDonationUrl builds a URL from unescaped values (hardening)
Part of #3030 (follow-up epic) · tracked as **I10**.

**Priority:** Low (not exploitable as-is) · **Review ref:** CodeQL alert #55 · **Links to:** none (standalone)

**What is wrong**
A security scanner (CodeQL, `js/xss-through-dom`) flagged `<a href={donationUrl}>`. On
inspection it is **not** exploitable as an XSS here: the URL always begins with a fixed https
base (`NEXT_PUBLIC_DONATION_URL`), so its scheme cannot be attacker-controlled. But the
underlying pattern is a real hygiene issue: `getDonationUrl` concatenates `window.location.href`,
several `localStorage` values, and the token into a URL **without escaping**, and it is used in
several places.

**Where**
`src/utils/getDonationUrl.ts` (the flagged sink is `src/features/user/BulkCodes/components/BulkCodesError.tsx:43`)

**Suggested fix**
URL-encode the user-controlled parts with `encodeURIComponent` (the callback URL, country,
language, token, gift id, slug). Its own small PR, since `getDonationUrl` is shared.

**For the tracker** · Area: api · Labels: `security`, `tech-debt`, `priority: low` · Depends on: none
**Done when:** every user-controlled part of the URL passes through `encodeURIComponent`; CodeQL alert #55 is resolved or dismissed with a written reason.

sunilsabatp:
Resolves: #3040

## Summary
- `getDonationUrl` built its query string via raw template-literal concatenation, so values derived from `window.location.href`, `localStorage`, auth tokens, gift IDs, and slugs were spliced in unescaped.
- CodeQL flagged this as `js/xss-through-dom` (alert #55) because that tainted data flows into a value used as `href`. The base URL is a fixed, app-controlled HTTPS origin, so it isn't directly exploitable as XSS, but the missing encoding was a real correctness bug: `callback_url` is frequently a full URL with its own `?a=1&b=2` query string, and injecting it raw let its `&`/`=` characters be parsed as extra top-level params of the donation URL - potentially overriding `token`, `tenant`, or `s`.
- Fix: wrap every dynamic segment in `encodeURIComponent` before it is concatenated into the URL.

## What else changed during review
- **URL composition restructured.** The nested template literal became an array of params joined with `&`, so each param is one readable line and no separator is placed by hand.
- **`JSON.parse` on `directGift` is guarded.** A malformed localStorage value used to throw out of `getDonationUrl` during render, breaking every donate and add-balance link. It now falls back to no direct gift.
- **`country` and `locale` have real defaults.** They previously sent the literal string `null` when localStorage had not been written yet, which happens on first paint and whenever the config fetch fails. They now fall back to `DE` and `en`.
- **Param types narrowed.** `embed`, `callbackUrl`, and `slug` are `string | undefined`, matching what `queryParamStore` actually holds. `useInitializeParams` already collapses repeated query params, so the `string[]` in the old signature was unreachable. `directGift` is now typed as the object it becomes rather than the `string` it starts as. All three `String()` coercions are gone as a result.
- **Tests added.** `src/utils/getDonationUrl.test.ts` covers encoding, the defaults, embed mode, the `s` param precedence, the malformed-gift case, and pins the full URL and param order as a characterization test.

## Behavior
- No change for "normal" values (ids, slugs, tenant names, tokens) - `encodeURIComponent` is a no-op on them.
- Values containing reserved URL characters are now percent-encoded rather than injected raw. Any standard query-string consumer decodes this transparently, so callers see the same effective value - and, for callback URLs with their own query string, now get the *correct* full value instead of a corrupted/truncated one.
- Two deliberate behavior changes: `country`/`locale` now send `DE`/`en` instead of `null`, and a corrupt `directGift` now yields a working link instead of an exception.
- All 4 call sites (`BulkCodesError`, `AccountDetails`, `DonateButton`, `ProjectInfoSection`) just plug the return value into an `href` and are unaffected.

## Test plan
- [x] `npm run test` passes (45 tests across 4 files).
- [x] Verify the "insufficient PlanetCash balance" add-balance link (BulkCodesError) still opens the donation flow correctly.
- [x] Verify project donate buttons still deep-link with `tenant`, `callback_url` intact when embedded.
- [x] Confirm a callback URL containing its own query string (e.g. `?ref=x&y=1`) round-trips correctly on the donation side instead of being truncated.
- [x] Clear localStorage, then confirm a donate link carries `country=DE&locale=en`.

## Follow-ups
- #3078 - validate the callback URL scheme at ingestion. Encoding makes the value safe to transport; it does not check what kind of URL it is.
- #3079 - replace the seven positional params with an options object. The new characterization test makes that provable.


mohitb35 (changes requested):
@sunilsabatp

Functionally looks fine, but can we structure the url composition in a more maintainable way?

sunilsabatp:
> @sunilsabatp
> 
> Functionally looks fine, but can we structure the url composition in a more maintainable way?

Yep, I have refactored the URL composition to make it more maintainable and easier to extend.

mohitb35 (approved):
@sunilsabatp Thanks, the array-based composition is much easier to follow. I pushed three follow-ups on top rather than sending it back for another round:

1. **`country` and `locale` defaults** (2be1b19). These were sending the literal string `null` whenever localStorage had not been written yet, which happens on first paint and whenever the config fetch fails. They now default to `DE` and `en`. This is a behavior change, not encoding parity, so it is worth a look.

2. **Param types narrowed** (2be1b19). `embed`, `callbackUrl`, and `slug` were typed `string | string[]`, left over from when this helper read `router.query` itself. `useInitializeParams` already collapses repeated params before they reach the store, so the array case was unreachable. Narrowing to `string` makes it unreachable by type instead of by convention, and removes the three `String()` coercions.

3. **Tests** (14f27e6). `src/utils/getDonationUrl.test.ts`, 16 cases. The important one asserts that a callback URL carrying `?ref=abc&utm_source=news` stays a single param and cannot inject a `tenant` or `token` into the donation URL, which is the actual bug this PR fixes. There is also a characterization test pinning the exact output URL, so the next refactor here is provable by a green run.

Two things I deliberately left out and filed instead:

- #3078 - the callback URL scheme is still unvalidated. Encoding makes the value safe to transport, but nothing stops a `javascript:` value. Needs a check on what the donation app does with `callback_url` before we decide priority.
- #3079 - the seven positional params should become an options object. Out of scope here, and the new characterization test makes it safe to do later.

I have also updated the PR description to cover the extra changes. 

Approving the PR.

PR #3064, opened 2026-08-12
TEST - do not review, will be closed

mohitb35:
Throwaway draft PR testing whether `pull_request_target` fires when the workflow is on the base branch but not the default branch. `PR_LABELS_DRY_RUN=true` is set, so nothing can be written. Closing shortly.


mohitb35:
Test complete, closing. Findings recorded in #3063.

PR #3068, opened 2026-08-17
fix: update user profile store directly instead of triggering a refetch

### LINKED ISSUE #3038: I8 — Redundant profile GET after a profile update
Part of #3030 (follow-up epic) · tracked as **I8**.

**Priority:** Low · **Review ref:** redundant GET (ties to the store-shape cleanup) · **Links to:** A4 (store-shape cleanup)

**What is wrong**
The "set targets" flow sends a PUT to `/app/profile` and gets the **full updated profile back**
in the response. It then calls `refetchUserProfile()`, which fires a **second** request — a GET
to `/app/profile` — only to refresh the store. The GET is redundant, because the PUT response
already contains the updated profile.

**Where**
`src/features/user/Profile/ForestProgress/TargetsModal.tsx:102-109`

**How to reproduce**
1. Open a profile and change a target.
2. Watch the network tab: you will see a PUT and then a GET to `/app/profile`.
3. The whole profile page also briefly reloads (the extra GET flips the readiness flag and
   remounts the page — related to the readiness reshape in Part 3).

**Suggested fix**
Update the store directly from the PUT response (`setUserProfile(res)`) and drop the refetch.
This removes the extra round-trip and the page remount. Apply the same idea to the other callers
that refetch after a mutation (`RedeemModal`, `IssueCodesForm`), but first confirm each of their
endpoints returns the full profile. Best done as part of the store-shape cleanup (Part 3, A4).

**For the tracker** · Area: user-store · Labels: `enhancement`, `performance`, `priority: low` · Depends on: A4
**Done when:** the set-targets flow updates the store from its PUT response and fires no follow-up GET, and the profile page does not remount after a target change; the same pattern is applied to `RedeemModal` and `IssueCodesForm` where their endpoint returns the full profile.

sunilsabatp:
Resolves #3038 

## Summary

When saving targets, `TargetsModal` called `refetchUserProfile()`, which increments a counter and triggers a separate network request to fetch the profile again.

This is unnecessary because the save request response (`res`) already contains the updated profile. It also means the UI has to wait for a second round trip before the profile data is refreshed.

This change uses `setUserProfile(res)` to update the store directly with the response from the save request, matching the pattern already used in `EditProfileForm`, `EmbedModal`, and other places that update the profile after a mutation.

## Why

* Removes an unnecessary network request on every target save.
* Avoids a window where the store contains stale data while the refetch is in flight.
* Updates the UI immediately using the mutation response.

## Test Plan

* Open the Targets modal from the Forest Progress section.
* Update a target and save.
* Confirm the updated target values appear immediately without a visible delay.
* Confirm no additional profile fetch is triggered in the Network tab after saving.

### DIFF
--- a/src/features/user/Profile/ForestProgress/TargetsModal.tsx
+++ b/src/features/user/Profile/ForestProgress/TargetsModal.tsx
@@ -64,9 +64,7 @@ const TargetsModal = ({
   );
   //store: action
   const setUserInfo = useMyForestStore((state) => state.setUserInfo);
-  const refetchUserProfile = useUserStore(
-    (state) => state.refetchUserProfile
-  );
+  const setUserProfile = useUserStore((state) => state.setUserProfile);
   const setErrors = useErrorHandlingStore((state) => state.setErrors);
 
   const handleClose = () => {
@@ -106,7 +104,7 @@ const TargetsModal = ({
           payload,
         });
         const forestUserInfo = transformProfileToForestUserInfo(res);
-        refetchUserProfile();
+        setUserProfile(res);
         if (forestUserInfo !== undefined) {
           setUserInfo(forestUserInfo);
           setTreesPlantedTargetLocal(forestUserInfo.targets.treesDonated ?? 0);


mariahosfeld (commented):

mohitb35:
P2 - after merging child PRs.

PR #3069, opened 2026-08-17
refactor(user-store): route fetchUserProfile through the shared apiCl…

### LINKED ISSUE #3042: P2 — Route the profile fetch through the shared apiClient
Part of #3030 (follow-up epic) · tracked as **P2**.

**Review ref:** #29 · **Size:** small-medium · **Risk:** behavior-adjacent (review carefully) · **Links to:** A4 (folded into the store-shape cleanup)

**Scope**
`fetchUserProfile` calls `fetch()` directly and hand-rolls its own response and error handling,
instead of using the shared `apiClient` that every other request uses. Swap the raw `fetch` for
`apiClient`.

**What changes (not a pure no-op)**
- `apiClient` builds the `response.ok → APIError`, JSON parse, and base-URL logic in one place.
- The thrown error's body changes from a hardcoded string back to the server's payload (this
  actually **restores** the old `develop` behavior, which the migration had replaced with a
  constant string).
- The `!result` empty-check is dropped.

**Where**
`src/stores/userStore.ts:117-135`; shared client at `src/utils/apiRequests/apiClient.ts`.

**Note**
Even after this, the store still assembles the standard headers itself (that logic lives in the
`useApi` hook, which a store action cannot call) and lacks the browser-side token-expiry check
and API versioning. Fully unifying the two API paths is part of the store-shape cleanup (Part 3,
A4). This PR is the cheap first step.

**For the tracker** · Area: user-store · Labels: `refactor`, `priority: low` · Depends on: none
**Done when:** `fetchUserProfile` uses the shared `apiClient`; the thrown error carries the server payload; the success, 401, impersonation-403, and network-failure paths are verified unchanged.

sunilsabatp:
Fixes #3042

- Replace the raw `fetch()` call in `fetchUserProfile` (userStore.ts) with the shared `apiClient` utility, so error handling matches the rest of the codebase.
- Remove the redundant `!result` empty-response check, since `apiClient` already parses and returns the JSON body.
- Thrown errors now carry the server's actual response payload instead of the hardcoded `'Failed to fetch user profile'` string.

Header assembly and impersonation logic stay in the store as-is. Full API path unification (via `useApi`, token-expiry checks) is deferred to a later cleanup phase, per the linked issue.

Success, 401, impersonation-403, and network-failure paths are unchanged - only the transport and error-payload source changed.

### DIFF
--- a/src/stores/userStore.ts
+++ b/src/stores/userStore.ts
@@ -3,6 +3,7 @@ import type { ImpersonationData } from '../utils/impersonation';
 
 import { create } from 'zustand';
 import { devtools } from 'zustand/middleware';
+import apiClient from '../utils/apiRequests/apiClient';
 import getsessionId from '../utils/apiRequests/getSessionId';
 import { setHeaderForImpersonation } from '../utils/apiRequests/setHeader';
 import {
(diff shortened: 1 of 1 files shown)


mohitb35:
P2 - after P1 is complete.

Issue #3044, opened 2026-08-10
A1 — One owner for the "auth is ready" flag (the big one)


mohitb35:
Part of #3030 (follow-up epic) · tracked as **A1**.
Source: `docs/reviews/zustand-user-migration-followups.md`

**Phase 1 (do first).** **Linked items:** P3 (the selectors read this flag), I3 (the 401 redirect
path), I9 (token refresh).
**Fixes:** the mid-session flicker/remount, the stuck-loader, the cross-store write, and the
missing redirect-loop breaker. **Coordinate with:** PR 2878 (which rewrites the gate).

**The problem**
`isAuthResolved` (the "safe to show the app" flag) lives in `authStore`, but the thing that
flips it true for logged-in users is `fetchUserProfile` in `userStore`. So a user-store action
owns an auth-store flag, and it re-runs on **every** profile fetch. That single split causes
three bugs:
- **Mid-session flicker/remount:** every profile refetch flips the flag false then true, which
  makes the layout unmount and remount the whole user area. (Most visibly, it briefly destroyed
  the bulk-codes success screen before a targeted work-around was added.)
- **Stuck loader:** if the token comes back empty (no error, so no redirect), nothing sets the
  flag and the app sits on the loading spinner forever.
- **Cross-store write:** a user-store action reaching into the auth store.

**The reshape**
Move ownership of `isAuthResolved` into the **auth layer** (`useInitializeAuth`):
- `fetchUserProfile` stops touching `isAuthResolved` entirely.
- The auth layer sets it false → true **once**, after the token is obtained **and** the first
  profile attempt has settled — even when the token is empty.

**Keep this meaning exactly**
The flag must mean **"auth is resolved AND the first profile attempt is done"**, not "a token
exists". This matters because the layout redirects to `/login` when `isAuthResolved &&
!userProfile`. If the flag turned true the instant a token arrived (before the first profile
loaded), a logged-in user would be bounced to login while their profile was still loading.

**Also include here: the missing redirect-loop breaker**
There is a second, related bug that belongs in this PR because it touches the same gate logic.
The app can bounce a user to `/login` for two reasons: (1) it cannot get a token, and (2) a
profile request returns 401. Only case (1) has a "stop after 3 tries" counter. A persistent
case (2) — for example a revoked account — loops forever with no brake.

The fix is not "reuse the same counter", because that counter resets every time a token is
[...]

PR #3070, opened 2026-08-17
fix(auth): show a failed sign-in screen instead of an endless loader

### LINKED ISSUE #3034: I4 — Token is not cleared from the store on logout
Part of #3030 (follow-up epic) · tracked as **I4**.

**Priority:** Low · **Review ref:** #16 · **Links to:** A5 (one exit policy)

**What is wrong**
`logoutUser` exits impersonation and clears some storage but never calls `setToken(null)`. It
relies on the Auth0 full-page redirect to wipe memory.

**Where**
`src/hooks/useAuthSession.ts:16-24`

**Why it matters**
Low today, because the token lives only in memory and the redirect clears it. But if any logout
path ever does not fully navigate away, a stale token lingers in the store.

**Suggested fix**
Call `setToken(null)` inside `logoutUser` so the store is cleared regardless of navigation.

**For the tracker** · Area: auth · Labels: `bug`, `tech-debt`, `priority: low` · Depends on: none
**Done when:** `logoutUser` calls `setToken(null)`; the store token is null after logout even without a full page navigation.

mohitb35:
Stacked on #3055. Base is `fix/auth-resolution-gate` — do not merge before #3055, then retarget to `develop`.

## Summary

When the auth redirect guard gives up, the app now shows a "Sign in failed" screen with a way out, instead of a loader that never resolves.

## Why

#3044 asks for three things in its redirect-loop breaker: reset the counter on profile success, route the profile 401 through the guarded redirect, and decide where the user lands when the brake trips. #3055 does the first two. This PR does the third.

Without it, the brake stops the loop but leaves a dead end. `isAuthResolved` is true with no profile, so `UserLayout` pushes to `/login`, and `login.tsx` returns early and renders the loader skeleton forever. A reload does not help, because the counter lives in `sessionStorage`, so the user is stuck until they close the tab.

## What changed

- New `hasAuthFailed` flag in `authStore`, set when the redirect limit trips on a token failure or a profile 401, and when the access token comes back empty.
- New `AuthFailed` screen, rendered from `Layout` in place of the page content. It sits in `Layout` because the profile fetch runs globally, so a sign-in can fail on any page, and only `UserLayout` pages redirect to `/login`.
- Two actions, both logging out first so the next attempt starts clean: **Go to home page**, and **Try signing in again**, which lands signed out on `/login` and reuses the same `loginWithRedirect` the header's sign-in button calls.
- `logoutUser` now clears the token and the redirect count. Clearing the count matters, or a stale 3 would block the retry before it tried once.
- Four new English locale keys. Non-en comes from Lingohub.

## Screenshot of new UI
Screenshot 2026-08-17 at 1 39 21 PM

## Requires an Auth0 change (done)

`${origin}/login` must be in the tenant's **Allowed Logout URLs**, or "Try signing in again" lands on Auth0's error page instead of returning.

## How to test (tested locally)

A persistent profile 401 is needed, since Auth0 keeps issuing valid tokens here.

1. Temporarily change the header in `userStore.ts` to `Authorization: \`Bearer ${token}invalid\``.
2. Clear `sessionStorage`, log in, land on `/profile`.
3. Watch three bounces through Auth0, with `authRedirectCount` going 1, 2, 3.
4. On the fourth 401 the screen appears. Before this PR you would get the loader skeleton on `/login`.
5. Check both buttons leave you signed out, one on `/`, one at a fresh Auth0 prompt.
6. Remove the patch and confirm normal login, logout, and the 303 complete-signup redirect still work.

Automated checks pass: 8 new unit tests for `authRedirectGuard` (29 in the full suite), no new type errors, eslint and prettier clean. The manual steps above are not yet done, and overlap with the three unticked boxes on #3055.

## Notes

- No auto-redirect timer. A timed redirect is a time limit under WCAG 2.2.1 (Level A) and would need a way to turn it off; the button does not count, since it acts early rather than extending. Add a countdown with a "Stay on this page" control later if people sit on the screen.
- Focus moves to the heading on mount, because the content swaps without a navigation and assistive tech would not announce it.

## Related

- Closes #3034 (token not cleared on logout), via the `logoutUser` change.
- Together with #3055 this covers #3044's redirect-loop breaker. #3044 stays open for removing the bulk-codes flicker workaround, which is best done in the #2878 rebase since that PR owns the file.


sunilsabatp (approved):
LGTM

PR #3073, opened 2026-08-19
fix(api): stop impersonation headers from leaking onto public requests

### LINKED ISSUE #3031: I1 — Impersonation headers are sent on requests that do not need login
Part of #3030 (follow-up epic) · tracked as **I1**.

**Priority:** Low · **Review ref:** #8 · **Links to:** A5 (one impersonation owner), A6 (keep it off the hot path)

**What is wrong**
While a support agent is impersonating a user, the app adds the impersonation headers
(`X-SWITCH-USER` = the target's email, `X-USER-SUPPORT-PIN`) to **every** API request, including
public ones that do not need a logged-in user.

**Why**
`setHeaderForImpersonation` reads `impersonationData` from `localStorage` as a fallback even when
the caller passes nothing, and `useApi` calls it for the unauthenticated request helpers too.

**Where**
- `src/utils/apiRequests/setHeader.ts:9-12`
- `src/hooks/useApi.ts:143`

**How to reproduce**
1. Start impersonating a user.
2. Open the browser network tab.
3. Watch a public/unauthenticated request — it also carries `X-SWITCH-USER` and
   `X-USER-SUPPORT-PIN`.

**Why it matters**
The target's email and support pin leak onto requests that never needed them. If leftover
impersonation data stays in `localStorage`, this can even fire when the user is not actively
impersonating.

**Suggested fix**
Only attach the impersonation headers to authenticated requests, or only when the caller
explicitly passes impersonation data. Do not use the `localStorage` fallback for public helpers.

**For the tracker** · Area: impersonation · Labels: `bug`, `security`, `priority: low` · Depends on: none
**Done when:** impersonation headers are attached only to authenticated requests (or only when the caller passes impersonation data); a network trace during impersonation shows no `X-SWITCH-USER` or `X-USER-SUPPORT-PIN` on public requests.

sunilsabatp:
Fixes #3031

## What

While a support agent is impersonating a user, `callApi` (in `useApi.ts`) added `X-SWITCH-USER` and `X-USER-SUPPORT-PIN` to every request, including public/unauthenticated ones that never needed a logged-in user.

`setHeaderForImpersonation` reads the stored impersonation credentials from `localStorage` as a fallback whenever the caller passes nothing, and `useApi` called it unconditionally for both the authenticated and public request helpers.

## Change

`callApi` now only runs the header through `setHeaderForImpersonation` when `authRequired` is true:

```ts
const finalHeader = authRequired
  ? setHeaderForImpersonation(headers)
  : headers;
```

Public requests (`getApi`, `postApi`, `putApi`) get the plain header object and never touch impersonation state. Authenticated requests (`getApiAuthenticated`, `postApiAuthenticated`, `putApiAuthenticated`, `deleteApiAuthenticated`) are unchanged.

## Why

The target user's email and support pin were leaking onto requests that had no reason to carry them. Because the fallback reads straight from `localStorage`, this could also fire on any lingering `impersonationData` even when the agent believed impersonation had ended.

## Testing

- `npx vitest run`: 21 tests pass (unchanged, this PR does not touch `setHeader.ts`, so its existing coverage still applies to the authenticated path).
- `npx tsc --noEmit`: no new errors introduced by this change.
- `npx eslint src/hooks/useApi.ts`: clean.
- Manual repro before/after, seeding `localStorage.impersonationData` and reloading:
  - Before: the public `GET /app/currencies` request (fired on every load by `useInitializeCurrency`) carried `X-SWITCH-USER` / `X-USER-SUPPORT-PIN`.
  - After: the same request carries only `x-locale`, `tenant-key`, `X-SESSION-ID`.
  - Authenticated requests (e.g. profile fetch) still carry both impersonation headers when impersonation data is present.

No behavior change for authenticated requests or for the impersonation feature itself.

### DIFF
--- a/src/hooks/useApi.ts
+++ b/src/hooks/useApi.ts
@@ -141,7 +141,11 @@ export const useApi = () => {
       }
       headers.Authorization = `Bearer ${token}`;
     }
-    const finalHeader = setHeaderForImpersonation(headers);
+    // Impersonation headers (and the localStorage read behind them) only apply to
+    // authenticated requests, so public requests never carry the target's email/pin.
+    const finalHeader = authRequired
+      ? setHeaderForImpersonation(headers)
+      : headers;
     const requestOptions =
       method === 'POST' || method === 'PUT'
         ? { method, url, data, queryParams, additionalHeaders: finalHeader }


mohitb35:
P3 - along with A5

Issue #11, opened 2020-07-13
Data fetching from API and backend


harshvitra:
Fetching all the data from the backend using the APIs. 

NextJS has a different way of fetching & loading the data find more info here - 
https://nextjs.org/docs/basic-features/data-fetching

PR #3076, opened 2026-08-19
fix(user): split RedeemPopup effect dependency

### LINKED ISSUE #3032: I2 — RedeemPopup effect uses a combined boolean as a dependency
Part of #3030 (follow-up epic) · tracked as **I2**.

**Priority:** Low · **Review ref:** #11 · **Links to:** P3 (RedeemPopup is one of the group-2 files)

**What is wrong**
An effect's dependency array holds one combined value instead of the two separate values:
```ts
}, [isAuthResolved && userProfile]);
```
React compares the single collapsed result, so the effect may not re-run when it should.

**Where**
`src/features/common/Layout/RedeemPopup/index.tsx:35`

**How to reproduce**
Hard to trigger by hand (it is a timing bug). It is clear from reading the code.

**Why it matters**
The effect can miss the moment the user becomes available, so it runs with stale data or does
not run at all.

**Suggested fix**
List the values separately:
```ts
}, [isAuthResolved, userProfile]);
```

**For the tracker** · Area: user · Labels: `bug`, `priority: low` · Depends on: none
**Done when:** the effect lists `isAuthResolved` and `userProfile` as separate dependencies and re-runs when either one changes.

sunilsabatp:
## Summary
- `RedeemPopup`'s effect watched `[isAuthResolved && userProfile]`, a single collapsed boolean, so React's dependency comparison could miss the moment `userProfile` actually changed and the effect could run with stale data or not at all.
- Split it into `[isAuthResolved, userProfile]` so each value is compared separately, matching how the effect body already uses both.

This addresses item I2 from the Zustand user/auth migration follow-up list (Part 1, Review ref #11).

## Test plan
- [ ] Sign in and confirm the redeem popup still hides once the profile loads
- [ ] `npm run lint` / `npm run build` pass

### DIFF
--- a/src/features/common/Layout/RedeemPopup/index.tsx
+++ b/src/features/common/Layout/RedeemPopup/index.tsx
@@ -32,7 +32,7 @@ export default function RedeemPopup() {
     if (isAuthResolved && userProfile) {
       setShowRedeemPopup(false);
     }
-  }, [isAuthResolved && userProfile]);
+  }, [isAuthResolved, userProfile]);
 
   const isMountedRef = useRef(false);
 


mohitb35 (approved):
LGTM

mohitb35 (approved):
LGTM

PR #3077, opened 2026-08-19
fix(profile): log the real API error cause instead of a generic message

### LINKED ISSUE #3037: I7 — Profile error logs print "Something went wrong" instead of the real cause
Part of #3030 (follow-up epic) · tracked as **I7**.

**Priority:** Low · **Review ref:** #3 (log gap) · **Links to:** A2 (the error shape upstream)

**What is wrong**
The profile error handler logs `err.message`, but the `APIError` class always hardcodes
`.message` to the string `"Something went wrong"` and puts the real text in `.errors`. So these
two log lines always print "Something went wrong", whatever actually failed.

**Where**
`src/hooks/useProfileErrorHandler.ts:67, 72`

**Note**
The caller log at `src/hooks/useInitializeUser.ts:40` logs the raw error object, so the real
cause **is** available there (for example `TypeError: Failed to fetch`). Only these two handler
logs are unhelpful.

**Why it matters**
Low. It is a debugging annoyance, not a functional bug.

**Suggested fix (not a one-liner)**
It cannot be fixed by just reading `err.errors.message`, because different errors store their
text differently (HTTP errors put a plain string in `.errors`; wrapped network errors put an
object `{ message }`), and the `APIError.errors` type does not expose `.message`. The clean fix
touches the error shape — either a purpose-built error type, or building the wrapped errors as
`{ code: 0, message }`. The root cause is upstream: `errors.d.ts` in `@planet-sdk/common`
hardcodes `.message`.

**For the tracker** · Area: user · Labels: `bug`, `tech-debt`, `priority: low` · Depends on: A2
**Done when:** the two profile-error logs print the real failure cause instead of "Something went wrong".

sunilsabatp:
## Summary

- `useProfileErrorHandler` logged `err.message` for the 500 and default cases, but `APIError.message` is hardcoded by `@planet-sdk/common` to the literal string `"Something went wrong"`, so both log lines always printed that instead of the actual failure.
- The real cause lives in `err.errors`, which `fetchUserProfile` (in `userStore.ts`) builds in two different shapes: a plain string for HTTP failures (e.g. `'Failed to fetch user profile'`), or an object `{ message }` for wrapped non-API errors (e.g. network failures).
- Added a local `getProfileErrorMessage` helper that reads `err.errors` and handles both shapes, falling back to `err.message` only if neither matches, and used it in place of the two `err.message` log calls.

This only changes `console.error` output for debugging. Nothing here is rendered in the UI: the 401 redirect-limit path shows a static, hardcoded `AuthFailed` screen that never reads `profileApiError` or any error text.

Fixes #3037

## Test plan

- [x] `npx tsc --noEmit` shows no errors introduced in `useProfileErrorHandler.ts`
- [x] `npx eslint src/hooks/useProfileErrorHandler.ts` passes clean
- [ ] Manually trigger a profile 500 and a network failure and confirm the console now prints the real cause instead of "Something went wrong"

### DIFF
--- a/src/hooks/useProfileErrorHandler.ts
+++ b/src/hooks/useProfileErrorHandler.ts
@@ -11,6 +11,24 @@ import {
   registerRedirectAttempt,
 } from '../utils/authRedirectGuard';
 
+// `APIError.message` is always hardcoded to "Something went wrong" by the SDK.
+// The real failure text lives in `.errors`, which is a plain string for
+// HTTP errors and an object with a `message` field for wrapped errors
+// (see fetchUserProfile in userStore.ts). Fall back to `err.message`
+// only if neither shape matches.
+const getProfileErrorMessage = (err: APIError): string => {
+  if (typeof err.errors === 'string') return err.errors;
+  if (
+    err.errors &&
+    typeof err.errors === 'object' &&
+    'message' in err.errors &&
+    typeof err.errors.message === 'string'
+  ) {
+    return err.errors.message;
+  }
+  return err.message;
+};
+
 const useProfileErrorHandler = () => {
   const router = useRouter();
   const locale = useLocale();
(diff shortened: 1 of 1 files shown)


mohitb35:
Could relate to #2927 as well - cross check

mohitb35:
Heads-up on overlap: the draft onboarding PR #2927 adds `src/utils/parseApiError.ts`, which does the same "dig the real cause out of `APIError.errors`" job as `getProfileErrorMessage` here. Coverage differs: `parseApiError` handles the `{error}` 409 shape but not the `{message}` network shape this helper relies on, so neither replaces the other today. Flagging so we can converge on one shared util later rather than keep two. No change needed to merge this.

PR #3093, opened 2026-08-26
Migrate donation receipt state to Zustand with persist middleware

### LINKED ISSUE #3090: Donation receipt store: replace manual sessionStorage writes with the persist middleware
`donationReceiptStore` saves to sessionStorage by hand, once per action. The old `DonationReceiptContext` saved from an effect, so every state change was saved whatever caused it.

Nothing is broken today. This is about how the guarantee is held.

## What the old code did

```js
useEffect(() => {
  sessionStorage.setItem('donationReceiptContext', JSON.stringify(state));
}, [state]);
```

The effect could not be forgotten.

## What the store does

`persistState(get())` runs at the end of three actions: `initForVerification`, `initForIssuance` and `updateDonorAndAddress`.

Coverage is currently complete. Checked at the time of PR #2963:

1. Those three actions are the only ones that change state.
2. `initializeFromSession` writes state but has nothing new to save.
3. Nothing calls `useDonationReceiptStore.setState` directly.

## Why it is worth changing

The guarantee moved from structural to conventional. A fourth action that forgets the call would save nothing and would fail quietly, because the in-memory state stays correct for the rest of the session. It would only show up on a reload.

## Suggested change

Use Zustand's `persist` middleware with `partialize`. That moves the saving into the middleware instead of each action, and removes `getDataState`, which is the third place all 21 field names are listed after the interface and `defaultState`.

To be fair to the current code, that triplication is type safe: `getDataState` is annotated as returning `DonationReceiptState`, so leaving a field out is a compile error rather than a silent gap.

## Do this together with the hydration timing

`persist` also changes **when** the saved receipt is read, so this overlaps the hydration point raised on PR #2963. Doing them separately means touching the same timing twice.

Taken together, `persist` plus its `onRehydrateStorage` callback provides the `isHydrated` flag that the hydration fix needs, so the two changes are cheaper as one.

## Done when

- Saving happens in middleware, not in individual actions.
- `getDataState` is gone, or exists only as `partialize`.
- A new action needs no extra step to be persisted.

## Related

PR #2963, where the store was ported from React Context. Raised in review and deferred deliberately.

sunilsabatp:
Fixes : #3090 

## Summary
- Migrates `DonationReceiptContext` (React Context + manual `sessionStorage` read/write) to a Zustand store, following the same pattern already used for other stores in the app.
- Replaces the manual `persistState(get())` calls scattered across individual actions with Zustand's `persist` middleware, so persistence and hydration are handled centrally and any future action persists automatically without extra code.
- `skipHydration` is set since `sessionStorage` doesn't exist during SSR; hydration is triggered manually from `useInitializeDonationReceipt` on mount via `StoreInitializer`.
- `partialize` builds the persisted slice from `defaultState`'s own keys (an allowlist) rather than destructuring out known non-data fields (a denylist), so a future store-only field (an action, a flag, etc.) can't accidentally leak into `sessionStorage`.
- `onRehydrateStorage` clears storage and still marks the store hydrated if the stored JSON is corrupted, so the app never gets stuck on a broken cache.
- `clearSessionStorage` now uses `persist.clearStorage()` instead of `sessionStorage.removeItem`, same effect, but goes through the middleware.

## Test plan
- [ ] Fresh load with no prior session shows the skeleton then loads normally, no console errors.
- [ ] Refreshing mid-flow (verification or issuance) restores donor, address, and donation data from `sessionStorage`.
- [ ] Manually corrupting the `donationReceiptContext` sessionStorage entry and reloading does not crash the app; storage is cleared and the page falls back to a fresh load.
- [ ] Switching identity in the same tab (e.g. exiting impersonation) clears the previous donor's `sessionStorage` entry and does not leak it to the new user.
- [ ] Direct link to an issuance-type receipt still renders for a non-owner (ownership check is bypassed for `RECEIPT_STATUS.ISSUE`).
- [ ] Editing donor/address data persists to `sessionStorage` immediately without a separate save step.


mariahosfeld (commented):
Refresh of pre-verified donation receipt renders error "no access" -> no impersonation used.

PR #3094, opened 2026-08-27
Improve the skeleton UI across the app

sunilsabatp:
## Summary
- Add a generic `SkeletonBlock` primitive and per-route skeleton loading shapes, replacing the generic `react-loading-skeleton` placeholder used across the app.
- Cover UserLayout, bulk-codes, manage-payouts, manage-projects, settings/single-form pages, and the project list and project details pages under projectsV2.
- Each skeleton mirrors the shape of the real content it precedes (cards, sections, form rows) so layout doesn't shift once data loads, and mobile variants match their real component's responsive styles.

## Test plan
- [ ] Verify each route's skeleton renders correctly on both desktop and mobile viewports: user layout shell, bulk-codes, manage payouts, manage projects, settings/single-form pages, project list, and project details.
- [ ] Confirm the skeleton-to-content transition doesn't cause layout shift on slow network throttling.
- [ ] Run `npm run lint` and `npm run build`.

### DIFF
--- a/pages/sites/[slug]/[locale]/profile/widgets.tsx
+++ b/pages/sites/[slug]/[locale]/profile/widgets.tsx
@@ -12,6 +12,7 @@ import { useState, useEffect } from 'react';
 import UserLayout from '../../../../../src/features/common/Layout/UserLayout/UserLayout';
 import EmbedModal from '../../../../../src/features/user/Widget/EmbedModal';
 import styles from '../../../../../src/features/common/Layout/UserLayout/UserLayout.module.scss';
+import SettingsFormSkeleton from '../../../../../src/features/user/Settings/SettingsFormSkeleton';
 import Head from 'next/head';
 import { useTranslations } from 'next-intl';
 import {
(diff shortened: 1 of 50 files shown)


pages/sites/[slug]/[locale]/profile/widgets.tsx
```diff
@@ -41,7 +42,7 @@ function ProfilePage(): ReactElement {
   if (!isInitialized) return <></>;
   // TO DO - change widget link
   return (
-    <UserLayout>
+    <UserLayout skeleton={<SettingsFormSkeleton variant="none" />}>
```
Comment (mariahosfeld) [inline pages/sites/[slug]/[locale]/profile/widgets.tsx:45]: Check behavior of loader, elements of the skeleton move from center to (correctly) aligned left for the widget during loading.

mariahosfeld (commented):
Should the loader for the donation receipt section also be included in the new format?

sunilsabatp:
> Should the loader for the donation receipt section also be included in the new format?

Yes, I think the donation receipt loader should use the new format too, just to keep things consistent.

PR #3095, opened 2026-08-27
Refactor getDonationUrl to take a single options object

### LINKED ISSUE #3079: Refactor getDonationUrl to take a single options object
`getDonationUrl` takes seven positional parameters. Callers that need a later one have to pad the earlier ones with `undefined`, and the order is not guessable from a call site.

Raised while reviewing PR #3057.

## Current signature

[src/utils/getDonationUrl.ts](../../src/utils/getDonationUrl.ts):

```ts
export const getDonationUrl = (
  tenant: string | undefined,
  id: string,
  token: string | null,
  embed?: string,
  callbackUrl?: string,
  slug?: string,
  utmCampaign?: string
): string
```

## The observed pain

Two of the four call sites pad with placeholders.

[ProjectInfoSection.tsx](../../src/features/projectsV2/ProjectSnippet/microComponents/ProjectInfoSection.tsx) skips `slug` to reach `utmCampaign`:

```ts
const donateLink = getDonationUrl(
  tenantId,
  slug,
  token,
  embed || undefined,
  callbackUrl || undefined,
  undefined,
  utmCampaign || undefined
);
```

[DonateButton.tsx](../../src/features/user/Profile/MyContributions/DonateButton.tsx) skips `embed` and `callbackUrl` to reach `slug`:

```ts
const donateLink = getDonationUrl(
  tenantId,
  projectSlug,
  token,
  undefined,
  undefined,
  type === 'supported' ? props.supportedTreecounter : undefined
);
```

Note also that the second parameter is named `id` but receives a project slug at three call sites and the literal `'planetcash'` at the other two. And `slug` is a different thing again: the direct-gift recipient slug for the `s` param. Positional order makes that confusion easy to miss; named fields make it visible.

## Proposal

```ts
type DonationUrlOptions = {
  tenant?: string;
  /** Project slug, or 'planetcash' to top up a PlanetCash balance. */
  target: string;
  token?: string | null;
  embed?: string;
  callbackUrl?: string;
  /** Direct-gift recipient slug, used for the `s` param. */
  recipientSlug?: string;
  utmCampaign?: string;
};

export const getDonationUrl = (options: DonationUrlOptions): string
```

Call sites become order-independent and the placeholders disappear:

```ts
const addBalanceLink = getDonationUrl({ tenant: tenantId, target: 'planetcash', token });
```

## Scope

Four call sites: [BulkCodesError.tsx](../../src/features/user/BulkCodes/components/BulkCodesError.tsx), [AccountDetails.tsx](../../src/features/user/PlanetCash/components/AccountDetails.tsx), [DonateButton.tsx](../../src/features/user/Profile/MyContributions/DonateButton.tsx), [ProjectInfoSection.tsx](../../src/features/projectsV2/ProjectSnippet/microComponents/ProjectInfoSection.tsx). One file plus its test.

The internal query-param array and `join('&')` do not need to change. This is about the boundary only.

## Why now

[src/utils/getDonationUrl.test.ts](../../src/utils/getDonationUrl.test.ts) landed in #3057 and includes a characterization test pinning the exact output URL and param order. That makes this refactor provable by a green run instead of a line-by-line read, which is exactly the case [docs/plans/testing-roadmap.md](../../docs/plans/testing-roadmap.md) describes. Renaming the parameters means updating the test's call sites, but the asserted output URLs should not change at all. Any diff in an expected URL means the refactor changed behaviour.

sunilsabatp:
## Summary
- Converts `getDonationUrl` from seven positional parameters to a single `DonationUrlOptions` object, so callers no longer need to pad earlier params with `undefined` to reach a later one.
- Renames `id` to `target` (it's a project slug or the literal `'planetcash'`) and `slug` to `recipientSlug` (the direct-gift recipient slug used for the `s` param) to remove the ambiguity flagged in the issue.
- Updates the four call sites (`DonateButton.tsx`, `AccountDetails.tsx`, `BulkCodesError.tsx`, `ProjectInfoSection.tsx`) to pass named fields.
- Internal query-param array and `join('&')` composition are untouched, as scoped by the issue.

Closes #3079

## Test plan
- [x] `getDonationUrl.test.ts` updated to the new call signature; the characterization test's asserted output URL is unchanged, proving the refactor is behavior-preserving
- [x] `npx vitest run` — all 45 tests pass
- [x] `npx tsc --noEmit` — no new errors introduced by this change
- [x] `npx eslint` and `npx prettier --check` clean on all changed files

### DIFF
--- a/src/features/projectsV2/ProjectSnippet/microComponents/ProjectInfoSection.tsx
+++ b/src/features/projectsV2/ProjectSnippet/microComponents/ProjectInfoSection.tsx
@@ -40,15 +40,14 @@ const ProjectInfoSection = (props: ProjectInfoProps) => {
   const callbackUrl = useQueryParamStore((state) => state.callbackUrl);
   const token = useAuthStore((state) => state.token);
 
-  const donateLink = getDonationUrl(
-    tenantId,
-    slug,
+  const donateLink = getDonationUrl({
+    tenant: tenantId,
+    target: slug,
     token,
-    embed || undefined,
-    callbackUrl || undefined,
-    undefined,
-    utmCampaign || undefined
-  );
+    embed: embed || undefined,
+    callbackUrl: callbackUrl || undefined,
+    utmCampaign: utmCampaign || undefined,
+  });
   const donationLabel = useMemo(() => {
     if (unitCount === undefined) {
       return;
--- a/src/features/user/BulkCodes/components/BulkCodesError.tsx
+++ b/src/features/user/BulkCodes/components/BulkCodesError.tsx
@@ -34,7 +34,11 @@ const BulkCodesError = (): ReactElement | null => {
     }
 
     if (planetCash.balance + planetCash.creditLimit <= 0) {
-      const donationUrl = getDonationUrl(tenantId, 'planetcash', token);
+      const donationUrl = getDonationUrl({
+        tenant: tenantId,
+        target: 'planetcash',
+        token,
+      });
 
       return (
         <div>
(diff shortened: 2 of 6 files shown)


mariahosfeld (commented):

PR #3096, opened 2026-08-27
refactor(auth): shared is-the-user-ready selectors

### LINKED ISSUE #3043: P3 — Shared "is the user ready" selectors
Part of #3030 (follow-up epic) · tracked as **P3**.

**Review ref:** #21 · **Size:** ~15 files · **Risk:** low if done carefully (see warning) · **Links to:** A1 (the flag these selectors read)

**Scope**
Some form of an "is the user ready" check is copy-pasted inline across the app. Two clean groups
already use the same wording and are the safe targets for a selector. A third set of call sites
uses `isAuthResolved` in **near-variant** or **different** ways and needs judgment per site, not a
blind swap.

**Group 1 — "auth ready for an API call"** = `token !== null && isAuthResolved` — 9 files, all
written as the same store selector:
`useAddressOperations` (line 46), `ApiKeyForm` (49), `ProjectSelector` (35), `EmbedModal` (38),
`Transactions` (38), `PlanetCash/index` (46), `TargetsModal` (63), `ProjectsContainer` (136),
`ManagePayouts/index` (55).

**Group 2 — "profile ready"** = `isAuthResolved && userProfile` — 5 files:
`ProfileLayout` (41), `DonationReceipts` (59), `RedeemModal` (61), `CookiePolicy` (19),
`RedeemPopup` (32).

**Extra sites (read each before touching — do not fold blindly into a group):**
- **Same meaning as group 1, but written inline** as `isAuthResolved && token` instead of the
  selector: `EditProfileForm` (lines 208, 248). These *can* move to `useIsAuthReady()`, but the
  wording differs, so a plain find-and-replace will miss them. Convert by hand.
- **`isAuthResolved` used on its own** (no `token`, no `userProfile`): `BulkCodes/index` (116,
  paired with `planetCashAccount`), `CompleteSignup` (92), `UserLayout` (282). These are not a
  readiness check for a request or a profile; leave them, or give them a plain
  `useIsAuthResolved()` if that reads better. Do not force them into either selector.
- **`isAuthResolved && !userProfile` — the login-redirect gate** (resolved AND *no* profile):
  `UserLayout` (271, `!isAuthResolved || userProfile`), `CompleteSignup` (169,
  `!isAuthResolved || userProfile || !token`). This looks like group 2 but means the **opposite**:
  "auth is settled and there is still no profile, so send the user to sign in". Never replace this
  with `useIsProfileReady()`.
- **Group 1 plus extra conditions:** `CompleteSignup` (146,
  `!agreedToTerms || !country || !isAuthResolved || !token`). The auth-ready part can use the
  selector, but the extra form conditions stay. Convert by hand.

**Change**
Create two selectors — `useIsAuthReady()` (group 1) and `useIsProfileReady()` (group 2) — and
migrate those two groups. For the extra sites, decide case by case as noted above; a small
`useIsAuthResolved()` for the "on its own" set is optional.

**Warning (this is the whole point)**
Do **not** collapse the checks into one selector. Two failure modes to avoid:
1. Replacing a "profile is present" check (group 2) with a "token is present" check (group 1): a
   screen that should wait for the profile would proceed as soon as a token exists.
2. Replacing the login-redirect gate (`isAuthResolved && !userProfile`) with `useIsProfileReady()`:
   the gate flips to the opposite meaning and can bounce a signed-in user to login.

**Why its own PR**
It is a pure refactor touching ~15 files; it reads cleanly on its own and should not be hidden
inside a larger change.

**Sequencing note**
Do this after the four stacked module PRs (2873 PlanetCash, 2874 Payout, 2878 Bulk-codes, 2963
Donation receipt) merge. Each rewrites the readiness checks in its own files, which are the same
files this PR edits. Running P3 first means conflicts and rework; running it after lets you
re-derive the file list once against the settled code. The line numbers above will shift, so
confirm them before starting.

**For the tracker** · Area: auth · Labels: `refactor`, `priority: low` · Depends on: A1; the four module PRs (2873, 2874, 2878, 2963) merged first
**Done when:** group 1 files use `useIsAuthReady()`, group 2 files use `useIsProfileReady()`, the extra sites are handled per the notes, and the login-redirect gate is unchanged.

sunilsabatp:
## Summary
Closes #3043.

- Adds `useIsAuthReady()` and `useIsProfileReady()` in `src/hooks/useAuthReadiness.ts`.
- Migrates the 9 group 1 call sites (`token !== null && isAuthResolved`) to `useIsAuthReady()`.
- Migrates the 4 group 2 call sites (`isAuthResolved && userProfile`) to `useIsProfileReady()`. `RedeemPopup` from the issue was already removed from the codebase, so its site no longer exists.
- Hand-converts the two extra sites written in near-variant wording (`EditProfileForm.tsx`, `CompleteSignup/index.tsx` line 146) to `useIsAuthReady()`.
- Leaves the standalone `isAuthResolved` checks (`BulkCodes/index.tsx`, `CompleteSignup/index.tsx` line 92, `UserLayout.tsx` line 282) and the login-redirect gates (`UserLayout.tsx` line 271, `CompleteSignup/index.tsx` line 169) untouched, per the issue's explicit warning against folding those into either selector.

## Test plan
- [x] `tsc --noEmit` shows no new errors introduced by this change (compared the same run against `develop`).
- [ ] Manually verify in the browser: profile page load, address management, API key screen, PlanetCash, payouts, bulk codes, targets modal, cookie notice, redeem modal, donation receipts, complete-signup flow, and the login-redirect behavior for a signed-in vs signed-out user.

### DIFF
--- a/src/features/common/Layout/CookiePolicy/index.tsx
+++ b/src/features/common/Layout/CookiePolicy/index.tsx
@@ -4,22 +4,20 @@ import IconButton from '../../IconButton';
 import styles from './CookiePolicy.module.scss';
 import { useLocale, useTranslations } from 'next-intl';
 import themeProperties from '../../../../theme/themeProperties';
-import { useAuthStore, useUserStore } from '../../../../stores';
+import { useIsProfileReady } from '../../../../hooks/useAuthReadiness';
 
 export default function CookiePolicy() {
   const [showCookieNotice, setShowCookieNotice] = useState(false);
   const t = useTranslations('Common');
   const locale = useLocale();
 
-  //store: state
-  const userProfile = useUserStore((state) => state.userProfile);
-  const isAuthResolved = useAuthStore((state) => state.isAuthResolved);
+  const isProfileReady = useIsProfileReady();
 
   useEffect(() => {
-    if (isAuthResolved && userProfile) {
+    if (isProfileReady) {
       setShowCookieNotice(false);
     }
-  }, [isAuthResolved, userProfile]);
+  }, [isProfileReady]);
 
   const isMountedRef = useRef(false);
 
--- a/src/features/user/BulkCodes/components/ProjectSelector.tsx
+++ b/src/features/user/BulkCodes/components/ProjectSelector.tsx
@@ -6,11 +6,8 @@ import ProjectSelectAutocomplete from '../../../common/ProjectSelectAutocomplete
 import UnitCostDisplay from './UnitCostDisplay';
 import { handleError } from '@planet-sdk/common';
 import { useApi } from '../../../../hooks/useApi';
-import {
-  useAuthStore,
-  useUserStore,
-  useErrorHandlingStore,
-} from '../../../../stores';
+import { useIsAuthReady } from '../../../../hooks/useAuthReadiness';
+import { useUserStore, useErrorHandlingStore } from '../../../../stores';
 import { useBulkCodeStore } from '../../../../stores/bulkCodeStore';
 import { useTranslations } from 'next-intl';
 
@@ -26,10 +23,8 @@ const ProjectSelector = ({
 }: ProjectSelectorProps): ReactElement | null => {
   const tBulkCodes = useTranslations('BulkCodes');
   const { getApiAuthenticated } = useApi();
+  const isAuthReady = useIsAuthReady();
   //store: state
-  const isAuthReady = useAuthStore(
-    (state) => state.token !== null && state.isAuthResolved
-  );
   const userProfile = useUserStore((state) => state.userProfile);
   const projectList = useBulkCodeStore((state) => state.projectList);
   const planetCashAccount = useBulkCodeStore(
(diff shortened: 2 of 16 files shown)


mariahosfeld (commented):

mohitb35:
Note on one more call site, which this PR will need to pick up.

`feature/project-onboarding` (#2927) adds `src/features/user/DueDiligence/index.tsx`, which hand-rolls the group 1 check inline:

```ts
const isAuthReady = useAuthStore(
  (state) => state.token !== null && state.isAuthResolved
);
```

That file is not on `develop` yet, which is why it is absent from this PR's file list.

**#2927 is landing first.** So by the time this PR merges, `DueDiligence/index.tsx` will be on `develop` and will be the one remaining inline copy of the group 1 selector. Worth adding it to the conversions here, so #3043 closes with nothing left behind.

I test-merged this PR against #2927 and it is clean either way, including `ProjectsContainer.tsx`, which both branches change. So this is a one-line addition after you merge `develop`, not a conflict to untangle.

(Edited: my first version of this note had the merge order the wrong way round.)

PR #3100, opened 2026-09-01
fix: return profile fetch errors to the caller instead of store state

### LINKED ISSUE #3045: A2 — Return fetch errors to the caller instead of through shared state
Part of #3030 (follow-up epic) · tracked as **A2**.

**Phase 2.** **Linked items:** I7 (unblocks the log fix), I3 (shared error handling).

**Scope**
Today `fetchUserProfile` writes failures into a shared `profileApiError` state, and a separate
effect reads that and handles it. This "error through state" indirection caused two of the bugs
the migration had (an error being silently dropped, and stale error state re-firing).

**The reshape**
A fetch's failure should return to whoever triggered it. The store should hold committed data
only, not a transient error channel.

**When:** Phase 2, after A1. Its own PR (it changes the exact surface the earlier PRs read from).

**For the tracker** · Area: user-store · Labels: `architecture`, `refactor`, `priority: medium` · Depends on: A1
**Done when:** a fetch failure returns to its caller, the store holds committed data only, and the error-through-state effect is removed.

sunilsabatp:
## Summary
- `fetchUserProfile` used to write failures to a `profileApiError` field in the store, which `useInitializeUser` picked up in a separate effect to call `handleProfileError`. That added an extra render cycle between the failure happening and it being handled, and split the error-handling logic across two files.
- `fetchUserProfile` now rejects with a normalized `APIError` instead. Callers (`useInitializeUser`, `ImpersonateUserForm`, `ImpersonationActivated`, and `useProfileErrorHandler`'s own impersonation-restore retry) catch the rejection and call `handleProfileError` directly, except for the impersonation-start 403 case, which `ImpersonateUserForm` still handles locally as before.
- Removed `profileApiError` and `clearProfileApiError` from `userStore`, since nothing reads or clears that state anymore.
- Added `src/stores/userStore.test.ts` covering `fetchUserProfile`: success, 500 (profile kept), 401/303 (profile cleared), a raw network failure normalized into an `APIError`, and the impersonation 403 path (profile untouched, rejects for the caller to handle).

## Test plan
- [x] `src/stores/userStore.test.ts` added and passing
- [ ] Manually verify profile fetch failure still surfaces the global error handling (e.g. session expiry redirect) in the browser
- [ ] Manually verify starting impersonation with an invalid email/support pin still shows the inline error instead of the global handler

### DIFF
--- a/src/features/user/Settings/ImpersonateUser/ImpersonateUserForm.tsx
+++ b/src/features/user/Settings/ImpersonateUser/ImpersonateUserForm.tsx
@@ -11,6 +11,7 @@ import styles from './ImpersonateUser.module.scss';
 import { isEmailValid } from '../../../../utils/isEmailValid';
 import { APIError } from '@planet-sdk/common';
 import useLocalizedPath from '../../../../hooks/useLocalizedPath';
+import useProfileErrorHandler from '../../../../hooks/useProfileErrorHandler';
 import { useAuthStore, useTenantStore, useUserStore } from '../../../../stores';
 
 const ImpersonateUserForm = (): ReactElement => {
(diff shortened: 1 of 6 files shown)


mariahosfeld (commented):

PR #3101, opened 2026-09-01
fix: surface the real profile fetch error message instead of a generic one

### LINKED ISSUE #3037: I7 — Profile error logs print "Something went wrong" instead of the real cause

sunilsabatp:

- `fetchUserProfile` used to write errors to `profileApiError` in the store, and a separate effect in `useInitializeUser` picked that up to call `handleProfileError`. That added an indirection and a render cycle between the failure and its handling. Callers now catch the rejected promise and call `handleProfileError` directly, so `profileApiError` and `clearProfileApiError` are gone.
- `APIError.message` is always the generic string `"Something went wrong"` - the real cause lives in `.errors`. `useProfileErrorHandler` now reads the message off `err.errors.message` via a `getProfileErrorMessage` helper, falling back to `err.message` for errors that don't follow the `{ code, message }` shape. `userStore` now sets `errors.code` on every thrown `APIError` so that shape holds consistently.
- Added test coverage for `userStore.fetchUserProfile`: success, 401/303/500 handling, network failure normalized into an `APIError`, and the 403 impersonation path.

### DIFF
--- a/src/hooks/useProfileErrorHandler.ts
+++ b/src/hooks/useProfileErrorHandler.ts
@@ -11,6 +11,15 @@ import {
   registerRedirectAttempt,
 } from '../utils/authRedirectGuard';
 
+// APIError.message is always the string "Something went wrong" - the real
+// text lives in `.errors`, which fetchUserProfile always shapes as
+// { code, message }. Fall back to err.message for errors from elsewhere
+// that don't follow that shape.
+const getProfileErrorMessage = (err: APIError): string =>
+  'message' in err.errors && typeof err.errors.message === 'string'
+    ? err.errors.message
+    : err.message;
+
 const useProfileErrorHandler = () => {
   const router = useRouter();
   const locale = useLocale();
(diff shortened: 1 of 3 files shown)


mariahosfeld (commented):
Functionality looks good.

PR #3103, opened 2026-09-02
fix: date labels update immediately after a locale switch

### LINKED ISSUE #3020: Follow-up: remove remaining localStorage.language readers
### Why
PR #3018 restores the missing `localStorage.language` writer and fixes the original regression from #3015.

There is still follow-up work left: some date formatting paths still read `localStorage.language` directly. That can leave parts of the UI stale after a locale switch, especially in payment history / recurring payment views.

### Scope
- Replace active UI reads of `localStorage.language` with route locale / `useLocale()` where available.
- Update date formatting so locale changes are reflected immediately.
- Keep `localStorage.language` only for any remaining legacy compatibility until the last reader is removed.

### Suggested approach
- Update date formatting utilities to accept locale explicitly, or derive it at the call site.
- Audit remaining getItem('language') readers and replace them where practical.
- Verify payments/history screens update immediately when the language changes.

### Acceptance criteria
- Payment-history date labels update immediately after a locale switch.
- No active UI path depends on `localStorage.language` for current locale rendering.
- Any remaining readers are either removed or clearly marked compatibility-only.

sunilsabatp:
## Summary

Follow-up to #3018 / #3015, addresses #3020.

Date-formatting paths still read `localStorage.language` directly instead of using the route's current locale, so payment history, recurring donation, and project date labels stayed in the old language after a locale switch (they only picked up the new language on a full page reload once `localStorage` caught up).

- `formatDate()` (`getFormattedDate.ts`) now takes `locale` as an explicit parameter instead of reading `localStorage` internally, matching the pattern already used by `getFormattedCurrency()`. All 11 call sites now pass `locale` from `useLocale()`.
- Four components that built `date-fns` locale objects directly from `localeMapForDate[localStorage.getItem('language')]` (project reviews, reports, contribution/registration summaries) now use `useLocale()` instead.
- `EditModal.tsx` (recurring donation edit) and `ManageProjects/index.tsx` each held a `userLang` state populated once from `localStorage` on mount, even though `locale` from `useLocale()` was already available in the same component. Removed the dead state/effect in favor of the existing reactive value.
- The remaining `localStorage.language` reads (Auth0 `ui_locales` login/signup redirects, donation widget URL) are out of scope here since they redirect to external services rather than render in-app UI. They're now marked with a compatibility comment per the issue's acceptance criteria, instead of being silently left as-is.

## Why this approach

The codebase already has an established pattern for this: `getFormattedCurrency(locale, ...)` takes locale explicitly and callers pass `useLocale()`. Matching that for dates keeps the fix small and idiomatic, and makes the locale dependency explicit at each call site rather than hiding a `localStorage` read (and its staleness) inside the formatter.

## Test plan

- [x] `tsc --noEmit` shows the same pre-existing error count before and after (no new type errors)
- [x] `eslint` clean on all changed files
- [ ] Manually switch locale on Payment History / Recurring Donations screens and confirm dates update immediately without a page reload

### DIFF
--- a/pages/sites/[slug]/[locale]/claim/[type]/[code].tsx
+++ b/pages/sites/[slug]/[locale]/claim/[type]/[code].tsx
@@ -142,6 +142,7 @@ function ClaimDonation(): ReactElement {
         localStorage.setItem('redirectLink', router.asPath);
         loginWithRedirect({
           redirectUri: `${window.location.origin}/login`,
+          // Legacy compatibility read, kept until Auth0 redirects are migrated off localStorage.language (see issue #3020).
           ui_locales: localStorage.getItem('language') || 'en',
         });
       }
--- a/pages/sites/[slug]/[locale]/login.tsx
+++ b/pages/sites/[slug]/[locale]/login.tsx
@@ -75,6 +75,7 @@ export default function Login(): ReactElement {
     // Not authenticated → login
     loginWithRedirect({
       redirectUri: `${window.location.origin}/login`,
+      // Legacy compatibility read, kept until Auth0 redirects are migrated off localStorage.language (see issue #3020).
       ui_locales: localStorage.getItem('language') || 'en',
     });
   }, [userProfile, isAuthResolved, isInitialized]);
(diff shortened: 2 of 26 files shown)


mariahosfeld (commented):

Issue #3010, opened 2026-07-29
Embedded project map shows no data when project_details=false


mohitb35:
**Describe the bug**
An embedded project map with the details pane hidden renders with no data. No interventions appear, and no request is made for either the project or its interventions.

**Is it on a Production or Development Build?**
Production. Introduced by #2857, merged 2026-07-22.

**To Reproduce**
1. Open `https://web.plant-for-the-planet.org/en/proj_WZkyugryh35sMmZMmXCwq7YY?embed=true&project_details=false&locale=en`
2. Open the network tab and filter on `projects`
3. See that no project or intervention request is made

**Expected behavior**
The map draws the project's interventions. Hiding the details pane should hide UI, not stop data loading.

**What actually happens**
An empty map. Everything the project fetch triggers is missing, not just the two requests:
- interventions
- the time travel config, so the time travel tab is silently dead
- the error redirect, so a bad slug leaves an empty map with no signal
- the `backNavigationUrl` session write

**Desktop**
- OS: latest macOS
- Browser: latest Chrome

**Smartphone**
- Also affected. On mobile the layout forces map-only mode when details are hidden, so the component never mounts there either.

**Additional context**
The project fetch lives inside `ProjectDetails`, the component that `project_details=false` is designed to hide, and it is the only API call site in the details subtree.

This configuration previously worked by accident. `page` used to be a synchronous prop, so for exactly one render the layout gate was true while `embed` was still undefined. `ProjectDetails` mounted, its effect fired, and the data outlived the unmount. #2857 moved `page` into a store where it starts as `null`, which closed that window and took the fetch with it.

The project list embed, `?embed=true&project_list=false`, is not affected, since that fetch already sits in an initializer above the layout gate.

PR #3108, opened 2026-09-04
Derive current page synchronously

### LINKED ISSUE #3012: Derive the current page synchronously so it has no null window
Follow-up to #3010, which was caused by this pattern. That PR moved the fetch out of the gated component, which fixes the symptom. This issue removes the cause.

## Problem

`viewStore.page` starts as `null` and is only set from an effect after `router.isReady`. Every consumer therefore has a first render where the current page is unknown. Where that value feeds a render gate, the consequence is permanent: a component that never mounts never fires its effect. That is exactly how #3010 happened.

## Proposal

Derive the page from `router.pathname`, which is known synchronously on both server and client, so no `router.isReady` and no query access is needed and the `null` window disappears rather than shrinking.

```
'/sites/[slug]/[locale]'      -> 'project-list'
'/sites/[slug]/[locale]/[p]'  -> 'project-details'
otherwise                     -> null
```

## Scope

About 15 files, one new, roughly +45/-55, every edit mechanical.

- New `useCurrentPage` hook
- `viewStore` drops `page` and `setPage`, keeps `Page`, `EMBEDDABLE_PAGES`, `isEmbeddablePage` and `selectedMode`
- `useInitializeView` loses its derivation effect
- 12 read sites swap the selector for the hook: `ProjectsMap`, `ProjectSnippet` with `TpoName`, `ProjectBadge` and `ImageSection`, `ProjectListControlForMobile`, `Layout/index`, `MobileProjectsLayout`, `ProjectsLayout/index`, and the project, single project and intervention initializers

Leaving the `Page` type in `viewStore` keeps `MapControls` untouched.

## Do this first: the SSR surface grows

Today `page` is `null` on the server, so `ProjectsLayout` renders no children and the details subtree never server-renders on project routes. Once `page` is correct on the server, that whole subtree starts being server-rendered, so anything in it touching `window`, `sessionStorage` or `localStorage` during render rather than in an effect would begin throwing. `ImageSection` guards its `ontouchstart` check, but the subtree has not been audited.

Audit that before writing any code.

## Blocked, do not attempt in the same change

Deleting the `mounted` workaround in `ImageSection`. With `page` synchronous but `isEmbedMode` and `showBackIcon` still async, `showBackButton` becomes true on first render, so an embed with `back_icon=false` would render the back arrow, including in the server HTML, and then remove it. That is worse than today, and it is exactly what `mounted` suppresses.

It needs `queryParamStore` to derive synchronously first, which means undoing the `isContextLoaded` latch. The embed Navbar and `ViewModeTabs` flashes come from that same async `embed`, so they are not fixed by the page change either. Those flashes date from #2820, not #2857.

## Smaller consequences

- The `currentPage` effect deps in `ProjectsMap` and the three initializers stop firing an extra time on the null-to-value flip. Same end state, fewer runs
- The comment in `useInitializeProject` about `currentPage` starting `null` becomes obsolete

sunilsabatp:
## Summary

`viewStore`'s `page` field was only set once the Next.js router became ready, via an effect in `useInitializeView`. This left a `null` window on first render during which any page-gated logic could permanently miss its mount (issue #3010).

`useCurrentPage` derives the page directly from `router.pathname`, which is known synchronously on both server and client, so there is no ready-wait and no `null` window.

- `useCurrentPage.ts` (new) replaces `viewStore`'s `page` state and `setPage` action
- All consumers (`Layout`, `ProjectsLayout`, `MobileProjectsLayout`, the `useInitialize*` hooks, `ProjectsMap`, `ProjectListControlForMobile`, `ProjectSnippet` and its microComponents) switch from `useViewStore((state) => state.page)` to `useCurrentPage()`
- Fixes a related bug in `singleProjectStore.updateSingleProjectUrl`: it was pushing the tenant-facing display path as the router `pathname`, which corrupted `router.pathname` after the shallow push (no longer matching `/sites/[slug]/[locale]/[p]`), breaking page detection

### First render / hard reload (the main bug this PR fixes)

- [x] Hard reload (full browser refresh, not a client-side link) directly on a project list page (`/sites/<slug>/<locale>`) and confirm list-specific UI (filters, list/map toggle, embed handling) is correct from the very first paint, with no flash of the wrong state
- [x] Hard reload directly on a project details page (`/sites/<slug>/<locale>/<p>`) and confirm details-specific UI (project snippet, badge, TPO name, back button) is correct from the very first paint
- [x] Hard reload directly on a project details URL that also has `site` and `intervention` query params, and confirm the correct site/intervention render immediately, not just after a delay

### Embed mode edge cases

- [x] Load a project list page with `?embed=true` and confirm embed mode activates immediately (no flash of the non-embed layout)
- [x] Load a project details page with `?embed=true` and confirm the same
- [x] Load a non-embeddable page (e.g. a profile or claim page under `/sites/<slug>/<locale>/...`) with `?embed=true` and confirm embed mode does NOT activate, since `useCurrentPage` returns `null` there

### Mobile layout

- [x] On a mobile viewport, load the project list page directly and confirm the list/map view mode initializes correctly
- [x] On a mobile viewport, load the project details page directly and confirm the same
- [x] Toggle between list/map mode on mobile and confirm it still works after navigating between list and details

### Client-side navigation

- [x] From the project list, click into a project (client-side transition, not a hard reload) and confirm the page switches to details correctly
- [x] Navigate back from project details to the list and confirm the page switches back correctly
- [x] Navigate to a page outside the project routes (e.g. home) and confirm project-page-gated UI (embed mode, map, filters) fully turns off

### `updateSingleProjectUrl` regression check (the `singleProjectStore` fix)

- [x] On a project details page, select a different site and confirm the URL updates to show the site, the map/snippet reflect the new site, and the page is still detected as `project-details` afterward (no layout reset or flicker)
- [x] Do the same for selecting an intervention
- [x] After selecting a site/intervention, hard reload the resulting URL and confirm it still resolves to the project details page correctly (this is the scenario the old code silently corrupted `router.pathname` on)

### Project snippet components

- [x] Confirm `ImageSection`'s back button only shows in the contexts it should (embed vs non-embed, list vs details)
- [x] Confirm `ProjectBadge` renders the correct badge/tooltip on both list and details
- [x] Confirm `TpoName` links/styling are correct on both list and details

### DIFF
--- a/src/features/common/Layout/ProjectsLayout/MobileProjectsLayout.tsx
+++ b/src/features/common/Layout/ProjectsLayout/MobileProjectsLayout.tsx
@@ -10,6 +10,7 @@ import {
   useUserStore,
   useViewStore,
 } from '../../../../stores';
+import { useCurrentPage } from '../../../../hooks/useCurrentPage';
 
 export type ViewMode = 'list' | 'map';
 interface ProjectsLayoutProps {
@@ -28,7 +29,7 @@ const MobileProjectsLayout = ({ children, isMobile }: ProjectsLayoutProps) => {
   const isImpersonationModeOn = useUserStore(
     (state) => state.isImpersonationModeOn
   );
-  const currentPage = useViewStore((state) => state.page);
+  const currentPage = useCurrentPage();
   const isMapMode = useViewStore((state) => state.selectedMode === 'map');
 
   // store: action
--- a/src/features/common/Layout/ProjectsLayout/index.tsx
+++ b/src/features/common/Layout/ProjectsLayout/index.tsx
@@ -9,8 +9,8 @@ import {
   useProjectMapStore,
   useQueryParamStore,
   useUserStore,
-  useViewStore,
 } from '../../../../stores';
+import { useCurrentPage } from '../../../../hooks/useCurrentPage';
 
 interface ProjectsLayoutProps {
   children: ReactNode;
@@ -27,7 +27,7 @@ const ProjectsLayoutContent = ({ children }: ProjectsLayoutProps) => {
   const isImpersonationModeOn = useUserStore(
     (state) => state.isImpersonationModeOn
   );
-  const currentPage = useViewStore((state) => state.page);
+  const currentPage = useCurrentPage();
   // store: action
   const updateMapOption = useProjectMapStore((state) => state.updateMapOption);
 
--- a/src/features/common/Layout/index.tsx
+++ b/src/features/common/Layout/index.tsx
@@ -12,9 +12,9 @@ import {
   useAuthStore,
   useQueryParamStore,
   useTenantStore,
-  useViewStore,
 } from '../../../stores';
 import { isEmbeddablePage } from '../../../stores/viewStore';
+import { useCurrentPage } from '../../../hooks/useCurrentPage';
 
 const Layout = ({ children }: { children: ReactNode }) => {
   const { theme: themeType } = useTheme();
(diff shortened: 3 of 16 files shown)


mariahosfeld (commented):
Functionality looks good to me. All tests worked as expected.

PR #3109, opened 2026-09-04
fix: coerce null/undefined translation values for next-intl v4

### LINKED ISSUE #2967: next-intl v4: fix translation values passing null/undefined/boolean (no longer valid types)
## Summary

The next-intl v4 migration (PR #2955) tightened the type of translation interpolation values, and this surfaced pre-existing type errors where `null` / `undefined` / `boolean` were passed as translation values. These previously typechecked under v3 and now fail under v4.

## Root cause

next-intl v4 narrowed the accepted `TranslationValues` type:

**v3.26.5** (`use-intl@3.26.5`):
```ts
export type TranslationValue = string | number | boolean | Date | null | undefined;
```

**v4.13.0** (current):
```ts
export type TranslationValues = Record<string, string | number | Date>;
```

`null`, `undefined`, and `boolean` are no longer valid interpolation values — only `string | number | Date`.

## Known occurrence

`src/features/projectsV2/ProjectsMap/ProjectSiteDropDown/index.tsx`

```tsx
{tProjectDetails('siteCount', {
  siteId: getId(selectedSiteData?.id),   // getId returns `number | null`
  totalCount: siteList.length,
})}
```

`getId` returns `number | null`, producing:

```
Type 'number | null' is not assignable to type 'string | number | Date'.
  Type 'null' is not assignable to type 'string | number | Date'.
```

## Scope

There are likely **other spots** passing `null` / `undefined` / `boolean` into `t(...)` calls that were valid under v3. This issue should include a sweep across the codebase to find and fix all of them, not just the one above.

## Suggested fix

For each occurrence, coerce the value to an accepted type or guard the render. For the known case, decide the desired behavior when no site matches (fallback number vs. not rendering the count), e.g.:

```tsx
siteId: getId(selectedSiteData?.id) ?? 0,
```

## Context

- Introduced by: PR #2955 (next-intl v3 → v4.13.0)
- The affected code itself was authored earlier (commit f705ec7b); v4's stricter typing exposed it.

sunilsabatp:
## Summary

next-intl v4 (PR #2955) narrowed the `TranslationValues` type from `string | number | boolean | Date | null | undefined` down to `string | number | Date`. Several call sites across the codebase were passing `null` or `undefined` interpolation values that typechecked fine under v3 but now fail under v4.

This PR sweeps the codebase for every such occurrence and coerces each value at the call site (`?? 0` / `?? ''`), matching the fallback the issue itself suggested for the known case.

## Changes

Found 10 occurrences across 8 files via `tsc --noEmit`:

- `ProjectSiteDropDown/index.tsx` - `getId(...)` returns `number | null`
- `MultiTreeInfoHeader.tsx` - `area` was `string | undefined`; the message's ICU `select` matches the literal string `"undefined"` to hide the area chunk, so the fallback is now the string `'undefined'` instead of the JS value
- `PlantInfoCard.tsx` - `tag` was `string | null | undefined`
- `PlantingDetails.tsx` - `plantingDensity` was `number | null` (also fixes a latent bug where a density of `0` rendered as `null`)
- `MultiTreeInfo.tsx` - `item.tag` was `string | null`
- `ProjectListTabForMobile.tsx` (x2) - `topProjectCount` / `projectCount` were `number | undefined`
- `DonorDetails.tsx` (x2) - `donor.type` was `'individual' | 'organization' | null`
- `DeleteProfileForm.tsx` - `userEmail` was `string | undefined`

## Why this approach

Coercing at the call site is the smallest possible diff and keeps runtime behavior identical (verified against each translation message - plain interpolation or ICU `select` with an `other` fallback). Widening next-intl's types back would undo the point of the v4 upgrade, and editing the translation JSON files (en/de) for the ICU-select case would require larger, riskier changes for no behavioral gain.

## Test plan

- [x] `tsc --noEmit` before/after diffed: exactly the 10 target errors are gone, no new errors introduced (116 pre-existing, unrelated errors remain)
- [ ] Manual verification of the affected UI (project sites, tree counts, planting details, donation receipt donor info, delete account) in a running app

Fixes #2967

### DIFF
--- a/src/features/projectsV2/ProjectDetails/components/MultiTreeInfo.tsx
+++ b/src/features/projectsV2/ProjectDetails/components/MultiTreeInfo.tsx
@@ -42,7 +42,7 @@ const MultiTreeInfo = ({ activeMultiTree, isMobile }: Props) => {
       return {
         id: item.coordinates[0].id,
         image: item.coordinates[0].image ?? '',
-        description: tProjectDetails('sampleTreeTag', { tag: item.tag }),
+        description: tProjectDetails('sampleTreeTag', { tag: item.tag ?? '' }),
       };
     });
     return result;
--- a/src/features/projectsV2/ProjectDetails/components/microComponents/MultiTreeInfoHeader.tsx
+++ b/src/features/projectsV2/ProjectDetails/components/microComponents/MultiTreeInfoHeader.tsx
@@ -32,7 +32,8 @@ const MultiTreeInfoHeader = ({
             Number(totalTreesCount),
             1
           ),
-          area: hectaresCovered > 0 ? hectaresCovered.toFixed(3) : undefined,
+          // The message's ICU select matches the literal string 'undefined' to hide the area chunk
+          area: hectaresCovered > 0 ? hectaresCovered.toFixed(3) : 'undefined',
           areaContainer: (chunks) => <span>{chunks}</span>,
         })}
       </div>
--- a/src/features/projectsV2/ProjectDetails/components/microComponents/PlantInfoCard.tsx
+++ b/src/features/projectsV2/ProjectDetails/components/microComponents/PlantInfoCard.tsx
@@ -34,7 +34,7 @@ const PlantInfoCard = ({
     {
       label: tProjectDetails('treeTag'),
       data: tProjectDetails('tag', {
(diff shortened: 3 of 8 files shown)


mariahosfeld (commented):

PR #3115, opened 2026-09-08
fix: update project classification values for SDK 0.3.0

mohitb35:
## What

Bumps `@planet-sdk/common` to 0.3.0 and updates the project classification values it renames.

- `large-scale-planting` → `restoration-tree-planting`
- `other-planting` → `other-restoration`

## Why

The API already serves only the new values. I checked both environments on `_scope=map`: 325 tree projects on staging and 332 on production, with zero old values on either.

This caused two visible bugs on develop:

- Project cards showed the raw key `Project.classification.restoration-tree-planting` instead of a label.
- The "Large Scale Restoration" and "Other Planting" filters returned no results, because `filterByClassification` does an exact string match against a filter list still holding the old spellings.
- About 47% of tree projects were missing from the map. `getPointMarkerImageKey` returns `''` for a classification that is not in `POINT_MARKER_MAP`, and `ProjectMarkersGL` drops those projects from the `FeatureCollection` before it reaches maplibre. So they were not just missing a pin, they were absent from the layer and could not be hovered or clicked. On production that was 157 of 332 tree projects: 136 `restoration-tree-planting` plus 21 `other-restoration`.

## Changes

- `package.json` — SDK to `^0.3.0`
- `projectV2.ts` — new values in `availableFilters`
- Icon maps and switch cases in `ClassificationDropDown`, `markerImageRegistry`, `ClusterIcon`, `ProjectTypeIcon`, `common/ProjectTypeIcon`
- Two storybook fixtures
- `en/allProjects.json` — renamed keys
- `en/project.json` — new keys, with the old ones kept as aliases in case a draft project still holds an old value

Only `en` locale files are touched. The rest sync through Lingohub.

## Testing

- `tsc --noEmit` reports 128 errors, identical to the develop baseline. The bump introduced 12 errors and all 12 are fixed here.
- ESLint clean on all changed files.
- `vitest run src/utils` — 45 pass.
- Verified in the browser: labels, filters, and map markers all render.

## Notes

The SDK 0.3.0 bump also reshapes donation types and drops the deprecated `treeCount` and `quantity` fields. This produced no errors, because the donation code uses local types in `src/features/common/types/` rather than the SDK's.

`feature/project-onboarding` already carries this rename as part of a much larger change. This PR takes only the classification part so develop is fixed without waiting for that branch.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,7 @@
         "@mui/x-date-pickers": "^5.0.15",
         "@netlify/plugin-nextjs": "^4.33.0",
         "@next/bundle-analyzer": "^10.2.3",
-        "@planet-sdk/common": "^0.2.4",
+        "@planet-sdk/common": "^0.3.0",
         "@sentry/browser": "^6.15.0",
         "@sentry/integrations": "^6.19.2",
         "@sentry/node": "^6.19.2",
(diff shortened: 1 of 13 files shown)


mohitb35:
Filed #3116 for the wider theme: every one of these three symptoms failed silently, and the same enum-fallthrough pattern exists in four other places. Out of scope for this PR, which stays a straight rename.

mariahosfeld (approved):

mariahosfeld (commented):
LGTM!

PR #3129, opened 2026-09-11
feat(onboarding): say which file formats each upload takes

### LINKED ISSUE #3128: Onboarding upload fields do not say which formats they take, so an RO cannot pick a photo of a certificate
An RO whose only copy of a bylaw or a legal accreditation is on paper photographs or scans it. The onboarding upload field gives no sign that an image is acceptable, and a file picker with no `accept` attribute offers every file on the machine with no hint which of them will be taken. The RO either does not try, or tries and gets a rejected upload with no explanation.

The backend accepts those images already. Nothing in the UI says so.

## Where this comes from

Max, in [#feat-ro-onboarding on 2 September](https://planetfoundation.slack.com/archives/C0BSS500W7K/p1788333924145929):

> Customize in the same way as in project onboarding - images allowed, **state clearly in the UI**

The first half is done: `Plant-for-the-Planet-org/treecounter-platform#779` widened the organisation documents from PDF-only to pdf/png/jpeg, and project onboarding has accepted images for longer than that. The second half is this issue. It was tracked as the last open box on `Plant-for-the-Planet-org/treecounter-platform#778`, which assumed it lived here — correctly, but the RO-facing screens had no version of it at all, so there was nothing to amend.

## The API already carries the answer

No backend change is needed and none should be requested. Every row of both document checklists returns the accepted types, and, once `Plant-for-the-Planet-org/treecounter-platform#794` merges, a ready-made human sentence beside them:

```json
{
  "kind": "ro_bylaws",
  "label": "Organization Bylaws",
  "required": true,
  "acceptedMimeTypes": ["application/pdf", "image/png", "image/jpeg"],
  "acceptedFormats": "PDF, PNG or JPG",
  "maxByteSize": 26214400,
  "fulfilled": false,
  "current": null
}
```

- `GET /profile/dueDiligence/documents` — the six organisation documents, under `documents`
- `GET /projects/{guid}/documents` — the project checklist

## What to do

- [ ] Set `accept` on each file input from that row's `acceptedMimeTypes`, joined with commas. It is per document, not per screen.
- [ ] Show `acceptedFormats` next to the field, before a file is chosen.
- [ ] Do the same on both checklists, organisation and project.

**Render the values, do not hardcode them.** The two checklists genuinely differ — the project budget plan and species checklist also take `.xlsx`, and no organisation document does:

```json
"acceptedMimeTypes": ["application/pdf", "image/png", "image/jpeg",
                      "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"],
"acceptedFormats": "PDF, PNG, JPG or XLSX"
```

A hardcoded list would be wrong for one of the two on the day it is written, and wrong for both the next time the policy moves. That is exactly how the organisation documents came to be PDF-only for a week after the decision to widen them.

## Worth knowing

`accept` filters the picker, it does not enforce anything. The server-side check is the real gate and is unchanged, so a file that slips past the picker is still rejected — which is why the visible sentence matters as much as the attribute.

Both `maxByteSize` (25 MB) and `required` are on the same row if the field is worth annotating further. A photograph off a modern phone is nowhere near the limit, so size is not urgent, but an oversized file is the other silent rejection on this screen.

The same wiring landed on the two backend admin screens in `Plant-for-the-Planet-org/treecounter-platform#794` if a reference implementation is useful.

Refs `Plant-for-the-Planet-org/treecounter-platform#778`

jmiridis:
Closes #3128

An RO whose only copy of a bylaw or an accreditation is on paper photographs it. Neither document checklist gave any sign that an image would be taken, and the organisation picker went further and filtered to PDF only — which the backend stopped requiring in `Plant-for-the-Planet-org/treecounter-platform#779`. The RO either did not try, or tried and got a silent rejection.

Both checklists already return the answer on every row, so this renders those values instead of restating the policy in the frontend.

## What changed

**Organisation checklist** (`DueDiligence/components/DocumentRow.tsx`)
- `accept` and `maxSize` now come from the row's `acceptedMimeTypes` and `maxByteSize`, as the project checklist already did.
- PDF and 25 MB remain only as fallbacks for a backend that has not shipped those fields yet. The comment says not to widen them by hand.

**Both checklists**
- The row's own `acceptedFormats` sentence renders under the label, so it is there before a file is picked. It also stays visible when replacing a file, since that is the same choice again.
- A file the picker drops for its type now names the formats that row does take, via a new `documentWrongTypeWithFormats` key. The old generic message stays as the fallback when the backend has not sent the sentence.
- Organisation `documentsHint` no longer says "Please upload a PDF for each item", and `documentWrongType` no longer says "Please upload a PDF."

**Types** — `acceptedFormats?: string` added to `DueDiligenceDocument` (along with `acceptedMimeTypes` and `maxByteSize`) and to `DocumentChecklistItem`. All optional, so a deploy against an older backend degrades to the previous behaviour rather than breaking.

## Nothing is hardcoded

The two checklists genuinely differ — the project budget plan and species checklist also take `.xlsx`, no organisation document does — so every format string and MIME list on screen comes from the row it belongs to.

## Worth knowing

`accept` only filters the picker; the server check is unchanged and is still the real gate. That is why the visible sentence matters as much as the attribute.

No backend change is needed. The `acceptedFormats` field arrives with `Plant-for-the-Planet-org/treecounter-platform#794`; until it merges the sentence simply does not render, and `accept` already works off `acceptedMimeTypes`.

## Checks

`tsc --noEmit` (no new errors in the touched files), `eslint`, `prettier`, and the DueDiligence / ManageProjects vitest suites all pass.

Base is `feature/project-onboarding`, where both screens live.

Refs `Plant-for-the-Planet-org/treecounter-platform#778`


mohitb35 (approved):
LGTM

PR #3135, opened 2026-09-16
fix(profile): hide Donate button in map popup for projects that do not allow donations

mohitb35:
## Problem
The contributions map popup showed the Donate / Donate Again button even for projects with `allowDonations: false`.
The contributions list view gates the same button correctly, so the same project behaved differently in the two places.

Opening the project on donate.plant-for-the-planet.org (planet-donations) confirms it: the project is not donatable there.

## Cause
`ProjectInfoSection` did not read `allowDonations` off `projectInfo`.
The field is present on the object, typed `boolean` on `MyForestProject`.

## Fix
Gate the button on `allowDonations === true`, the same check `ProjectItemCard` already uses.

## Note
`DonateButton` itself still has no guard, so any future caller can leak the button again.
Moving the check inside the button is a separate cleanup, not done here.


mariahosfeld (approved):
LGTM!

PR #3139, opened 2026-09-18
Upgrade Next.js to 15.5.25 (Phase 2)

sunilsabatp:
## Summary

Upgrades Next.js from `14.2.35` to `15.5.25`, the latest stable 15.x version before Next.js 16, while keeping the Pages Router unchanged.

This is Phase 2 of the `staged Next.js 16 upgrade plan`

### Changes

* Upgraded `next` to `^15.5.25`
* Moved `experimental.serverComponentsExternalPackages` to the stable top-level `serverExternalPackages`
* Removed the forced `.babelrc` using `next/babel`
* Removed unused `babel-loader` and `@babel/plugin-transform-unicode-regex`
* Switched the app to Next.js's default SWC compiler

This also resolves Phase 1.8 of the upgrade plan: **Babel removed, using SWC**.

## Why Babel was removed

With `.babelrc` enabled, `npm run build` failed on Next.js 15 during `Collecting page data` with:

```text
jsxDEV is not a function
```

The issue came from forcing `next/babel` through the Next.js 15 build pipeline.

Removing `.babelrc` and allowing Next.js to use SWC fixed the build without any other code changes.

Before removing Babel, we confirmed:

* `@emotion/babel-plugin` was never installed, so Emotion did not depend on Babel
* There is no `browserslist` config requiring the unicode-regex transform
* Supported browsers already handle the unicode-regex behavior natively

## Out of scope

This PR does not change:

* Cypress
* CI typecheck baseline job
* Storybook
* ESLint modernization
* Heroku/Express custom-server decision
* App Router work
* `next/navigation`
* Metadata API migration

These are tracked separately and do not block the Next.js 15 upgrade.

## Test plan

* [x] `npm run build` — successful, all 45 pages generated
* [x] `npx vitest run` — 68/68 tests passed
* [x] `npm run lint:errors` — no errors
* [ ] Manual smoke test in a deployed preview for tenant routing, locale handling, auth, and donation flow


mariahosfeld (dismissed):
Not able to replicate on production, so please check: Sometimes, after viewing the project details, clicking on home does not center the whole map again, but the view is rather "stuck" somewhere on the landmass. Only a second click on home resolves this. Does not happen all the time.
Another instance: zooming out to just show the ocean, no landmass.

mohitb35 (approved):
LGTM

PR #3140, opened 2026-09-18
fix(map): keep project pins above explore raster layers

mohitb35:
## Problem

Reported on the Platform: toggling on an explore layer such as Forest Cover draws it over the project pins, hiding some of them completely.

The project pins are no longer HTML markers. Commit 65e7281 ("perf(map): render explore markers as a GPU symbol layer") replaced the ~259 DOM `<Marker>` nodes with one GeoJSON source plus a symbol layer. HTML markers live in an overlay div above the canvas, so a raster layer could never cover them. Once the pins moved into the canvas, the style's layer order started to decide what covers what.

The explore layers are added with no `beforeId`:

```tsx
<Layer id={`${layerKey}-layer`} source={layerKey} type="raster" />
```

`addLayer` with no `before` argument appends the layer to the end of the draw order, so a layer toggled on lands above the pins that are already there. On first load the order happens to be fine, because `ExploreLayers` renders before `MultipleProjectsView`. The bug only appears when a layer is toggled on later, which is exactly what was reported.

## Fix

One effect in `ProjectMarkersGL`, which keeps the marker layer last in the draw order:

```tsx
useEffect(() => {
  const map = mapInstance?.getMap();
  if (!map) return;
  const keepMarkersOnTop = () => {
    const order = map.getLayersOrder();
    if (
      order.includes(PROJECT_MARKERS_LAYER) &&
      order[order.length - 1] !== PROJECT_MARKERS_LAYER
    ) {
      map.moveLayer(PROJECT_MARKERS_LAYER);
    }
  };
  keepMarkersOnTop();
  map.on('styledata', keepMarkersOnTop);
  return () => {
    map.off('styledata', keepMarkersOnTop);
  };
}, [mapInstance]);
```

`moveLayer` with no second argument moves a layer to the top. The guard makes the handler a no-op once the pins are already last.

### Why `styledata` and not a React dependency

Layer order lives in MapLibre, not in React. A layer can end up above the pins on a render that `ProjectMarkersGL` is not part of, for example when `exploreLayersData` arrives after the toggle. One `styledata` listener covers every path that changes the style. A dependency array such as `[mapInstance, mapOptions]` would only cover the plain toggle.

### Why not `beforeId`

`beforeId={PROJECT_MARKERS_LAYER}` on the explore layer looks shorter, but if the pins layer is not in the style, MapLibre refuses the `addLayer` call and only fires an error event, so the raster silently never renders. That happens whenever the Projects toggle is off, which would turn this bug into a worse one.

## Resulting layer order

`updateMapOption` allows at most one non-project explore layer at a time. In the project list view with Forest Cover on, the canvas holds (bottom to top):

1. base style layers (basemap fills, roads, place labels)
2. `forestCover-layer`
3. `project-markers-gl`

The popup, the navigation control and the side panels are DOM above the canvas and are unaffected. In project details view the pins layer does not exist and `ExploreLayers` does not render, so site, intervention and satellite ordering is untouched.

## Performance

- `getLayersOrder()` returns `[...style._order]`, a shallow copy of the layer id array. No style clone, unlike `getStyle()`.
- `styledata` is not one event per mutation. `addLayer` and `moveLayer` set `style._changed`, and `Style.update()` batches the queued changes once per render frame and fires a single `styledata` if anything changed.
- `addImage` does fire synchronously, so `registerMarkerIcons` triggers the handler about once per pin icon at first load. Each of those is a compare plus an early return.
- A toggle costs one `moveLayer`. That is an array splice on the layer order plus a repaint, not a source reload or a re-tessellation, and the pins keep their source, their icons and their event handlers.
- There is no per-frame work while panning or zooming. Nothing was added to the render path.
- Known trade-off: there is one frame, about 16ms, where the raster sits above the pins before the correction lands, because the handler runs on the next `styledata`. Not perceptible in testing.

## Testing

- Toggle Forest Cover on while the pins are drawn. Pins stay on top.
- Switch straight from Forest Cover to Deforestation or Fire Risk. Pins jump back above the new raster.
- Toggle Projects off with a layer on, then back on. The raster stays visible throughout, which is the case `beforeId` would have broken.
- Enter a project and return to the list. Site and intervention layers look unchanged.
- `eslint` clean on the changed file, and `tsc --noEmit` reports no errors in it. The repo's other `tsc` errors are pre-existing and unrelated.


sunilsabatp (approved):
LGTM

PR #3142, opened 2026-09-18
chore(deps): upgrade Storybook from 8 to 10.6.0

sunilsabatp:
## Summary

This PR upgrades Storybook from `8.6.18` to `10.6.0` so it remains compatible with the upcoming Next.js 16 upgrade.

This PR is stacked on `#3139` (`feature/nextjs-15-upgrade`), so that PR should be reviewed and merged first. After that, this PR can be rebased onto `develop`.

## Why this change

The current Storybook version does not support Next.js 16.

Since Chromatic also builds Storybook in CI, Storybook needs to be upgraded before we can safely move the application to Next.js 16.

Storybook was upgraded directly through the required migration path:

`8 → 9 → 10`

### Storybook

* Upgraded Storybook packages to `10.6.0`.
* Updated all 45 story files to the new Storybook import format.
* Removed Storybook packages that are no longer required in the newer version.
* Added `@storybook/addon-docs` separately, as required by Storybook 10.
* Removed outdated Storybook configuration.
* Excluded `storybook-static` from TypeScript checks because it is generated build output.

### Chromatic

* Upgraded the `chromatic` package from `^6.24.1` to `^18.9.4`, and `chromaui/action` from `v1` to `v18`. Chromatic stopped moving the floating `v1` tag in March 2025, where it points at CLI `11.27.0`. That version predates Storybook 9 and 10 and cannot read a Storybook 10 build.
* The workflow now passes `storybookBuildDir: storybook-static`, so the action uploads the build from the previous step instead of building Storybook a second time.
* The workflow now also passes `exitZeroOnChanges: true`, matching the `chromatic` script in `package.json`. Without it `chromaui/action` exits non-zero whenever snapshots differ from the baseline.

### Node version

* Chromatic workflow: Node 18 to Node 24.
* ESLint workflow: Node 22 to Node 24.
* `.nvmrc`: Node 16 to Node 24.

Every Node signal in the repo now reads 24, matching `engines.node` in `package.json`.

`.nvmrc` at 16 was the one that mattered in practice. `eslint-plugin-storybook@10` is an ESM package where `0.6.15` was CommonJS, so linting through the CommonJS `.eslintrc.js` needs Node's `require(ESM)` support (`^20.19 || >=22.12`). `chromatic@18` separately declares `engines.node >= 22`. Anyone running `nvm use` was landing on a Node that could do neither.

One correction to the earlier reasoning for Node 18: no Storybook package declares an `engines` field, so `npm ci` on Node 18 would have succeeded. The Storybook 10 CLI exits on its own below Node `20.19` or `22.12`, so the failure comes at build time, not at install time.

`.nvmrc` is not purely a developer signal, so it was worth checking. Netlify resolves its build Node version from `.nvmrc` when `NODE_VERSION` is unset, and our `netlify.toml` sets none. No deployment is affected: Netlify deployments for this repo were blocked a while back and no current version of the app is served from there, which matches the repo signals (the last 100 GitHub deployments are all Vercel or manual `planet-app-sf` ones, and no Netlify check appears on `develop`). Heroku is unaffected too, because its Node buildpack reads `engines.node` from `package.json`, which already said `24.x`.

Deleting the dead `netlify.toml` and `@netlify/plugin-nextjs` is tracked separately as item 1.4 of the Next.js 16 plan, and is left out of this PR.

## Not included

The following changes were intentionally left out because they are separate upgrades:

* Migration from Webpack to Vite.
* `@storybook/addon-mcp`.

## Verification

* Storybook production build passes.
* All 45 story files are still detected, giving 90 stories and 45 docs pages.
* Lint passes with 0 errors on Node 24.
* 68 tests pass.
* Next.js 15 production build still passes.
* No new TypeScript errors were introduced.

### Story rendering

A passing `build-storybook` proves the bundle compiles. It does not prove the decorator chain mounts. So the built Storybook was served and every story was opened in headless Chrome.

* 88 of 90 stories mount with real content and no Storybook error overlay.
* The manager UI boots. The sidebar tree, the toolbar and the Controls panel all render, and clicking a story in the sidebar navigates to it and renders it.
* Switching the locale toolbar to German renders German copy and German number formatting. This confirms the `NextIntlClientProvider` and MUI theme decorators work end to end.

### Known issues that this upgrade did not cause

Both were checked against Storybook 8 and are older than this PR.

* `Projects/Details/Popup` (`Open` and `Close`) throws `Cannot read properties of undefined`. The story passes only `isOpen`, but the component reads `feature.properties.eventDate` when it mounts, and the story has never passed `feature`. This upgrade only changed a type import in that file.
* CDN images return 404 in 8 stories because `process.env.CDN_URL` is undefined in Storybook. `@storybook/nextjs` does not apply the `env` block from `next.config.js`, in either `8.6.18` or `10.6.0`. Current Chromatic baselines already show these images as broken.

## Remaining check

The Chromatic workflow only runs on `develop`, so the first real Chromatic run still happens after merge. Expect that run to report a large number of snapshot differences: Storybook 8 to 10 changes the preview runtime, the docs pages and the addon set, and the current baseline was captured by CLI 11 against Storybook 8. Those differences are real and need accepting once in Chromatic to set a new baseline. `exitZeroOnChanges: true` is there so that does not leave `develop` red while it happens.

What could be checked ahead of time was checked. The `chromaui/action` `v18` tag exists, and `storybookBuildDir`, `token` and `projectToken` are all valid inputs at that tag. `chromatic@18.9.4` installs and runs on Node 24.


plans/nextjs-16-staged-upgrade-plan-pages-router.md
```diff
@@ -231,29 +231,48 @@ This materially reduces the custom webpack surface before the Next.js 16 build-s
...
 
-### Task
+### What changed
 
-Upgrade the Storybook stack to a Next.js 16-compatible 9.x line and verify:
+- `@storybook/nextjs`, `storybook`, and `eslint-plugin-storybook` are pinned at `10.6.0`. `@storybook/addon-docs` was added at `10.6.0` as well.
+- Four packages were removed because Storybook 9 folded them into the `storybook` core package: `@storybook/addon-interactions` and `@storybook/addon-actions` into the core addon set, `@storybook/test` into `storybook/test`, and `@storybook/theming` into `storybook/theming`.
+- `@storybook/addon-essentials` was removed and split. Its component addons are part of core now, and its docs half became the separate `@storybook/addon-docs` entry in `.storybook/main.js`.
+- `@storybook/react` was removed and replaced by `@storybook/nextjs`, which is where the story types live under the framework-based configuration.
+- `@storybook/addon-links` was removed outright, with no replacement. Nothing in the repo imported it.
```
mohitb35: addon-links has been removed entirely.

mohitb35 (approved):
LGTM

PR #3143, opened 2026-09-21
chore: remove Netlify config and the dead plugin dependency

sunilsabatp:
Netlify is no longer a deployment path. Deployments for this repo were blocked a while back and no current version of the app is served from there. Vercel is the release path, so the adapter and its config are dead weight.

- Deleted `netlify.toml` (build command, publish dir, three redirects, plugin entry).
- Removed `@netlify/plugin-nextjs` v4 from `package.json`.
- Regenerated `package-lock.json`.

## No routing changes

All three redirects in `netlify.toml` (`/my-trees`, `/redeem`, `/yucatan-reforestation`) already exist in the `redirects()` block in `next.config.js`, so deleting the file loses nothing.

## Checks run

- Lockfile diff is clean: 150 packages removed, 0 added, 0 version changes. Everything removed belongs to the plugin's own tree (`@netlify/*`, `ipx`, `unstorage`, `h3`, `ioredis`, `execa`, `tar-fs`).
- Top-level `sharp` 0.33.5 is untouched. Only the copy nested under `ipx` went.
- No source file imports any removed package.
- No CI workflow, `Procfile`, or `app.json` mentions Netlify.
- No user-facing docs mention Netlify either, so nothing in `README.md` or `docs/` goes stale.

## Note for the reviewer

The planning docs in `plans/` still say to upgrade the plugin to v5. That text is already rewritten on `feature/storybook-10-upgrade` in commit `71a400f9b`, which has not reached `develop` yet. I left `plans/` untouched here so the two branches do not conflict on the same lines.


mohitb35 (approved):
LGTM

Issue #3149, opened 2026-09-21
Enable eslint-plugin-react-hooks: 47 rules-of-hooks violations are currently unchecked


sunilsabatp:
`eslint-plugin-react-hooks` has been a devDependency for years, but it has never been registered as a plugin in our ESLint config. None of its rules have ever run. The ESLint flat config migration upgraded the package to v5.2.0 and kept that status quo, because switching rules on is new enforcement rather than a tooling migration.

This issue is to decide whether we turn it on, and to schedule the cleanup if we do.

## What it would report today

Measured on `pages` and `src` with `react-hooks/recommended` layered on top of our current `eslint.config.mjs`:

| Rule | Count | Files |
| --- | --- | --- |
| `react-hooks/rules-of-hooks` | 47 | 16 |
| `react-hooks/exhaustive-deps` | 158 | 95 |

### `rules-of-hooks` — 47 hits, the ones that matter

These are the likely real bugs. A hook that does not run in the same order on every render can read another hook's state, which shows up as stale or swapped values rather than a crash, so these are hard to spot in review.

Breakdown by kind:

- 41 conditional calls (hook called after an early return or inside an `if`)
- 4 hooks called somewhere that is neither a component nor a custom hook
- 2 other

By file:

```text
16  src/features/user/PlanetCash/components/TopUpManagement.tsx
 5  src/features/common/Layout/Navbar/microComponents/NavbarItemGroup.tsx
 5  src/features/common/Layout/Navbar/microComponents/NavbarMenuItem.tsx
 4  src/features/projectsV2/ProjectsMap/MapFeatureExplorer/microComponents/SingleLayerOption.tsx
 4  src/features/user/DonationReceipt/microComponents/DonorContactForm.tsx
 2  src/features/user/Profile/ContributionsMap/Markers/PointMarkers.tsx
 2  src/features/user/Settings/EditProfile/AddressManagement/microComponents/FormattedAddressBlock.tsx
 1  src/features/common/Layout/Navbar/microComponents/NavbarBrandLogos.tsx
 1  src/features/common/Layout/Navbar/microComponents/NavbarMenuSection.tsx
 1  src/features/common/Layout/Navbar/microComponents/SignInButton.tsx
 1  src/features/common/WebappButton/index.tsx
 1  src/features/projectsV2/ProjectListControls/stories/ProjectListControls.stories.tsx
[...]

PR #3150, opened 2026-09-21
chore: migrate ESLint config to flat config format

sunilsabatp:
Replace .eslintrc.js with eslint.config.mjs, porting the existing rules, plugins, and globals to the flat config API and updating dependencies accordingly.

Fixes #

Tooling only. No product code changed.

## Dependencies

| Package | Before | After |
| --- | --- | --- |
| `eslint` | 8.26 | 9.39.5 |
| `@typescript-eslint/*` | v5 | `typescript-eslint` v8.70, plus `@typescript-eslint/parser` v8 for the cypress plugin peer |
| `eslint-plugin-cypress` | 2.12 | 6.4.4 |
| `eslint-plugin-storybook` | `^0.6.15` declared, 10.6.0 installed | 0.12.0 |
| `eslint-plugin-react` | ^7.20 | ^7.37.5 |
| `eslint-plugin-react-hooks` | ^4.1 | ^5.2.0 |
| `@eslint/js`, `globals` | absent | added, required by flat config |
| `eslint-plugin-import`, `eslint-plugin-jsx-a11y` | ^2.22, ^6.3 | removed |

`eslint-plugin-import` and `eslint-plugin-jsx-a11y` were never registered as plugins. The old config only named their rules as `off`, which flat config rejects outright, so nothing was lost by dropping them.

## This fixes a red CI job

The lockfile carried `eslint-plugin-storybook@10.6.0` against a `^0.6.15` declaration. Its Storybook 9 rule `no-renderer-packages` produced 45 errors, so `npm run lint:errors` was already failing before this PR. Pinning the plugin to 0.12.0, the generation that matches our Storybook 8.6, clears all 45.

## Rule mapping

Rule intent is carried over unchanged. A few names moved between v5 and v8:

- `ban-types` was split in v8, so the successors of our `ban-types: off` are also off: `no-unsafe-function-type`, `no-wrapper-object-types`, `no-empty-object-type`. The last of those also covers the old `no-empty-interface: off`.
- `no-var-requires: 0` becomes `no-require-imports: off`.
- `no-explicit-any` pinned to `warn`, since v8 promoted it to error.
- `no-non-null-assertion` pinned to `warn`, since v8 dropped it from `recommended`.
- `no-unused-expressions` is new in v8 `recommended` and set to `warn`, so a tooling upgrade does not turn 6 existing call sites into CI failures.

## Verification

| | Before | After |
| --- | --- | --- |
| Errors | 45 | **0** |
| Warnings | 37 | 53 |
| Files linted | 678 | 678 |

Every finding was diffed against a pre-migration baseline: **0 lost**, 12 new warnings. Six are `no-unused-expressions`, six are unused `catch` bindings that v8 now checks by default. The old config already set `caughtErrorsIgnorePattern: '^_'`, so that stricter default matches the original intent.

Also confirmed:

- `npm run lint:errors`, the CI command, exits 0.
- `npx next lint` reads the flat config, so the `next build` lint step passes.
- Emotion rules still fire under ESLint 9. Verified by triggering `@emotion/jsx-import`.
- Storybook rules still scope to `*.stories.*` only.
- Cypress globals and all six pinned cypress rules resolve.
- `eslint.config.mjs` passes `prettier --check`.

## Two notes for the reviewer

**`@emotion/eslint-plugin@11.12.0` declares `eslint: 6 || 7 || 8`.** It works on 9, verified above, but it calls `context.getSourceCode()`, `getFilename()` and `getScope()`, all removed in ESLint 10. This plugin is what caps us at ESLint 9. The committed `.npmrc` sets `legacy-peer-deps=true`, so installs do not fail on the peer range.

**ESLint 9 defaults `reportUnusedDisableDirectives` to warn.** That surfaces 4 stale `eslint-disable` comments. Left in place, since they sit in product files and this PR changes no product code.

`eslint-plugin-react-hooks` stays installed but unregistered, exactly as before this branch. Turning it on is new enforcement rather than a tooling migration, and it would report 47 `rules-of-hooks` and 158 `exhaustive-deps` violations. Tracked separately in #3149.


mohitb35 (approved):
LGTM

mohitb35 (approved):
LGTM

